PR #10, opened 2022-10-22
Initialize `stats` in `LogQueue` to avoid silent crash

cjpais:
Currently the `logWorker` is silently crashing. This is happening because the `stats.sourceStats[source.name]` object is uninitialized. When trying to access a field of this `undefined` object the worker silently fails

### DIFF
--- a/packages/ponder/src/core/indexer/logQueue.ts
+++ b/packages/ponder/src/core/indexer/logQueue.ts
@@ -66,7 +66,12 @@ export const createLogQueue = ({
       );
       return;
     }
-
+    if (!stats.sourceStats[source.name]) {
+      stats.sourceStats[source.name] = {
+        matchedLogCount: 0,
+        handledLogCount: 0,
+      };
+    }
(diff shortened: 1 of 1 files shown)


0xOlias:
Looks good - thanks!

I figured out why I didn't catch this sooner. The `fastq` queues that Ponder uses for much of the indexer logic (RPC requests during the backfill, processing logs, etc) swallows errors by default - you have to explicitly provide an error handler to catch errors. I added error handlers for the 3 RPC requests queues, but forgot to add one for the queue that processes logs.

[Here's an example of the error handler](https://github.com/0xOlias/ponder/blob/main/packages/ponder/src/core/indexer/liveBlockRequestQueue.ts#L40-L46) for the live block request queue
[And here's where the error handler is missing](https://github.com/0xOlias/ponder/blob/main/packages/ponder/src/core/indexer/logQueue.ts#L129) for the log queue

I'll merge this, add the error handler, and release a patch.

PR #44, opened 2023-01-07
Improve GQL filtering

0xOlias:
- Improves generation of filter fields to better match subgraph behavior. Known remaining gaps:
  - Relationship fields can now be queried by their ID, but the nested filter field `{fieldName}_: EntityTypeFilter` does not exist (yet)
  - Derived field filter fields do not exist
- Fixes related to the handling of filter fields, especially for booleans, null, and undefined

### DIFF
--- a/packages/core/src/db/entity/postgresEntityStore.ts
+++ b/packages/core/src/db/entity/postgresEntityStore.ts
@@ -3,7 +3,11 @@ import type PgPromise from "pg-promise";
 import { DerivedField, FieldKind, ScalarField, Schema } from "@/schema/types";
 
 import type { EntityFilter, EntityStore } from "./entityStore";
-import { sqlOperatorsForFilterType } from "./utils";
+import {
+  getColumnStatements,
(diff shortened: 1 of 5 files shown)


packages/core/src/db/entity/postgresEntityStore.ts
```diff
@@ -191,36 +165,23 @@ export class PostgresEntityStore implements EntityStore {
...
-
-        if (isList) {
-          finalValue = `(${(finalValue as (string | number)[]).join(",")})`;
-        } else {
-          finalValue = `'${finalValue}'`;
-        }
-
-        whereFragments.push(`"${fieldName}" ${operator} ${finalValue}`);
-      }
+      console.log({ whereFragments });
```
holic, line 184: ooc did you mean to leave this here?

0xOlias: no lol nice catch :D need a lint rule for this in CI

PR #45, opened 2023-01-07
spin up on PORT, fall back to 42069

holic:
This should make it easier to publish to PaaS (i.e. Railway) without extra config.


0xOlias (approved):
love it

PR #46, opened 2023-01-07
migrate get/insert/update/upsert calls to SQL parameters

holic:
We don't need no Little Bobby Tables spilling our milk

### DIFF
--- a/packages/core/src/db/entity/sqliteEntityStore.ts
+++ b/packages/core/src/db/entity/sqliteEntityStore.ts
@@ -74,8 +74,8 @@ export class SqliteEntityStore implements EntityStore {
         );
       }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/db/entity/sqliteEntityStore.ts
```diff
@@ -218,9 +227,9 @@ export class SqliteEntityStore implements EntityStore {
         );
       }
 
-      const statement = `DELETE FROM "${entityName}" WHERE "id" = @id`;
+      const statement = `DELETE FROM "${entityName}" WHERE "id" = ?`;
 
-      const { changes } = this.db.prepare(statement).run({ id: id });
+      const { changes } = this.db.prepare(statement).run(id);
```
holic, line 232: The usages of `id` weren't strictly necessary, but I moved em over to `?` too for consistency.

PR #65, opened 2023-01-16
Fix `--from-etherscan` for testnets

0xOlias:


packages/create-ponder/src/helpers/getEtherscanChainId.ts
```diff
@@ -1,21 +1,67 @@
...
+  "arbiscan.io": {
+    name: "arbitrum",
+    chainId: 42161,
+    apiUrl: "https://api.arbiscan.io/api",
+  },
+  "goerli.arbiscan.io": {
+    name: "arbitrum-goerli",
+    chainId: 421613,
+    apiUrl: "https://api-goerli.arbiscan.io/api",
+  },
```
holic, line 64: wonder if wagmi folks would be open to extending their chain database with blockexplorer API URLs:
https://github.com/wagmi-dev/references/blob/3c6104c0c0e96294bd7ab0c35449a437b6e2d38f/packages/chains/src/mainnet.ts

packages/create-ponder/src/helpers/getEtherscanChainId.ts
```diff
@@ -1,21 +1,67 @@
...
-  "goerli.arbiscan.io": { name: "arbitrum-goerli", chainId: 421613 },
+  "etherscan.io": {
+    name: "mainnet",
+    chainId: 1,
+    apiUrl: "https://api.etherscan.io/api",
+  },
+  "ropsten.etherscan.io": {
+    name: "ropsten",
+    chainId: 3,
+    apiUrl: "https://api-ropsten.etherscan.io//api",
```
el-ranye, line 13: btw this api url looks malformed

holic: isn't ropsten, rinkeby, etc. dead anyway?

0xOlias: nice catch! i figured i would leave them in there in case someone wanted to migrate an old subgraph to ponder and use `--from-etherscan` but yea they're dead

Issue #73, opened 2023-01-20
Fix leading zeros in RPC requests


0xOlias:
This error can occur from certain RPCs.
```
err: Error: processing response error (body="{\"jsonrpc\":\"2.0\",\"id\":202,\"error\":{\"code\":-32602,\
"message\":\"invalid argument 0: hex number with leading zero digits\"}}\n", error={"code":-32602},
requestBody="{\"method\":\"eth_getBlockByNumber\",\"params\":[\"0x07d6d1\",false],\"id\":202,\"jsonrpc\":\"2.0\"}", 
requestMethod="POST", url="https://canto.slingshot.finance", code=SERVER_ERROR, version=web/5.7.1)
```
It seems like Ponder is failing to strip leading zeroes from hexadecimal strings, see https://ethereum.org/en/developers/docs/apis/json-rpc/#quantities-encoding

0xOlias:
Fixed in #76

PR #76, opened 2023-01-26
Resolve leading zeros issue #73

fmhall:
Should fix issue #73 - note I was not able to get the tests running properly so I'm not certain, but should be a good starting point.

Essentially `toHexString()` converts a `BigNumber` to a padded bytes string of fixed length, where `utils.hexValue` converts them to a quantity and strips leading zeros.

### DIFF
--- a/packages/core/src/indexer/queues/blockFrontfillQueue.ts
+++ b/packages/core/src/indexer/queues/blockFrontfillQueue.ts
@@ -1,4 +1,4 @@
-import { BigNumber, utils } from "ethers";
+import { utils } from "ethers";
 import fastq from "fastq";
 
 import type { Network } from "@/config/networks";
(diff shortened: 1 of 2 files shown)


packages/core/src/indexer/queues/blockFrontfillQueue.ts
```diff
@@ -1,4 +1,4 @@
-import { BigNumber } from "ethers";
+import { BigNumber, utils } from "ethers";
```
0xOlias, line 1: Should be able to remove the `BigNumber` import entirely

0xOlias (changes requested):
Looks great, thanks!! Could you make the same change in `packages/core/src/indexer/queues/logBackfillQueue.ts` to fully fix this?

fmhall:
> Looks great, thanks!! Could you make the same change in `packages/core/src/indexer/queues/logBackfillQueue.ts` to fully fix this?

Updated

0xOlias (approved):
LGTM

PR #79, opened 2023-01-26
Fix file watching on linux

### LINKED ISSUE #78: Dev flow breaks on linux
Hello! It seems the dev flow breaks on linux, at least with node >=14. Simple to reproduce, just create a fresh project and `yarn dev` (or `npm dev` or `pnpm dev`) and observe: 

```
TypeError [ERR_FEATURE_UNAVAILABLE_ON_PLATFORM]: The feature watch recursively is unavailable on the current platform, which is being used to run Node.js
    at new NodeError (node:internal/errors:387:5)
    at watch (node:fs:2299:11)
    ...
```

[Probably related to this](https://stackoverflow.com/questions/61806341/how-to-fix-the-feature-watch-recursively-is-unavailable-on-the-current-platform). Recursive watch isn't available on linux so we get an error.

0xOlias:
Fixes #78

### DIFF
--- a/.changeset/dull-zebras-suffer.md
+++ b/.changeset/dull-zebras-suffer.md
@@ -1,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
+Fix file watching on linux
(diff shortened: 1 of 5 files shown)


0xOlias:
Reproduced #78 in CI here https://github.com/0xOlias/ponder/actions/runs/4019653915/jobs/6906751024#step:8:239

PR #168, opened 2023-05-07
vitepress docs

o-az:
This is a wip but wanted some feedback while finishing up. The `increment` and vue component stuff will be removed next commit, just keeping it for fun now.

Thoughts on the theme? This is the base / default one


o-az:
temp vercel preview: https://ponder-temp-amorfati.vercel.app/

PR #183, opened 2023-05-15
fix: added missing Float scalar in the codegen

pyk:
Currently, when we define schema field with `Float` of `Float!` it will raise an error like the following:

```
 FAIL  test/codegen/main.test.ts > entity types builder > normal entity
Error: TypeScript type not found for scalar: Float
 ❯ src/codegen/buildEntityTypes.ts:27:25
     25|                 const scalarTsType = gqlScalarToTsType[field.scalarTypeName];
     26|                 if (!scalarTsType) {
     27|                   throw new Error(
       |                         ^
     28|                     `TypeScript type not found for scalar: ${field.scalarTypeName}`
     29|                   );
 ❯ src/codegen/buildEntityTypes.ts:22:12
 ❯ Module.buildEntityTypes src/codegen/buildEntityTypes.ts:19:6
 ❯ test/codegen/main.test.ts:26:20
```

this pull request added the missing `Float` field in the `buildEntityTypes` and added some new test case. Let me know if the test is unnecessary, I will remove it. 

Thanks 🙏

### DIFF
--- a/.changeset/modern-nails-travel.md
+++ b/.changeset/modern-nails-travel.md
@@ -1,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
+---
+
+Fixed a bug where codegen would fail for schemas that include a field typed as `Float`.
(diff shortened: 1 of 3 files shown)


wyze:
Just ran into this as well. What can we do to get this merged @0xOlias ?

0xOlias:
> Just ran into this as well. What can we do to get this merged @0xOlias ?

@wyze Thanks for bumping, this is merged and released in [v0.0.64](https://github.com/0xOlias/ponder/releases/tag/%40ponder%2Fcore%400.0.64).

PR #184, opened 2023-05-21
feat: Add docker configs to create-ponder

roninjin10:
Adds dockerfile to create-ponder

### TODO
- [ ] Configure env variables
- [ ] Test it out locally
- [ ] Test it out deploying docker image to railway/render
- [x] Support Postgresql in the docker compose
- [ ] Add documentation?

### .dockerignore

Copy pasted a dockerignore file.  This helps context be smaller when docker builds which speeds up docker

### docker-compose.yaml

This allows us to configure how docker runs in a yaml file instead of command line.  The configuration runs a dev server where if the user changes the code locally the docker container will hot reload

### Dockerfile

- We optimize the speed of the docker build by only copying lockfile for pnpm so we only rebuild node modules when they change.   
- The default command is `ponder start`.   
- The docker-compose.yaml (since it's used by devs not prod usually) overrides the default with `yarn dev` while mounting the users filesystem into the docker container.
- We chose the node slim image because it's the smallest official node image.  Alpine image would likely work and be smaller but in my experience alpine is way more of a headache to maintain while node-slim "just works"

### DIFF
--- a/packages/create-ponder/src/index.ts
+++ b/packages/create-ponder/src/index.ts
@@ -159,7 +159,8 @@ export const run = async (
       "scripts": {
         "dev": "ponder dev",
         "start": "ponder start",
-        "codegen": "ponder codegen"
+        "codegen": "ponder codegen",
+        "docker:dev": "docker-compose up --build"
       },
       "dependencies": {
         "@ponder/core": "latest"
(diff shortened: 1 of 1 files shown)


packages/create-ponder/src/index.ts
```diff
@@ -214,9 +295,8 @@ export const run = async (
 
   const installCommand = overrides.installCommand
     ? overrides.installCommand
-    : `${packageManager} ${
-        packageManager === "npm" ? "--quiet" : "--silent"
-      } install`;
+    : `${packageManager} ${packageManager === "npm" ? "--quiet" : "--silent"
+    } install`;
```
roninjin10, line 299: linter will fix before landing

packages/create-ponder/src/index.ts
```diff
@@ -177,6 +178,74 @@ export const run = async (
...
+# TODO ask user to configure env variables
+
+EXPOSE 42069
+
+RUN pnpm fetch --prod
+
+COPY . ./
+RUN pnpm install -r --offline --prod
+
+EXPOSE 42069
```
roninjin10, line 210:
```suggestion
# TODO ask user to configure env variables

RUN pnpm fetch --prod

COPY . ./
RUN pnpm install -r --offline --prod

EXPOSE 42069
```
copy pasta bug

packages/create-ponder/src/index.ts
```diff
@@ -177,6 +178,67 @@ export const run = async (
...
+RUN curl -f https://get.pnpm.io/v6.16.js | node - add --global pnpm
+
+# copy only pnpm-lock.yaml so docker caches
+# the installation of deps when it's unchanged
+COPY pnpm-lock.yaml ./
+
+# install deps into pnpm store
+# TODO ask user to configure env variables
+
+RUN pnpm fetch --prod
```
roninjin10, line 198:
```suggestion
# TODO ask user to configure env variables

# install deps into pnpm store
RUN pnpm fetch --prod
```

0xOlias:
@eliobricenov added a Docker example project in https://github.com/0xOlias/ponder/pull/293 - thanks for your work on this @roninjin10 🖖

roninjin10:
Nice better to show folks how to do it themselves than add it to the template in hindsight

PR #188, opened 2023-05-25
fix: event handling issue with timestamp

saribmah:
This PR Fixes event handling timestamp ranges. Before we were always grabbing the lowest block timestamp for event handling even when there were other blocks with higher timestamp available to process events from. (mostly in case of multiple chains running on the same app).

With this fix: If the minimum cached to timestamp >= backfillCutoffTimestamp, then we set the new toTimestamp to the max value that exists in the DB and continue with fetching events.

### DIFF
--- a/.changeset/poor-radios-own.md
+++ b/.changeset/poor-radios-own.md
@@ -1,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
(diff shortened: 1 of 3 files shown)


0xOlias:
LGTM!

0xOlias (approved):
LGTM!

PR #206, opened 2023-06-07
Versioned user store

0xOlias:
This PR updates the user store to persist every version of every entity that the user inserts/updates/deletes. The versions are marked by timestamp using an `effectiveFrom`/`effectiveTo` range. When querying for entities using `findUnique`/`findMany` (as is done by the GraphQL server), the latest version of each entity will be returned by default.

This unblocks two features:
1. **Time-travel queries** - Using the timestamp argument to `findUnique`/`findMany`, we can add optional arguments to the GraphQL schema that enable users to query the database at a specific timestamp (Note: this is different that Graph Protocol time-travel queries, which only support querying by block height.)
2. **Reorg handling** - In a follow-up PR, I'll add a new method to the user store that reverts the state of the database back to a specified timestamp. When handling a reorg, the event handler service will call this method, then reprocess events from the reorg timestamp to latest.

Note that this will increase the size of the database, especially for Ponder apps that update entities a lot. It will also slow down event processing slightly because the user store methods make a few additional SQL queries to maintain entity versions.

### DIFF
--- a/packages/core/src/user-handlers/service.ts
+++ b/packages/core/src/user-handlers/service.ts
@@ -77,7 +77,8 @@ export class EventHandlerService extends Emittery<EventHandlerEvents> {
   private eventProcessingMutex: Mutex;
   private eventsHandledToTimestamp = 0;
 
-  private currentLogEventBlockNumber = 0n;
+  private currentEventBlockNumber = 0n;
+  private currentEventTimestamp = 0;
 
   constructor({
     resources,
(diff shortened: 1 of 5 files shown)


pyk:
> Note that this will increase the size of the database, especially for Ponder apps that update entities a lot. It will also slow down event processing slightly because the user store methods make a few additional SQL queries to maintain entity versions.

You may want to take a look at timescaledb extension for postgres, it's designed for query time series data very fast and you can also use the compression to reduce the database size.

PR #213, opened 2023-06-08
feat: add default pagination and put a max cap on skips

r0ohafza:


docs/pages/api-reference/graphql-query.mdx
```diff
@@ -0,0 +1,13 @@
+---
+description: "API reference for the GraphQL query"
+---
+
+# GraphQL query API
```
0xOlias, line 5: I think for now it makes sense to move this under "Guides" and call it "Query the API". Then I'll expand that page to explain how to sort, paginate, and filter (the filter argument in particular needs documentation).

packages/core/src/user-store/sqlite/store.ts
```diff
@@ -438,8 +440,6 @@ export class SqliteUserStore implements UserStore {
 
     if (skip) {
       query = query.offset(skip);
-      // SQLite doesn't support OFFSET without LIMIT, so we need to set a limit.
-      if (!first) query = query.limit(-1);
```
0xOlias, line 442: I guess this is unnecessary because after `parseModelFilter`, `first` will always be defined?

packages/core/src/user-store/utils.ts
```diff
@@ -153,3 +154,32 @@ export function formatModelInstance({
 
   return instance;
 }
+
+export function parseModelFilter(filter: ModelFilter): ModelFilter {
```
0xOlias, line 158: Nit, but AFAICT you could add a default of `{}` here.

0xOlias (commented):
Looks great, had one request re docs. Thanks again.

packages/core/src/user-store/sqlite/store.ts
r0ohafza: yup

Issue #209, opened 2023-06-08
Add database indices to the event store


0xOlias:
Ponder uses a SQL database to cache blocks, logs and transactions. We should add indices to speed our queries up (this is very low hanging fruit 😄).

Some thoughts:
- There might be an automated way to determine which indices should exist.
- All of the SQL queries are in [this file](https://github.com/0xOlias/ponder/blob/main/packages/core/src/event-store/sqlite/store.ts).
- Ponder uses Kysely as a query builder, which has helper methods for adding indices.
- I'm guessing we can add these indices as a new migration in [this file](https://github.com/0xOlias/ponder/blob/main/packages/core/src/event-store/sqlite/migrations.ts).
- We'll need to do this for both SQLite and Postgres.
- I'm guessing most of the indices we add will be related to the getLogEvents method/query.

0xOlias:
Fixed in #226

PR #226, opened 2023-06-19
Add indices to event store 

k-xo:
Address #209 

Adds an index to the event store, columns were chosen based on those that were tested in `service.test.ts`, but if you'd like to have this done differently lmk!


0xOlias (changes requested):
Hi - thanks for contributing! Couple notes:
- I think we should add these indices in a new migration. At the moment there's only one (called `2023_05_15_0_initial`), you can add a new one called `2023_06_19_0_indices`. This should ensure that any Ponder apps running in production that upgrade to this version will migrate smoothly. (While we're at it, it might be a good idea to add a test case or two to confirm that behavior. Feel free to take a crack at this, otherwise I can.)
- The two event store methods that are on the "hot path" (so to speak) are `getLogEvents` and `getContractReadResult`:
  - The `contractReadResults` table should already have the only index it needs because of the primary key constraint.
  - The query behind `getLogEvents` is quite complex. I'm honestly not sure how to determine which indices should be added for this one. Maybe there is a tool that can analyze the query to determine it? Otherwise we can make a best guess.
- I also think we want an index on `logFilterCachedRanges.filterKey`, because the `insertFinalizedBlock` method does a `deleteFrom("logFilterCachedRanges").where("filterKey", "=", logFilterKey)`.

k-xo:
@0xOlias formatted the migration better, and add the `logFilterCachedRanges.filterKey` as an index.

For `getLogEvents` going through the query, some of the indexes we can add are `logs.chainId`, `logs.address` `logs.topic0`, `logs.topic1`, `logs.topic2`, `logs.topic3`,`blocks.hash`,`logs.blockHash` ,`blocks.number`, `blocks.timestamp` ,`logs.logIndex`. I've added a few of these ((no real methodology around the selection), but not all of them, more indexes = longer writes, unless we're okay with indexing all of these columns?

packages/core/src/event-store/migration.test.ts
```diff
@@ -0,0 +1,186 @@
+import "@/utils/globals";
```
0xOlias, line 1: I don't think you need this here, right?

packages/core/src/event-store/migration.test.ts
```diff
@@ -0,0 +1,186 @@
...
+
+import { hexToNumber, RpcBlock, RpcTransaction } from "viem";
+import { beforeEach, expect, test } from "vitest";
+
+import { setupEventStore } from "@/_test/setup";
+
+beforeEach(
+  async (context) => await setupEventStore(context, { skipMigrateUp: true })
+);
+test("setup creates tables", async (context) => {
```
0xOlias, line 11: Is there a reason you're including this test here? It's already in `store.test.ts`.

packages/core/src/event-store/migration.test.ts
```diff
@@ -0,0 +1,186 @@
...
+  const tables = await eventStore.db.introspection.getTables();
+  const tableNames = tables.map((t) => t.name);
+  expect(tableNames).toContain("blocks");
+  expect(tableNames).toContain("contractReadResults");
+  expect(tableNames).toContain("logFilterCachedRanges");
+  expect(tableNames).toContain("logs");
+  expect(tableNames).toContain("transactions");
+});
+
+const blockOne: RpcBlock = {
```
0xOlias, line 23: Would be great to avoid copy/pasting these RPC objects in both test files. You could move them over to `_test/constants.ts` and then import them in both `store.test.ts` and this file.

packages/core/src/event-store/migration.test.ts
k-xo: yeah this doesn't need to be here, removing this

0xOlias (changes requested):
Looks great - one last thing. You'll need to [add a changeset](https://github.com/0xOlias/ponder/blob/main/.github/CONTRIBUTING.md#changesets-workflow). This will be a `@ponder/core` patch. Description could be something like `"Added a database migration to create indices in the event store. This should improve event handler performance."`

0xOlias (approved):
LGTM, thanks!

PR #230, opened 2023-06-22
Add ERC721 example app

0xOlias:
This PR adds an ERC271 example app for the Smol Brains NFT collection on Arbitrum. I also noticed some RPC provider inconsistencies while testing this out. Specifically, Alchemy seems to rename `transaction.type` to `transaction.arbType` for some bizarre reason. Quicknode works as expected.

I've ran into enough issues related to this specific field that it may be worth making it nullable in the database, but for now I added a default of `"0x0"` so that Alchemy still works.

Also fixed a bug where `block.baseFeePerGas` was assumed to be non-null when formatting event objects coming out of the event store.

### DIFF
--- a/.changeset/short-dingos-sip.md
+++ b/.changeset/short-dingos-sip.md
@@ -1,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
+Fixed transaction and block formatting to handle Arbitrum RPC data formats.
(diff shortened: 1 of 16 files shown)


wyze:
> Ah, makes sense. While we're here - any other patches you had to make?

I haven't got it tested yet, waiting on historical sync. But I exposed the `findMany` API to the Model/Entities. I always found that missing from subgraphs, so we'd have to store IDs and loop over them for some things.

PR #232, opened 2023-06-22
Update production.mdx

jcheese1:
Had to add DATABASE_URL to connect postgres to a railway service, so adding that in README


0xOlias:
Ah, looks like Railway changed this. It used to provide `DATABASE_URL` automatically if you had a Postgres service in the same project.

Thanks a bunch!

PR #234, opened 2023-06-26
feat: add timestamp to model filter

### LINKED ISSUE #214: Time-travel queries
Ponder should support time-travel queries in a manner similar to how The Graph does.

The entity store already stores every version of every entity, marked by `effectiveFrom` and `effectiveTo` timestamps (added in [this PR](https://github.com/0xOlias/ponder/pull/206)). This should make it pretty simple to add a `timestamp` option to the GraphQL filter argument and the entity store `findMany` method.

r0ohafza:
resolves #214 

- Added timestamp to singular and plural fields
- Added timestamp to Model filter


0xOlias:
Thanks @r0ohafza :D

PR #245, opened 2023-07-03
Update custom-filters.mdx

cbergz:
Added a new example with code for a config that uses filters for contracts on the same abi.


0xOlias (approved):
LGTM! Thanks for contributing!

PR #265, opened 2023-07-20
Salief/add comments to gql schema

salieflewis:
Opening this up :)

1. Is this sufficiently helpful? My initial thoughts are the link to the `examples` directory is, without that GH discussion exchange I wouldn't even know those examples existed.
2. How could I test a change like this locally?
3. Do I have to worry about formatting locally, or does the GH action run prettier and all that before a contribution gets included?

Thanks in advance!

cc: @0xOlias

### DIFF
--- a/packages/create-ponder/src/templates/basic.ts
+++ b/packages/create-ponder/src/templates/basic.ts
@@ -12,20 +12,22 @@ export const fromBasic = ({ rootDir }: { rootDir: string }) => {
   writeFileSync(abiAbsolutePath, abiFileContents);
 
   const schemaGraphqlFileContents = `
-  # The entity types defined below map to database tables.
-  # The functions you write as event handlers inside the \`src/\` directory are responsible for creating and updating records in those tables.
-  # Your schema will be more flexible and powerful if it accurately models the logical relationships in your application's domain.
-  # Visit the [documentation](https://ponder.sh/guides/design-your-schema) or view the [\`examples/\`](https://github.com/0xOlias/ponder/tree/main/examples) directory for further guidance on designing your schema.
+    # The entity types defined below map to database tables.
+    # The functions you write as event handlers inside the \`src/\` directory are responsible for creating and updating records in those tables.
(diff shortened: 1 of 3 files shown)


packages/create-ponder/src/templates/basic.ts
```diff
@@ -1,17 +1,21 @@
...
+import prettier from 'prettier';
 
-import type { PartialConfig } from "@/index";
+import type { PartialConfig } from '@/index';
 
 export const fromBasic = ({ rootDir }: { rootDir: string }) => {
   const abiFileContents = `[]`;
 
-  const abiRelativePath = "./abis/ExampleContract.json";
+  const abiRelativePath = './abis/ExampleContract.json';
```
0xOlias, line 10: Looks like you formatted (maybe on file save?) but it didn't use `.eslintrc`. No pressure to get this running (lmk if you'd rather me fix & ship this) but you might just need to `pnpm install` and then `pnpm lint:format` in the project root. Noticed that the `pnpm-lock.yaml` also has changes (it shouldn't).

0xOlias (changes requested):
Thanks for opening this!

Re: running tests, take a look at the [contribution guide](https://github.com/0xOlias/ponder/blob/main/.github/CONTRIBUTING.md). Lmk if you aren't able to get it working / have feedback on the guide.

packages/create-ponder/src/templates/basic.ts
salieflewis: Happy for you to take the reins on this. Was having a weird issue where even after installing deps none of the scripts were running. Deleted it locally, but then realized its kinda hard to get back my branch if it wasn't something I pushed up against my local fork.

For future, to avoid `pnpm-lock.yaml` getting pushed, would you recommend just hand picking the files I've changed rather than `git add .`. And second question, regarding the changeset. Does it update to include the latest changes, after which for example, you've requested a review? Hope that makes sense. Thanks for your time!

PR #276, opened 2023-07-26
Initial Telemetry Service implementation

eliobricenov:
Solves #277


packages/core/src/telemetry/anonymous-meta.ts
```diff
@@ -0,0 +1,44 @@
...
+};
+
+let traits: AnonymousMeta | undefined;
+
+export function getAnonymousMeta(): AnonymousMeta {
+  if (traits) {
+    return traits;
+  }
+
+  const cpus = os.cpus() || [];
```
0xOlias, line 27: Probably doesn't matter here, but interesting to see that this is a very slow synchronous function. https://twitter.com/jarredsumner/status/1683710506924933120?s=20

packages/core/src/telemetry/project-id.ts
```diff
@@ -0,0 +1,34 @@
+import { exec } from "child_process";
```
0xOlias, line 1: I wonder if rather than using exec directly, we should use `execa` here so we don't need to do the promise-ification manually. (https://www.npmjs.com/package/execa?activeTab=readme).

packages/core/src/types/telemetry-event.ts
```diff
@@ -0,0 +1,4 @@
+export type TelemetryEvent = {
+  eventName: string;
+  payload: object;
+};
```
0xOlias, line 4: This `types/` dir is kinda tech debt - I'd just put this at the top of the `service.ts` file. Because the `common.telemetry.record(...)` function will be nicely typed, callers will almost certainly be able to use the inferred type and won't need to import this.

0xOlias (changes requested):
Looking great - I did some more digging into the Next code and shared my findings.

packages/core/src/telemetry/service.ts
```diff
@@ -0,0 +1,79 @@
...
+    this.sessionId = randomBytes(32).toString("hex");
+  }
+
+  get anonymousId(): string {
+    return randomBytes(32).toString("hex");
+  }
+
+  get salt(): string {
+    return randomBytes(16).toString("hex");
+  }
```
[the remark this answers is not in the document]
eliobricenov: yeah definitely! this is why the distDir prop is needed, gonna add that

packages/core/src/telemetry/service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export class TelemetryService {
+  private readonly sessionId: string;
+  private readonly TELEMETRY_DISABLED: boolean;
+  private distDir: string;
+  private loadProjectId: string | null = null;
+
+  constructor({ distDir }: { distDir: string }) {
+    // we want to persist the config across restarts
+    this.distDir = distDir;
```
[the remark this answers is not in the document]
eliobricenov: awesome thank you!

packages/core/src/telemetry/project-id.ts
eliobricenov: hmm - I think for something this simple we can skip adding a package and save those 77kb

0xOlias: Agree we don't need `execa`! You know what - I'm backtracking 😅 , I actually think we should call this function `getGitRemoteUrl` and it should be in it's own file here called `git.ts` or `remote.ts`. I think what bothered me was calling it `project-id` because this file/fn doesn't actually generate the project ID, it just returns the raw git remote url (which is fine!).

packages/core/src/Ponder.ts
```diff
@@ -72,8 +74,9 @@ export class Ponder {
     });
     const errors = new UserErrorService();
     const metrics = new MetricsService();
+    const telemetry = new TelemetryService(options);
```
0xOlias, line 77: I almost always prefer passing an object with named/typed properties over an anonymous positional argument, so this line would read `const telemetry = new TelemetryService({ options} );` using the shorthand, and the `TelemetryService` constructor would be typed `constructor({ options }: { options: Options }) { ... }`.

packages/core/src/_test/setup.ts
```diff
@@ -45,11 +46,13 @@ declare module "vitest" {
...
 
 beforeEach((context) => {
+  const options = buildOptions({ cliOptions: { configFile: "", rootDir: "" } });
   context.common = {
-    options: buildOptions({ cliOptions: { configFile: "", rootDir: "" } }),
+    options,
     logger: new LoggerService({ level: "silent" }),
     errors: new UserErrorService(),
     metrics: new MetricsService(),
+    telemetry: new TelemetryService(options),
```
0xOlias, line 55: Hmm. Once you have the tests running, I'm curious where the config.json will be generated during tests.

packages/core/src/telemetry/service.ts
```diff
@@ -0,0 +1,165 @@
...
+export class TelemetryService {
+  private readonly conf: Conf<TelemetryConfig>;
+  private readonly sessionId: string;
+  private readonly TELEMETRY_DISABLED: boolean;
+  private distDir: string;
+  private loadProjectId: string | null = null;
+
+  constructor(options: Options) {
+    this.conf = new Conf({ projectName: "ponder", cwd: options.ponderDir });
+    this.distDir = options.ponderDir;
```
0xOlias, line 64: Don't think `distDir` is used anywhere anymore, right?

packages/core/src/telemetry/service.ts
```diff
@@ -0,0 +1,165 @@
...
+    this.TELEMETRY_DISABLED = Boolean(process.env.TELEMETRY_DISABLED);
```
0xOlias, line 65: I would do some manual testing do make sure this works as intended (if you haven't already). Also might just call the instance property `this.disabled`.

packages/core/src/telemetry/service.ts
```diff
@@ -0,0 +1,165 @@
...
+  private async getProjectId() {
+    this.loadProjectId = this.loadProjectId ?? (await getRawProjectId());
+    return this.oneWayHash(this.loadProjectId);
+  }
+
+  private get isDisabled() {
+    return this.TELEMETRY_DISABLED || !this.conf.get("enabled");
+  }
+
+  private notify() {
```
0xOlias, line 121: This looks awesome

0xOlias (changes requested):

packages/core/src/Ponder.ts
eliobricenov: oh good call!

packages/core/src/telemetry/project-id.ts
eliobricenov: agree - got too biased by Nextjs implementation

packages/core/src/telemetry/service.ts
eliobricenov: we're going to need this later for the `flushDetached` method but can remove for now and add later (if needed)

packages/core/src/telemetry/project-id.ts
0xOlias: Looks good - should be able to remove this file now.

0xOlias (approved):
One dangling file but other than that, LGTM!

0xOlias (approved):
LGTM

PR #280, opened 2023-07-26
add API handler to forward payload to Segment

### LINKED ISSUE #279: API route for Segment integration
We need to implement an API route that facilitates forwarding the payload sent by the Telemetry Service to the Segment API. This integration will allow us to track and analyze telemetry data effectively.

We can create a new API endpoint in the documentation app to receive the telemetry payload.

eliobricenov:
closes #279

### DIFF
--- a/docs/pages/api/telemetry/index.ts
+++ b/docs/pages/api/telemetry/index.ts
@@ -1,0 +1,49 @@
+import { NextApiRequest, NextApiResponse } from "next";
+
+async function forwardTelemetry(req: NextApiRequest, res: NextApiResponse) {
+  if (req.method === "POST") {
+    try {
+      if (!process.env.SEGMENT_WRITE_KEY) {
+        throw new Error("No Segment write key found.");
+      }
+
+      const telemetryPayload = req.body;
+      const segmentWriteKey = process.env.SEGMENT_WRITE_KEY;
+      const segmentEndpoint = "https://api.segment.io/v1/track";
+      const authorization = `${segmentWriteKey}:`;
+
+      const response = await fetch(segmentEndpoint, {
+        method: "POST",
+        headers: {
+          Authorization: `Basic ${Buffer.from(authorization).toString(
+            "base64"
+          )}`,
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify(telemetryPayload),
+      });
+
+      if (response.ok) {
+        res.status(200).json({
+          message: "Telemetry data processed successfully.",
+        });
+      } else {
+        const errorMessage = response.statusText;
+        console.error("Error processing telemetry data", errorMessage);
+        res.status(response.status).json({
+          error: "An error occurred while processing telemetry data.",
+        });
+      }
+    } catch (error) {
+      console.error("Error forwarding telemetry data to Segment API:", error);
+      res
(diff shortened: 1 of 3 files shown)


docs/pages/api/telemetry/index.ts
```diff
@@ -0,0 +1,38 @@
+import { Analytics } from "@segment/analytics-node";
+import { NextApiRequest, NextApiResponse } from "next";
+
+export default async function forwardTelemetry(
```
0xOlias, line 4: Do we want this to be an edge function? Genuinely not sure! (https://nextjs.org/docs/pages/building-your-application/routing/api-routes#edge-api-routes)

docs/pages/api/telemetry/index.ts
```diff
@@ -0,0 +1,38 @@
...
+    console.error('Missing required environment variable "SEGMENT_WRITE_KEY".');
+    return res
+      .status(500)
+      .json({ error: "An error occurred while processing telemetry data." });
+  }
+
+  try {
+    const analytics = new Analytics({
+      writeKey: process.env.SEGMENT_WRITE_KEY,
+    });
```
0xOlias, line 23: I wonder if we can do this stuff above, outside the function scope.

docs/pages/api/telemetry/index.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    await analytics.track({
+      ...req.body,
+    });
+
+    res.status(200).json({
+      message: "Telemetry data processed successfully.",
+    });
```
0xOlias, line 31: Do we need to await the request to Segment? I wonder if we can actually send the HTTP response back to the client immediately, then try to forward to segment.

0xOlias (approved):
LGTM! Couple ideas/questions

docs/pages/api/telemetry/index.ts
eliobricenov: we can do that - the tradeoff is that the client will never know if the event was forwarded successfully. In this case, I don't think it is important but it'd be good to track if the forwarding fails

docs/pages/api/telemetry/index.ts
eliobricenov: no :/ - I tried doing it outside of the function but Typescript complains.

this is what chatgpt says (although I'm not exactly sure this is true):

> In Node.js, environment variables are usually set when the process starts and are available for the duration of the process. However, in a serverless or "edge" environment (like the one you're running your Next.js code in), individual requests are handled in isolation and might not share the same environment variables state.

docs/pages/api/telemetry/index.ts
eliobricenov: i'm not sure either but it seems to work!

docs/pages/api/telemetry/index.ts
0xOlias: Yea. I wonder if Vercel has a dashboard for "failed edge function invocations" or something. Like, maybe we can just throw in the function body if the `analytics.track` call fails. In any case, this works as is for now.

I don't think the client should know/care if the forwarding failed - there's nothing they can do about it.

0xOlias (approved):
LGTM!

PR #283, opened 2023-07-29
Telemetry detached flush

### LINKED ISSUE #281: Add an event queue and flushDetached to the telemetry service
We should implement an event queue for the telemetry `record` actions in the Telemetry service. There should also be a `flushDetached` method that spawns a process to process the remaining events even if the entire process is shut down.

eliobricenov:
closes #281


packages/core/src/telemetry/service.ts
```diff
@@ -56,19 +129,29 @@ export class TelemetryService {
...
+    }
+
+    const context = await this.getContext();
+    const serializedEvent = { ...event, ...context };
 
-  async record(_events: TelemetryEvent | TelemetryEvent[]) {
-    const events = Array.isArray(_events) ? _events : [_events];
-    return this.submitEvents(events);
+    this.events.push(serializedEvent);
+    return this.queue.add(() => this.processEvent());
```
0xOlias, line 141: Don't think you need to return anything here

packages/core/src/telemetry/service.ts
```diff
@@ -1,38 +1,91 @@
...
+  private context?: Context;
+  private metadata?: AnonymousMeta;
 
   constructor({ options }: { options: Options }) {
-    this.conf = new Conf({ projectName: "ponder", cwd: options.ponderDir });
+    this.conf = new Conf({
+      projectName: "ponder",
+      cwd: options.ponderDir,
+      configName: "telemetry-config",
+    });
```
0xOlias, line 70: I think in our case we don't want to specify `cwd` ([see docs](https://github.com/sindresorhus/conf#cwd)). The purpose of this `Conf` is to persist across all Ponder projects on the same machine. If we use the `ponderDir` of this Ponder app as the location for the config, the user will get notified every time they create a Ponder app and will have a different anonymous ID and salt for each project. To fix, I think you can just remove the `cwd` arguments. And, it should be fine to omit `configName` as well, we can use the default.

To test it, you can probably just `config.delete` all of the keys in a beforeEach.

packages/core/src/telemetry/service.ts
```diff
@@ -1,38 +1,91 @@
...
+    this.distDir = options.ponderDir;
```
0xOlias, line 71: I would probably create a class property for `this.options: Options` rather than this. Then in the detachedFlush method, can `path.join(this.options.ponderDir, "events.json")` or something similar.

packages/core/src/telemetry/service.ts
```diff
@@ -1,38 +1,91 @@
...
     this.TELEMETRY_DISABLED = Boolean(process.env.TELEMETRY_DISABLED);
```
0xOlias, line 72: So, I actually think we should move this to `src/config/options.ts`. You can add a property to the `Options` type called `telemetryDisabled: boolean` which is set to `Boolean(process.env.TELEMETRY_DISABLED)`.

Then, in this class we can just use `this.options.telemetryDisabled`.

Would also rename the env var to `PONDER_TELEMETRY_DISABLED`.

packages/core/src/telemetry/service.ts
```diff
@@ -78,6 +161,27 @@ export class TelemetryService {
...
+    if (this.context) {
+      return this.context;
+    }
+
+    this.context = {
+      sessionId: this.sessionId,
+      projectId: await this.getProjectId(),
+      anonymousId: this.anonymousId,
+      meta: this.anonymousMeta,
+    };
```
0xOlias, line 180: Don't think we need private getters for anonymousId and anonymousMeta, because they will only ever be used here. Can inline them. This means we can also get rid of the `metadata` class property.

If it were me, I would also inline getProjectId, and get rid of the `rawProjectId` class property.

packages/core/src/telemetry/service.test.ts
```diff
@@ -1,23 +1,122 @@
...
+
+test("kill method should persis events queue and trigger detached flush", async () => {
+  const persistedEventsPath = options.ponderDir + "/telemetry-events.json";
+  const telemetry = new TelemetryService({ options });
+
+  fetchMocker.mockImplementation(() => {
+    throw { name: "AbortError" };
+  });
+
+  for (let i = 0; i < 10; i++) {
```
0xOlias, line 106: There's a `utils/range.ts` function, could do `for (const i of range(10)) { ... }` 🤓

packages/core/src/telemetry/service.test.ts
```diff
@@ -1,23 +1,122 @@
...
+
+const metadata = {
+  systemPlatform: os.platform(),
+  systemRelease: os.release(),
+  systemArchitecture: os.arch(),
+  cpuCount: cpus.length,
+  cpuModel: cpus.length ? cpus[0].model : null,
+  cpuSpeed: cpus.length ? cpus[0].speed : null,
+  memoryInMb: Math.trunc(os.totalmem() / Math.pow(1024, 2)),
+};
```
0xOlias, line 24: Don't think it's necessary to do a full object match here - fine to just check if the keys are present.

packages/core/src/telemetry/service.test.ts
```diff
@@ -1,23 +1,122 @@
...
 import { TelemetryService } from "@/telemetry/service";
 
-let originalEnv: NodeJS.ProcessEnv;
+const cpus = os.cpus() || [];
 
-beforeEach(() => {
-  originalEnv = { ...process.env };
+const options = buildOptions({
+  cliOptions: { configFile: "", rootDir: "" },
+});
```
0xOlias, line 14: If you take a look at `_test/setup.ts`, you can see that we're already doing this and including in the vitest test context that gets passed to every test. So, you should be able to do:
```ts
test("...", async ({ options }) => {
  const service = new TelemetryService({ options })
}
```

0xOlias (changes requested):
Looking good!

packages/core/src/telemetry/service.ts
eliobricenov: that's a great call actually - I do think we can keep `anonymousId` because it uses the persisted value from `conf` similar to `salt` and provide a readable way to say "hey, get this from the file otherwise set it"

packages/core/src/telemetry/service.ts
```diff
@@ -78,12 +160,46 @@ export class TelemetryService {
     return hash.digest("hex");
   }
 
-  private async getProjectId() {
-    if (this.rawProjectId) {
-      return this.oneWayHash(this.rawProjectId);
+  private flushDetached() {
+    const serializedEvents = JSON.stringify(this.events);
```
0xOlias, line 164: Might be worth adding a check that there are any events to be flushed. Should be able to return early (not kick off the detached flush) if `this.events.length === 0`.

This might make the tests more annoying (hard to tell from reading).

packages/core/src/telemetry/service.ts
```diff
@@ -78,12 +160,46 @@ export class TelemetryService {
...
+    const fileName = `${this.options.ponderDir}/telemetry-events.json`;
```
0xOlias, line 165: I think you should probably `path.join()` these (pretty sure this code would break on Windows).

0xOlias (approved):
Few suggestions!

PR #287, opened 2023-08-01
test(telemetry): enable tests

eliobricenov:


packages/core/src/telemetry/service.test.ts
```diff
@@ -8,17 +8,14 @@ import { TelemetryService } from "@/telemetry/service";
 
 const conf = new Conf({ projectName: "ponder" });
 
-let latestFetchBody: any = undefined;
-const fetchSpy = vi.fn().mockImplementation((_url, args) => {
-  latestFetchBody = JSON.parse(args.body);
-});
+const fetchSpy = vi.fn().mockImplementation(() => vi.fn());
```
eliobricenov, line 11: there is no need to store the fetch body in a separate variable if we can access the body with `fetchSpy.mock.calls`

Also, this doesn't work if we reset the mock in the `beforeEeach` hook. We do this to reset the function calls count between tests.

packages/core/src/telemetry/service.ts
```diff
@@ -15,7 +15,7 @@ import { postEvent } from "@/telemetry/post-event";
 import { getGitRemoteUrl } from "@/telemetry/remote";
 
 type TelemetryEvent = {
-  eventName: string;
+  event: string;
```
eliobricenov, line 18: right now this doesn't make any sense but Segment requires an `event` field for their track events. This doesn't mean we're tied to Segment but doesn't hurt to set it

packages/core/src/telemetry/detached-flush.test.ts
```diff
@@ -0,0 +1,37 @@
+import child_process from "node:child_process";
+import { rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import process from "node:process";
+import { expect, test } from "vitest";
+
+test("detached flush script should run without error", async ({ common }) => {
```
0xOlias, line 8: Moved this test to its own file, because it's really testing the detached-flush script on its own, not as it is used by the telemetry service.

packages/core/src/telemetry/detached-flush.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  const events = Array.from({ length: 5 }, () => ({
+    event: "test",
+    payload: {},
+  }));
+
+  const telemetryEventsFilePath = path.join(tmpdir(), "events.json");
```
0xOlias, line 14: Can just use the handy `tmpdir()` function here to avoid potentially leaving undesired files in the repo.

packages/core/src/telemetry/service.ts
```diff
@@ -11,11 +11,10 @@ import pc from "picocolors";
 import process from "process";
 
 import type { Options } from "@/config/options";
-import { postEvent } from "@/telemetry/post-event";
```
0xOlias, line 14: Decided to move this into `service.ts` considering it's not used anywhere else.

packages/core/src/telemetry/service.ts
```diff
@@ -134,23 +133,24 @@ export class TelemetryService {
...
+    const telemetryEventsFilePath = path.join(
+      this.options.ponderDir,
+      "telemetry-events.json"
+    );
+    fs.writeFileSync(telemetryEventsFilePath, serializedEvents);
+
     child_process.spawn(process.execPath, [
       path.join(__dirname, "flush-detached.js"),
-      fileName,
+      this.options.telemetryUrl,
```
0xOlias, line 147: Made the telemetry service URL  an argument to the script and passed it along here. Makes easier unit testing and allows us to only store the telemetry URL in one place.

packages/core/src/config/options.ts
```diff
@@ -50,11 +51,14 @@ export const buildOptions = ({
     generatedDir: "generated",
     ponderDir: ".ponder",
     logDir: ".ponder/logs",
-    telemetryDisabled: Boolean(process.env.PONDER_TELEMETRY_DISABLED),
+
     port: Number(process.env.PORT ?? 42069),
     maxHealthcheckDuration:
       configOptions?.maxHealthcheckDuration ?? railwayHealthcheckTimeout ?? 240,
 
+    telemetryUrl: "https://ponder.sh/api/telemetry",
```
0xOlias, line 59: Added the telemetry URL to options. Not currently configurable, but nice to have it in one place.

0xOlias (approved):
Left some quick comments!

PR #290, opened 2023-08-02
add setup and kill telemetry events - use nodejs runtime in telemetry api route

eliobricenov:


packages/core/src/telemetry/service.ts
```diff
@@ -143,7 +143,7 @@ export class TelemetryService {
     fs.writeFileSync(telemetryEventsFilePath, serializedEvents);
 
     child_process.spawn(process.execPath, [
-      path.join(__dirname, "flush-detached.js"),
+      path.join(__dirname, "detached-flush.js"),
```
eliobricenov, line 146: 🙃

packages/core/src/telemetry/service.ts
```diff
@@ -15,7 +15,7 @@ import { getGitRemoteUrl } from "@/telemetry/remote";
 
 type TelemetryEvent = {
   event: string;
-  payload: object;
+  payload?: object;
```
eliobricenov, line 18: I don't think we need to enforce a payload, only event is really necessary.

packages/core/src/telemetry/detached-flush.js
```diff
@@ -19,7 +19,7 @@ async function detachedFlush() {
     events.map(async (event) => {
       await fetch(telemetryUrl, {
         method: "POST",
-        body: JSON.stringify(event.payload),
+        body: JSON.stringify(event),
```
eliobricenov, line 22: We want to include the whole event object, including event name and meta.

docs/pages/api/telemetry/index.ts
```diff
@@ -1,44 +1,32 @@
...
-    { status: 200 }
-  );
+  res.status(200).json({
+    message: "Telemetry data processed successfully.",
+  });
 
   await analytics.track({
-    ...data,
+    ...req.body,
   });
```
0xOlias, line 31: Is there a reason you need to copy/spread here?

docs/pages/api/telemetry/index.ts
```diff
@@ -1,44 +1,32 @@
...
-      { status: 405 }
-    );
+    return res.status(405).json({
+      error: `Method ${req.method} Not Allowed`,
+    });
   }
 
   const analytics = new Analytics({
     writeKey: process.env.SEGMENT_WRITE_KEY,
   });
```
0xOlias, line 23: I wonder if, now that this is an API route, you can initialize this outside the function scope.

packages/core/src/telemetry/detached-flush.js
0xOlias: Makes sense, think this was an oversight on my part!

packages/core/src/telemetry/detached-flush.js
```diff
@@ -19,13 +19,15 @@ async function detachedFlush() {
...
         headers: {
           "Content-Type": "application/json",
         },
       });
     })
-  );
+  ).finally(() => {
+    fs.rmSync(eventsFilePath);
+  });
```
0xOlias, line 30: I would probably change this to
```ts
try {
  await Promise.all( ... )
} catch (error) {
  console.error(error)
}

fs.rmSync(eventsFilePath)
```

[See this](https://stackoverflow.com/a/58056140).

packages/core/src/telemetry/detached-flush.test.ts
```diff
@@ -27,7 +27,10 @@ test("detached flush script should run without error", async ({ common }) => {
     );
   });
 
-  rmSync(telemetryEventsFilePath);
+  // if the script ran successfully, the temporary file should be deleted
+  if (existsSync(telemetryEventsFilePath)) {
+    rmSync(telemetryEventsFilePath);
+  }
```
0xOlias, line 33: I think this makes more sense as an assertion

```ts
expect(existsSync(telemetryEventsFilePath)).toBeFalsy()
```

Or something like that. If the script is leaving around dead files, that should be a test failure.

packages/core/src/telemetry/service.ts
0xOlias: 3a7

0xOlias (commented):
Looks great.

Had a thought - I wonder if it makes sense to use `os.tmpdir()` rather than `/.ponder/` for the location of the `telemetry-events.json` file.

packages/core/src/telemetry/detached-flush.test.ts
eliobricenov: great call

docs/pages/api/telemetry/index.ts
eliobricenov: not at all - I missed changing this when we no longer needed to do `const data = await req.json()` from the edge version

docs/pages/api/telemetry/index.ts
eliobricenov: we can!

packages/core/src/telemetry/detached-flush.js
0xOlias: Hmm - I think it's fine to ignore any errors coming from the script tbh. The only real source of errors is segment HTTP request failure/error. Our test for the script is pretty thorough.

PR #310, opened 2023-08-10
feat: add support for Ponder ESLint config

### LINKED ISSUE #305: Warning on missing promise await on upsert operations
It's not uncommon for new users of Ponder (including myself 😅) to overlook the need for awaiting entities' upsert operations. This oversight can lead to issues where the node process doesn't wait for the new record to be created before proceeding with the syncing process. As a result, the subsequent sync cycle may attempt to [update an instance from the previous sync cycle](https://github.com/0xOlias/ponder/blob/main/packages/core/src/user-store/postgres/store.ts#L324).

Example:
```typescript
ponder.on("Kongz:Transfer", async ({ event, context }) => {
  context.entities.Account.upsert({
    // ^? this is missing an await
    id: event.params.from,
    create: {},
    update: {},
  });
});
```

Generates this error:
```
Error: Cannot update an instance in the past
```

To address this issue, one potential solution is to introduce a lint rule that can alert users about the absence await keyword misisng. Similar to React's `exhaustive-deps` warning.

### LINKED ISSUE #315: Create shared ESLint configuration for Ponder
It'd be great if we can have a base ESLint config that users can extend from. This file will have all the configs we believe are useful for Ponder apps.

eliobricenov:
closes #315


packages/create-ponder/src/index.ts
```diff
@@ -201,6 +202,17 @@ export const run = async (
     prettier.format(tsConfig, { parser: "json" })
   );
 
+  const eslintConfig = `
+    {
+      "extends": "eslint-config-ponder"
```
0xOlias, line 207: Pretty sure this can just be `"extends": "ponder"` 🤓

0xOlias (commented):
Just pushed a commit with some small tweaks to get it working in the monorepo

packages/create-ponder/src/index.ts
eliobricenov: yes!

examples/ethfs/package.json
```diff
@@ -11,6 +11,7 @@
   },
   "devDependencies": {
     "abitype": "^0.8.11",
+    "eslint-config-ponder": "workspace:latest",
```
0xOlias, line 14: Should also add `eslint` as a dev dependency for all the examples

examples/token-erc20/package.json
```diff
@@ -4,13 +4,16 @@
   "scripts": {
     "dev": "ponder dev",
     "start": "ponder start",
-    "codegen": "ponder codegen"
+    "codegen": "ponder codegen",
+    "lint": "eslint . --debug"
```
0xOlias, line 8: Can remove the `--debug`. Also, all of the examples should have this lint script as I mentioned below.

packages/create-ponder/src/index.ts
```diff
@@ -159,12 +165,14 @@ export const run = async (
       "scripts": {
         "dev": "ponder dev",
         "start": "ponder start",
+        ${options.eslint ? `"lint": "eslint .",` : ""}
```
0xOlias, line 168: Oh this is great! I'd add this lint script to all of the examples/ projects too. In general I think it's nice for the examples/ projects to look like they were just generated using create-ponder.

packages/eslint-config-ponder/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "eslint-config-ponder",
+  "version": "0.0.81",
```
0xOlias, line 3: Should make this equal to whatever the latest version is on main (for @ponder/core and create-ponder).

packages/eslint-config-ponder/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "eslint-config-ponder",
```
0xOlias, line 2: I think we want to call the package `eslint-config-ponder` as shown here. I saw in a few other places it's referenced as `@ponder/eslint-config`, so I think you'll have to change those.

0xOlias (commented):
Some nits

packages/eslint-config-ponder/package.json
eliobricenov: jfc - that's what I get for coding before the first cup of coffee

0xOlias (approved):

PR #324, opened 2023-08-18
Fix realtime error crashing service

0xOlias:


packages/core/src/realtime-sync/service.test.ts
```diff
@@ -198,6 +204,47 @@ test("handles new blocks", async (context) => {
   await service.kill();
 });
 
+test("handles error while fetching new latest block gracefully", async (context) => {
```
0xOlias, line 207: Before adding the try/catch, this test fails with an error similar to the one reported.

PR #333, opened 2023-08-29
Watch and rebuild config file on change

nikugogoi:
Part of https://github.com/0xOlias/ponder/issues/236

> There's really no reason that we can't also hot reload ponder.config.ts, so we should implement that.

Implemented hot-reloading of `ponder.config.ts` according to implementation description in issue

- Refactored code in `bin/ponder.ts` and `Ponder.ts`
  - Watch and build config from config file path in build service
  - Moved setting up services based on config to new `init` method
  - Updated build service listener for `newConfig` (triggered on config file change)
    - Kill config based services
    - Reinitialize services using `dev` method
  - Added option for overriding config by passing a resolved config in constructor (required in tests)

### DIFF
--- a/packages/core/src/Ponder.ts
+++ b/packages/core/src/Ponder.ts
@@ -120,6 +120,10 @@ export class Ponder {
         ? new SqliteUserStore({ db: database.db })
         : new PostgresUserStore({ pool: database.pool }));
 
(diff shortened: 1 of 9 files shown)


0xOlias:
Resolved in #466

PR #335, opened 2023-09-04
refactor all blob encoded bigints to postgres numeric 78

kliyer-ai:
See https://github.com/0xOlias/ponder/issues/334 for reference.

This PR changes the Postgres Store to use `numeric(78)` for the BigInts. This is not backwards compatible because it directly edits the migrations file. Maybe someone can help write a migration to convert all `bytea` encoded BigInts to `numeric(78)`.

Secondly, it creates views for the most current model without the version id. This should make directly accessing the DB and morgen with the most recent model version much easier.

### DIFF
--- a/.changeset/tasty-planets-buy.md
+++ b/.changeset/tasty-planets-buy.md
@@ -1,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
+use numeric32 for bigint on postgres instead of bytea
--- a/packages/core/src/user-store/postgres/store.ts
+++ b/packages/core/src/user-store/postgres/store.ts
@@ -23,7 +23,7 @@ const gqlScalarToSqlType = {
   Boolean: "integer",
   Int: "integer",
   String: "text",
-  BigInt: sql`numeric(32)`,
+  BigInt: sql`numeric(78)`,
   Bytes: "text",
   Float: "text",
 } as const;
--- a/packages/core/src/user-store/postgres/store.ts
+++ b/packages/core/src/user-store/postgres/store.ts
@@ -3,7 +3,6 @@ import { CompiledQuery, Kysely, PostgresDialect, sql } from "kysely";
 import { Pool } from "pg";
 
 import type { Schema } from "@/schema/types";
-import { blobToBigInt } from "@/utils/decode";
 
 import type {
   ModelInstance,
@@ -24,7 +23,7 @@ const gqlScalarToSqlType = {
   Boolean: "integer",
   Int: "integer",
   String: "text",
-  BigInt: sql`bytea`,
+  BigInt: sql`numeric(32)`,
   Bytes: "text",
   Float: "text",
 } as const;
@@ -696,9 +695,7 @@ export class PostgresUserStore implements UserStore {
       }
 
       if (field.kind === "SCALAR" && field.scalarTypeName === "BigInt") {
-        deserializedInstance[field.name] = blobToBigInt(
-          value as unknown as Buffer
-        );
+        deserializedInstance[field.name] = value;
         return;
       }
 
@@ -706,9 +703,7 @@ export class PostgresUserStore implements UserStore {
         field.kind === "RELATIONSHIP" &&
         field.relatedEntityIdType.name === "BigInt"
       ) {
-        deserializedInstance[field.name] = blobToBigInt(
-          value as unknown as Buffer
-        );
+        deserializedInstance[field.name] = value;
         return;
       }
 
--- a/packages/core/src/user-store/postgres/store.ts
+++ b/packages/core/src/user-store/postgres/store.ts
@@ -70,7 +70,8 @@ export class PostgresUserStore implements UserStore {
           tx.schema
             .createView(model.name)
             .orReplace()
-            .as(tx.selectFrom(tableName).selectAll());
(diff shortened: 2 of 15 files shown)


kliyer-ai:
@0xOlias I just realized that I edited an existing migration, which of course makes no sense. Is it possible to do the type conversion of all the `bytea` columns to numeric in postgres? I'm not sure after trying for a bit... Alternative would be to drop all tables and recreate them which would result in data loss...

packages/core/src/user-store/sqlite/store.ts
```diff
@@ -149,6 +149,23 @@ export class SqliteUserStore implements UserStore {
...
+        this.schema!.entities.map(async (model) => {
+          const tableName = `${model.name}_${this.versionId}`;
+          tx.schema
+            .createView(model.name)
+            .orReplace()
+            .as(tx.selectFrom(tableName).selectAll())
+            .execute();
+        })
+      );
+    });
```
0xOlias, line 166: It seems like these views could be created just once during `reload`. Or am I missing something?

.changeset/tasty-planets-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
+use numeric32 for bigint on postgres instead of bytea
```
0xOlias, line 5: "Updated Postgres bigint columns to use `numeric32` rather than a custom byte encoding. This makes it easier to run queries directly against the indexed entity tables."

0xOlias (commented):
Awesome work, thanks a bunch for opening this. A few questions:

1. Does the `numeric(78)` representation handle the full range of EVM `int256` and `uint256` (specifically curious about signed integers)? Will `ORDER BY` a column typed this way using `ASC` or `DESC` behave as expected? I think the answers are yes, and we can add tests to confirm.
2. It would be great to make this same change for SQLite (and get rid of the janky custom encoding entirely). Doing some digging, it looks like SQLite's `REAL` type might work here. I can also take a crack at this if you'd like.
3. This change will be the first time we need a "data migration". Looks like Kysely [supports this out of the box](https://kysely.dev/docs/migrations#postgresql-migration-example:~:text=Migrations%20can%20also%20run%20normal%20queries%20to%20modify%20data.). So, we should add a new migration in `event-store/postgres/migrations` called something like `2023_09_08_0_convert_bigint_to_numeric78`. Unfortunately, because the custom encoding function can't be recreated in SQL (afaik) we can't do something like `alter table blocks alter column difficulty type numeric(78) using cast(difficulty as numeric(78))`. Let me know if you have any ideas here. One naive idea is to copy the data to a temporary column, update the original column, and drop the temp. Messy.

.changeset/tasty-planets-buy.md
kliyer-ai: Yeah this should be numeric78 instead of 32. I initially used numeric32 because I thought you specified the number of bytes but that's not true. It's the number of digits.

kliyer-ai:
> Awesome work, thanks a bunch for opening this. A few questions:
> 
> 1. Does the `numeric(78)` representation handle the full range of EVM `int256` and `uint256` (specifically curious about signed integers)? Will `ORDER BY` a column typed this way using `ASC` or `DESC` behave as expected? I think the answers are yes, and we can add tests to confirm.
> 2. It would be great to make this same change for SQLite (and get rid of the janky custom encoding entirely). Doing some digging, it looks like SQLite's `REAL` type might work here. I can also take a crack at this if you'd like.
> 3. This change will be the first time we need a "data migration". Looks like Kysely [supports this out of the box](https://kysely.dev/docs/migrations#postgresql-migration-example:~:text=Migrations%20can%20also%20run%20normal%20queries%20to%20modify%20data.). So, we should add a new migration in `event-store/postgres/migrations` called something like `2023_09_08_0_convert_bigint_to_numeric78`. Unfortunately, because the custom encoding function can't be recreated in SQL (afaik) we can't do something like `alter table blocks alter column difficulty type numeric(78) using cast(difficulty as numeric(78))`. Let me know if you have any ideas here. One naive idea is to copy the data to a temporary column, update the original column, and drop the temp. Messy.

1. Yes but let's add some tests.
2. Would be great if you could take a look. I'm no expert in SQLite but we can try `REAL`. It should be pretty much analogous to the Postgres changes then.
3. Yes you're right, we need a data migration. I wasn't able to do this conversion directly in SQL, so we will have to do it in the migration handler.

packages/core/src/user-store/postgres/store.ts
```diff
@@ -61,6 +60,23 @@ export class PostgresUserStore implements UserStore {
...
+  linkToCurrent = async () => {
+    if (!this.schema) return;
+
+    await this.db.transaction().execute(async (tx) => {
+      await Promise.all(
+        this.schema!.entities.map(async (model) => {
+          const tableName = `${model.name}_${this.versionId}`;
+          tx.schema
+            .createView(model.name)
+            .orReplace()
```
kliyer-ai, line 72: Replacing a view breaks if the schema changes, e.g. we added a new column to the table. It's safer to delete a view and recreate them instead of replacing.

0xOlias:
Closing; will resolve via #370 and #381

PR #355, opened 2023-09-24
Replace rpcUrl with viem transport in configuration options.

arberx:
### Description 

Addresses: https://github.com/0xOlias/ponder/issues/325 by replacing `rpcUrl` with a `viem` transport in the configuration options.

**New behavior:** Since `viem` transports default to a public RPC depending on the chain, the user doesn't actually have to specify anything in the transport option. 

### Todo

- [x] Add test


packages/core/src/config/config.ts
```diff
@@ -26,6 +27,10 @@ export type ResolvedConfig = {
     chainId: number;
     /** RPC URL. Default: if available, a public RPC provider. */
     rpcUrl?: string;
+    /** Custom viem transport: https://viem.sh/docs/clients/transports/http.html
+     * If `undefined`, a default transport will be used based on `rpcUrl`.
+     */
+    transport?: Transport;
```
0xOlias, line 33: If we go with a "one or the other" approach here, you can use [the `field?: never` trick](https://github.com/wagmi-dev/viem/blob/main/src/errors/base.ts#L3-L16) to produce a TS error if the user tries to provide both options.

0xOlias (commented):
Welcome and thanks for opening this! The major question I have here is whether it actually makes sense to replace the `rpcUrl` option with `transport` entirely.

Pros (of replacing `rpcUrl` with `transport`):
- Smaller/simpler public API (big benefit IMO)

Cons:
- Breaking change (not a big deal though, will be immediately apparent thanks to TS errors in `ponder.config.ts`)
- Introduces footguns, users could pass a poorly configured transport
- Potentially more confusing for beginners - you can use the `rpcUrl` option without knowing anything about `viem`.

Here's the diff that would go in the changelog:
```diff
+ import { http } from "viem";

export const config: Config = {
  networks: [
    {
      name: "mainnet",
      chainId: 1,
-     rpcUrl: process.env.PONDER_RPC_URL_1,
+     transport: http(process.env.PONDER_RPC_URL_1),
    },
  ],
  // ...
}
```

What do you think?

0xOlias:
There was consensus in the chat that we should just replace `rpcUrl` with `transport`.

Questions/todo:
1. It's probably worth trying out the websocket and fallback transports to make sure they work.
2. Now that `viem` is required (to import `http` or another transport), do we need to add it to package.json `peerDependencies`? We currently automatically add it in `devDependencies` because we use some types, but the transport is runtime code.
3. We'll need to update the `examples/` projects, the docs, some of the tests, and the `create-ponder` templates to reflect this change

packages/core/src/Ponder.ts
```diff
@@ -174,24 +174,6 @@ export class Ponder {
 
     this.registerServiceDependencies();
 
-    // If any of the provided networks do not have a valid RPC url,
-    // kill the app here. This happens here rather than in the constructor because
-    // `ponder codegen` should still be able to if an RPC url is missing. In fact,
-    // that is part of the happy path for `create-ponder`.
-    const networksMissingRpcUrl: Network[] = [];
```
arberx, line 181: This is unnecessary since the `viem` transport will default to using a public RPC if not specified.

0xOlias (commented):
Added a few commits:
- Made `transport` required (`rpcUrl` was optional to allow people to use the public/default RPC. Now you can do this with `http()`.)
- Created a function to get all RPC URLs for a viem Client, and a function to check if an RPC URL is in the list of public/default RPC URLs from `viem/chains`
- Added a warning log if the user is using a public RPC URL
- Small refactor in Ponder.ts and nits

FYI - soon, we're going to update the config file loading + validation codepath to make the config file hot-reloadable.

Last thing left to do is add a changeset. Other than that, LGTM!

packages/core/src/Ponder.ts
```diff
@@ -164,34 +174,8 @@ export class Ponder {
...
-        networksMissingRpcUrl.push(network);
-      }
-    });
-    if (networksMissingRpcUrl.length > 0) {
-      return new Error(
-        `missing RPC URL for networks (${networksMissingRpcUrl.map(
-          (n) => `"${n.name}"`
-        )}). Did you forget to add an RPC URL in .env.local?`
-      );
-    }
```
0xOlias, line 193: This is effectively replaced by a warning log that looks like:
```
WARN  config     Using public RPC URL for network "arbitrum". Ponder requires an RPC URL with a higher rate limit.
```

packages/core/src/Ponder.ts
```diff
@@ -206,32 +190,20 @@ export class Ponder {
...
     });
 
-    if (setupError) {
-      this.common.logger.error({
-        service: "app",
-        msg: setupError.message,
-        error: setupError,
-      });
-      return await this.kill();
-    }
```
0xOlias, line 233: Nice to get rid of this, it was always a hack. We're slowly adding more granular validation for each component separately.

PR #359, opened 2023-09-27
Respond with 503 while historical sync is completing

bankisan:


0xOlias (commented):
Looks great! Pasting from our chat on the proposed updates:

1) We add a new graphqlHTTP endpoint (same schema/resolvers) at `/graphql`, and return errors if the server is not healthy. This behaves the same in dev and prod.
2) For now, we leave the existing endpoint at / as it is. No breaking change in this PR, but in the changeset, urge users to update their client apps to use the `/graphql` path.
4) In a future breaking release, we replace `/` with the dev UI, and no longer serve GraphQL queries at that endpoint.

Issue #347, opened 2023-09-14
[Bug] Requests made before sync happens do not result in an error


Willyham:
Hey, we encountered an issue where Ponder restarted, and as it was coming back online, it had some issues syncing due to instability with our RPC provider.

```
> start
> ponder start
22:09:28.477 DEBUG app        Started using config file: ponder.config.ts
22:09:28.484 INFO  server     Started listening on port 6086
22:09:29.008 DEBUG codegen    Wrote new file at generated/index.ts
22:09:29.352 DEBUG codegen    Wrote new file at generated/index.ts
22:09:29.392 DEBUG codegen    Wrote new file at generated/schema.graphql
22:09:29.418 DEBUG handlers   Paused event queue (versionId=undefined)
22:09:29.591 DEBUG handlers   Reset user store (versionId=e601b59f)
22:09:35.110 INFO  realtime   Fetched latest block at 9689903 (network=goerli)
22:09:35.295 INFO  historical Started sync with 100% cached (logFilter=Staking network=goerli)
22:09:35.421 INFO  historical Started sync with 100% cached (logFilter=Oracle network=goerli)
22:09:36.176 INFO  realtime   Fetched finalized block at 9689871 (network=goerli)
22:09:44.216 ERROR historical Log sync task failed (network=goerli, logFilter=Staking)
HttpRequestError: HTTP request failed.
Status: 503
URL: https://eth-goerli.g.alchemy.com/v2/foo
Request body: {"method":"eth_getLogs","params":[{"address":"0x72e6a6d798be85157f4ec8d71444cdd2a7612d3d","fromBlock":"0x93daf4","toBlock":"0x93db0f"}]}
Details: {"code":-32000,"message":"Unable to complete request at this time."}
Version: viem@1.5.4
```

This is expected, and ponder did the correct thing of retrying until it was able to sync ✨. However, in the period between `22:09:36` where the finalized block was loaded, and `22:09:44` where the sync failed, ponder was serving results with incorrect values.

One of our services was made a request to load events at `22:09:37`, and ponder responded with `0` events found. This threw off our service as it was an unexpected result. 

I think the correct behaviour in this case is to always return an error until the sync has completed, as the events are unknown rather than known to be empty.

0xOlias:
Thanks for the well-researched bug report @Willyham!

[...]

PR #360, opened 2023-09-27
Respond with 503 while historical sync is completing

bankisan:
**Changes**
- Creates new `/graphql` endpoint with note to deprecate root `/`
- On prod, responds with `503` while waiting for historical sync to complete. On dev, responds with graphql ui.
- `shouldWaitForHistoricalSync` option

---
Fixes: #347

### DIFF
--- a/packages/core/src/server/service.test.ts
+++ b/packages/core/src/server/service.test.ts
@@ -57,15 +57,25 @@ const graphqlSchema = buildGqlSchema(schema);
 const setup = async ({
   common,
   userStore,
+  options = {
+    hasCompletedHistoricalSync: true,
+  },
 }: {
   common: Common;
   userStore: UserStore;
+  options?: {
+    hasCompletedHistoricalSync?: boolean;
+  };
 }) => {
   await userStore.reload({ schema });
 
   const service = new ServerService({ common, userStore });
   await service.start();
   service.reload({ graphqlSchema });
+
+  if (options.hasCompletedHistoricalSync) {
+    service.setIsHistoricalEventProcessingComplete();
+  }
 
   const gql = async (query: string) =>
     request(service.app)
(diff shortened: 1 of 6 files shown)


0xOlias (commented):
So, on review I'm realizing that the implementation here is slightly different from what I had in mind (emphasis on the difference):
1. Add a new GraphQL endpoint (same schema/resolvers) at `/graphql` that returns errors if the server is not healthy. **This new endpoint behaves the same in dev and prod.**
2. For now, we leave the existing route at `/` as it is. So, no breaking change in this PR, but in the changeset, urge users to update their client apps to use the /graphql path.
3. In a future breaking release, we'll replace `/` with the dev UI, and no longer serve GraphQL queries at that endpoint at all.

This would accomplish two goals:
- People can visit `/` during development to test out queries _even if historical indexing is still in progress_. In a future release, we'll replace the GraphiQL interface here with a full dev UI, which will be an even better debugging experience (perhaps it will include GraphiQL as one of many tabs).
- The new endpoint at `/graphql`, which is meant to be consumed by client applications, behaves the same way in dev and prod - it returns errors until healthy.

That said, I'm still open to the idea that the new endpoint at `/graphql` should return incomplete data when ran with `ponder dev`, so that folks can test/debug via their client application in development. Wdyt? cc @Willyham

Willyham:
I think the upgrade path sounds really sensible, and I really like the idea of distinguishing between the devUI and the actual query endpoint. 

> That said, I'm still open to the idea that the new endpoint at /graphql should return incomplete data when ran with ponder dev, so that folks can test/debug via their client application in development. Wdyt? cc @Willyham

My feeling is that dev and prod behaviour should stay as closely aligned as possible. I think one of ponder's biggest strengths compared to other products is that you can run the system locally - that benefit is diluted in my mind if what you're running isn't truly representational of the final product. I also think that in reality environment lines are increasingly blurred. What would we do for staging? Preview environments? Automated tests?

0xOlias:
> I think one of ponder's biggest strengths compared to other products is that you can run the system locally - that benefit is diluted in my mind if what you're running isn't truly representational of the final product.

Agree! Let's go with the approach above (new endpoint works the same in dev and prod), and see if folks complain about it in dev. Another thing to note is that historical indexing taking a while is a "bug" - our goal should be to make it as fast as possible via caching, concurrent indexing, or tricks (only index a subset of events). That's a much bigger topic, but wanted to mention it.

packages/core/src/server/service.test.ts
```diff
@@ -57,19 +57,29 @@ const graphqlSchema = buildGqlSchema(schema);
...
+  };
 }) => {
   await userStore.reload({ schema });
 
   const service = new ServerService({ common, userStore });
   await service.start();
   service.reload({ graphqlSchema });
 
+  if (options.hasCompletedHistoricalSync) {
+    service.setIsHistoricalEventProcessingComplete();
```
0xOlias, line 77: Quick note on naming. At the moment, some of the terminology in the codebase is not internally consistent. I recently came to the conclusion that there are two components: "sync" and "indexing", which are relatively clear in the [architecture diagram](https://link.excalidraw.com/l/9zcSSq4IYVb/2nIu3McwAQP).

Specific to this PR, it'd be nice to rename "HistoricalEventProcessing" -> "HistoricalIndexing" (a find and replace across the repo should work). And, you should rename the option here to match (hasCompletedHistoricalIndexing).

Soon we'll do a wholesale renaming PR. This will also change the public terminology: "event handlers" -> "indexing functions".

packages/core/src/server/service.ts
```diff
@@ -147,6 +148,29 @@ export class ServerService {
...
+          ),
+        ];
+        const result: FormattedExecutionResult = {
+          data: undefined,
+          errors,
+        };
+        return res.status(503).json(result);
+      }
+
+      return graphqlMiddleware(req, res);
```
0xOlias, line 169: Tested this locally. Looks great!

packages/core/src/server/service.ts
```diff
@@ -147,6 +148,29 @@ export class ServerService {
       graphiql: true,
     });
 
+    this.app?.use("/graphql", (req, res) => {
+      // While waiting for historical event processing to complete, we want to respond back
```
0xOlias, line 152: "historical indexing" per previous comment

0xOlias (approved):
LGTM

Issue #356, opened 2023-09-26
[Feature] Support for factory contracts


0xOlias:
See ongoing discussion: https://github.com/0xOlias/ponder/discussions/264

0xOlias:
Shipped in #361

Issue #343, opened 2023-09-12
[Feature] Improve internal cache design


0xOlias:
As we add more complex features like factory contracts + transaction call indexing, we should rethink how we keep track of what raw blockchain data gets cached during the sync process.

This will involve a bit of research and cuts close to the core of Ponder's sync engine.

See https://github.com/0xOlias/ponder/discussions/332 for more details.

0xOlias:
Shipped in https://github.com/0xOlias/ponder/pull/361

Issue #334, opened 2023-09-03
[Feature] Encode BigInt as postgres numeric data type


kliyer-ai:
This custom encoding that is currently used for `BigInt` makes working directly with the database a very painful experience.

How it looks:
![image](https://github.com/0xOlias/ponder/assets/27199575/45ef9098-db8c-4079-afb9-60fd584a1e9e)

How it should look:
![image](https://github.com/0xOlias/ponder/assets/27199575/24ef2b4c-9688-4a28-8fc2-cd017ae76876)

Numeric docs: https://www.postgresql.org/docs/current/datatype-numeric.html

0xOlias:
Fixed in #361. Note that direct database access is not done yet - still need to create the views.

PR #361, opened 2023-09-27
Factory contracts and sync engine upgrades

0xOlias:
This PR introduces factory contract support and makes some improvements to the sync cache design.

### WARNING
- Existing Ponder apps that upgrade to this version will need to re-sync their event store from scratch. Large apps that have a long sync time should consider testing the re-sync on a new branch/railway project before upgrading on main.

### How to test
1. To create a new project on the prerelease version -> `pnpm create ponder@next`
2. To upgrade an existing app -> `pnpm upgrade @ponder/core@next` (be warned, this will bust your local sync cache)
3. Open up the [new factory contracts documentation](https://ponder-docs-git-typedarray-sync-cache-upgrades-cantrip.vercel.app/advanced/factory-contracts), and try to sync/index your favorite factory contract!

### Fixes
- #356 
- #343 
- #334


shrugs:
i imagine you've thought of this but i think it's be very useful for the handlers of a child contract to receive a contract in the context that's bound to the address, rather than accessing `event.log.address`

it may also be useful to give handlers some sort of `setup` handler for when a new child contract is added, for example to ensure that there's an appropriate entity representing that new contract in the database

```ts
// LSSVMPair.ts
ponder.on("new", async ({ event, context }) => {
  const { Pair, TokenContract } = context.entities;
  const { LSSVMPair } = context.contracts;

  const [owner, nft, fee, spotPrice, delta] = await Promise.all([
    LSSVMPair.read.owner(),
    LSSVMPair.read.nft(),
    LSSVMPair.read.fee(),
    LSSVMPair.read.spotPrice(),
    LSSVMPair.read.delta(),
  ]);

  await Promise.all([
    Pair.upsert({
      id: addressId(LSSVMPair.address),
      create: {
        fee,
        spotPrice,
        delta,
        nft: addressId(nft),
        owner: addressId(owner),
      },
      update: {},
    }),
    TokenContract.upsert({
      id: addressId(nft),
      create: { },
      update: {},
    }),
  ]);
});
```

0xOlias:
> it may also be useful to give handlers some sort of `setup` handler for when a new child contract is added, for example to ensure that there's an appropriate entity representing that new contract in the database

Just seeing this. Awesome idea. I wonder what the event should be called? Your idea seemed to suggest that we would infer the contract name from the file name (I think) which is not currently a pattern. The only "special" event name we have currently is "setup" - all others are `${sourceName}:${eventName}`. 

Rambling a bit now, but soon enough we'll also want to add transaction call events, which could look like `"UniswapPool.swap"` (using `.` as the separator instead of `:`).

In the case of this "child contract created" event, maybe we could use `_` as the separator (`"UniswapPool_created"`. Don't like that very much though.

shrugs:
ah, didn't mean to imply file name detection, i just forgot to include the appropriate information in the event name. perhaps "::" as like a "class level" delimiter could work, but perhaps too easily confused with ":" for events. @ or | also exist but i'm less enthused by then

perhaps "setup:ContractName"? though setup maybe not be the right word for this

packages/core/src/utils/lowercase.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Transforms the input string to lower case.
+ */
+export function toLowerCase<T extends string>(value: T) {
+  return value.toLowerCase() as T;
```
kyscott18, line 5: There's a cool generic in typescript called Lowercase that you could use here. 

```ts
return value.toLowerCase() as Lowercase<T>;
```

0xOlias: Just pushed a commit that uses `Lowercase<T>` and updates those spots to use the util - thanks!

shrugs:
oh and re: setup event, this could just be typed as a handler of the factory event itself, so perhaps to support this pattern one adds the factory itself as a contract, listens to the correct event, and uses that as the setup? this requires that the factory's handler runs before any handlers in the contract, which i imagine is already the case. i'll test it out

shrugs:
bug report on `0.0.93-next.1`: multiple factory configs using the same event but with different parameters doesn't seem to work in my case— only the first contract definition has its handlers invoked. the llama example works for me out of the box

```ts
const event = parseAbiItem("event NewNftaddress pair, address nft)");

export const config: Config = {
  networks: [network],
  contracts: [
    {
      network: network.name,
      name: "LSSVMPair",
      abi: "./abis/LSSVMPair.json",
      startBlock: factoryConfig.startBlock,
      factory: {
        address: factoryConfig.address,
        event,
        parameter: "pair",
      },
    },
    {
      network: network.name,
      name: "NFT",
      abi: "./abis/NFT.json",
      startBlock: factoryConfig.startBlock,
      factory: {
        address: factoryConfig.address,
        event,
        parameter: "nft",
      },
    },
  ],
};

```

if i change the order of the contracts, it's always the first one that has its handlers invoked (i'm destroyng the ponder cache between runs)

0xOlias (commented):
LGTM
obama-giving-obama-award

PR #362, opened 2023-09-29
Create many overflow

### LINKED ISSUE #336: [Bug] `createMany` fails with too many variables
> found a downside to createMany —  it can fail with "too many SQL variables" if there are too many (not sure what the number is) datas. the Promise.all structure doesn't have that limitation because it generates a new sql query per entity

kyscott18:
Fixes #336 . Batch createMany() internally to a maximum size of 1000. This fixes a bug where adding too many values with `createMany()` fails. This was between 6000-7000 for me.

### DIFF
--- a/packages/core/src/user-store/postgres/store.ts
+++ b/packages/core/src/user-store/postgres/store.ts
@@ -538,15 +538,21 @@ export class PostgresUserStore implements UserStore {
       effectiveTo: MAX_INTEGER,
     }));
 
-    const instances = await this.db
-      .insertInto(tableName)
-      .values(createInstances)
-      .returningAll()
-      .execute();
+    const chunkSize = 1000;
 
-    return instances.map((instance) =>
-      this.deserializeInstance({ modelName, instance })
+    const chunkedInstances = [];
+    for (let i = 0, len = createInstances.length; i < len; i += chunkSize)
+      chunkedInstances.push(createInstances.slice(i, i + chunkSize));
+
+    const instances = await Promise.all(
+      chunkedInstances.map((c) =>
+        this.db.insertInto(tableName).values(c).returningAll().execute()
+      )
     );
+
+    return instances
+      .flat()
+      .map((instance) => this.deserializeInstance({ modelName, instance }));
   };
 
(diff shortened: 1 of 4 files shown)


packages/core/src/user-store/postgres/store.ts
```diff
@@ -538,15 +538,21 @@ export class PostgresUserStore implements UserStore {
...
-      .insertInto(tableName)
-      .values(createInstances)
-      .returningAll()
-      .execute();
+    const chunkSize = 1000;
 
-    return instances.map((instance) =>
-      this.deserializeInstance({ modelName, instance })
+    const chunkedInstances = [];
+    for (let i = 0, len = createInstances.length; i < len; i += chunkSize)
```
0xOlias, line 544: Gigabrain for loop

packages/core/src/user-store/postgres/store.ts
```diff
@@ -538,15 +538,21 @@ export class PostgresUserStore implements UserStore {
       effectiveTo: MAX_INTEGER,
     }));
 
-    const instances = await this.db
-      .insertInto(tableName)
-      .values(createInstances)
-      .returningAll()
-      .execute();
+    const chunkSize = 1000;
```
0xOlias, line 541: I would probably call this `MAX_BATCH_SIZE` and move it to the top of the file alongside `MAX_INTEGER`. Same goes for the SQLite store.

packages/core/src/user-store/store.test.ts
```diff
@@ -840,6 +840,29 @@ test("createMany() inserts multiple entities", async (context) => {
   await userStore.teardown();
 });
 
+test("createMany() inserts a large number of entities", async (context) => {
+  const { userStore } = context;
+  await userStore.reload({ schema });
+
+  const num = 100_000;
```
0xOlias, line 847: Would consider calling this `ENTITY_COUNT` or something like that.

0xOlias (approved):
Nits. LGTM!

0xOlias (approved):

PR #366, opened 2023-10-03
update eslint vscode settings

o-az:
small tweaks

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,3 +3,5 @@ LICENSE
 **/generated
 **/.next
 **/dist
+.gitignore
+.prettierignore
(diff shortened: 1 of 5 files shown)


.prettierignore
```diff
@@ -2,4 +2,6 @@ pnpm-lock.yaml
 LICENSE
 **/generated
 **/.next
-**/dist
\ No newline at end of file
+**/dist
+.gitignore
```
0xOlias, line 6: Noticed this annoyance locally also

.vscode/settings.json
```diff
@@ -1,13 +1,23 @@
 {
-  "editor.defaultFormatter": "rvest.vs-code-prettier-eslint",
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
```
0xOlias, line 2: Did you notice any differences after switching this?

0xOlias (approved):
LGTM

.vscode/settings.json
o-az: Yeah not having to have that extra so I was able to uninstall it and keep prettier&eslint exts

PR #367, opened 2023-10-03
swap express-graphql to graphql-http

o-az:
~CI won't pass until not wip~

- Replace `express-graphql` with `graphql-http`,
- Add `GraphiQL` playground (since `graphql-http` is pretty light so no playground included)


packages/core/src/user-store/utils/pagination.ts
```diff
@@ -1,8 +1,12 @@
+import { createGraphQLError } from "graphql-yoga";
+
 const MAX_SKIP = 5000;
 
 export function validateSkip(skip: number) {
   if (skip > MAX_SKIP) {
-    throw new Error("Invalid query. Cannot skip more than 5000 rows.");
+    throw createGraphQLError(
```
o-az, line 7: this fixes this error:

https://github.com/0xOlias/ponder/actions/runs/6387778466/job/17336588381?pr=367#step:5:54

packages/core/src/_test/art-gobblers/app.test.ts
```diff
@@ -53,7 +53,7 @@ const setup = async ({ context }: { context: TestContext }) => {
 
   const gql = async (query: string) => {
     const response = await request(ponder.serverService.app)
-      .post("/")
+      .post("/graphql")
```
o-az, line 56: fixes this:

https://github.com/0xOlias/ponder/actions/runs/6388656128/job/17338850870?pr=367#step:5:87

packages/core/src/_test/ens/app.test.ts
```diff
@@ -53,7 +53,7 @@ const setup = async ({ context }: { context: TestContext }) => {
 
   const gql = async (query: string) => {
     const response = await request(ponder.serverService.app)
-      .post("/")
+      .post("/graphql")
```
o-az, line 56: fixes this:

https://github.com/0xOlias/ponder/actions/runs/6388656128/job/17338850870?pr=367#step:5:103

packages/core/src/server/service.test.ts
```diff
@@ -1270,9 +1270,9 @@ test("throws if limit is greater than 1000", async (context) => {
   `);
 
   expect(response.body.errors[0].message).toBe(
-    "Invalid query. Cannot take more than 1000 rows."
+    "Invalid query. Cannot take more than 1000 rows. Received: 1005 rows."
   );
-  expect(response.statusCode).toBe(500);
+  expect(response.statusCode).toBe(200);
```
o-az, line 1275: @0xOlias After some thinking, I updated the expected status code here from `500` to `200`.

Since the expected error in this test is due to incorrect/bad input, this is more of a bad request than a [server error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500), so it makes sense that Yoga is returning status code `200` for this request.

packages/core/src/server/service.test.ts
```diff
@@ -1318,9 +1318,9 @@ test("throws if skip is greater than 5000", async (context) => {
   `);
 
   expect(response.body.errors[0].message).toBe(
-    "Invalid query. Cannot skip more than 5000 rows."
+    "Invalid query. Cannot skip more than 5000 rows. Received: 5005 rows."
   );
-  expect(response.statusCode).toBe(500);
+  expect(response.statusCode).toBe(200);
```
o-az, line 1323: same comment as https://github.com/0xOlias/ponder/pull/367/files#r1343497742

packages/core/src/user-store/utils/pagination.ts
o-az: got this from here: https://github.com/dotansimha/graphql-yoga/blob/0f1483c939ccb76f5e3c809c0489fa6cc9aba864/examples/error-handling/src/yoga.ts#L43-L51

Without it, the error doesn't include the message. The error message is [`"Unexpected error."`](https://github.com/0xOlias/ponder/actions/runs/6387778466/job/17336588381?pr=367#step:5:116)

package.json
```diff
@@ -9,7 +9,7 @@
     "install:packages": "pnpm --filter \"./packages/**\" install",
     "install:examples": "pnpm --filter \"./examples/**\" install",
     "lint": "eslint --cache .",
-    "lint:fix": "pnpm lint --fix",
+    "lint:fix": "pnpm lint --fix --ignore-path .gitignore --ignore-path .prettierignore",
```
0xOlias, line 12: Should this be added to eslint config rather than inline?

o-az:
As it is right now graphiql is quite fast (thanks in part to the unpkg.com cdn).
So keep it as it is right now then later when building the dashboard consider rendering graphqil as ssr instead of serving from cdn? čč @0xOlias

packages/core/src/_test/art-gobblers/app.test.ts
0xOlias, line 56: Good call, we should have made this change earlier.

packages/core/src/historical-sync/service.test.ts
```diff
@@ -224,43 +224,50 @@ test("setup() updates cached block, total block, and scheduled task metrics", as
...
 
-  await service.kill();
-});
+    await service.kill();
+  },
+  {
+    // default is 5000
+    timeout: 7_000,
+  }
+);
```
0xOlias, line 270: Did this time out locally or in CI? Would prefer to make this change in another PR, could you delete the changes in this file?

packages/core/src/server/service.test.ts
0xOlias: Makes perfect sense. OOC, what does the GraphiQL interface look like if you run a request that exceeds the limit?

packages/core/src/server/service.ts
```diff
@@ -164,12 +163,17 @@ export class ServerService {
...
 
     // NOTE: Deprecating use of root endpoint for GraphQL queries in favor of /graphql.
     // This will be removed in a future release.
-    this.app?.use("/", graphqlMiddleware);
+    this.app?.use("/", (_, response) => {
+      response.setHeader("Content-Type", "text/html");
+      response.send(
+        graphiQLPage({ endpoint: "http://localhost:42069/graphql" })
+      );
+    });
```
0xOlias, line 176: See [this PR for context](https://github.com/0xOlias/ponder/pull/360).

Given that context, I think what we actually need to do here is update BOTH the `/` and `/graphql` endpoints to serve GET requests with GraphiQL and POST requests using the `graphql-http` middleware. The only difference is that `/graphql` returns a `503` for both GET and POST requests until historical indexing is complete.

0xOlias (commented):
Few quick questions

package.json
o-az: done

packages/core/src/server/service.ts
o-az: Oh yeah my plan was to refactor this in a separate PR and basically handle the comment right above this. So GET goes to playground, POST is request, etc. So I plan to do that right after this in another pr if that works.

packages/core/src/historical-sync/service.test.ts
o-az: only in my machine it times out :/

o-az:
@0xOlias answer to https://github.com/0xOlias/ponder/pull/367?notification_referrer_id=NT_kwDOAWhjf7M3OTIwNTg0MTA3OjIzNjE4NDMx#discussion_r1344795840

I haven't tried with the same exact query in that test but I tried in `ethfs`'s playground  
CleanShot 2023-10-03 at 15 22 00@2x

Then when > 1k

CleanShot 2023-10-03 at 15 23 41@2x

packages/core/src/server/service.test.ts
o-az: I haven't tried with the same exact query in that test but I tried in `ethfs`'s playground  
CleanShot 2023-10-03 at 15 22 00@2x

Then when > 1k

CleanShot 2023-10-03 at 15 23 41@2x

packages/core/src/historical-sync/service.test.ts
o-az: all changes to this file removed

0xOlias (approved):
LGTM

PR #369, opened 2023-10-03
upgrading a number stale dependencies

o-az:
Upgrading deps


examples/with-docker/.eslintrc.json
```diff
@@ -1,3 +1,4 @@
 {
+  "root": true,
```
0xOlias, line 2: What does this do?

scripts/clean.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
0xOlias, line 1: What's the purpose of this file?

.prettierrc.cjs
```diff
@@ -0,0 +1,7 @@
+/** @type {import('prettier').Config} */
```
0xOlias, line 1: Let's use whatever settings don't lead to a massive diff when ran, lol. This looks good to me.

scripts/clean.sh
o-az: clean everything in one command. I can git ignore the directory if you don't see it useful

examples/with-docker/.eslintrc.json
o-az: > The `"root"` field in an ESLint configuration file is a boolean that tells ESLint whether the configuration file is the root or not.
>
> When ESLint runs, it starts from the directory where it was called and searches upwards for configuration files. However, once it finds a configuration file with `"root": true`, it stops looking for other configuration files.
>
> Here's what it does:
>
> - If `"root": true` is in a configuration file, ESLint won't search for any other configuration files in parent folders.
> - If `"root": false` or the `"root"` key is not present, ESLint will continue searching upwards through the directory structure looking for more configuration files.
>
> This can be useful in monorepo setups or when you have different ESLint configurations for different parts of your project. By setting `"root": true` in a configuration file, you can ensure that ESLint uses only that configuration for the directory and its subdirectories.

each of those eslintrc files are the root eslints of their respective project. Without this field you get some errors with the up to date ESLint and ESLint typescript pkgs

.prettierrc.cjs
o-az: too much to type. Will discuss why this file is necessary in the call

scripts/clean.sh
0xOlias: Mostly curious what workflow inspired you to add it

o-az: Found myself needing to uninstall/reinstall deps, removing node_modules and `dist`s, etc. so this is a handy script I grabbed. The reason I'm doing uninstalls/reinstalled and all that stuff is cause I'm jumping between prs and deps are changing vastly across those prs. Should be temporary

PR #371, opened 2023-10-04
chore: Update GraphQL client & GraphiQL endpoints & methods

o-az:


packages/core/src/server/service.ts
```diff
@@ -142,12 +141,14 @@ export class ServerService {
   }
 
   reload({ graphqlSchema }: { graphqlSchema: GraphQLSchema }) {
+    const SERVER_URL = "http://localhost:42069";
```
0xOlias, line 144: This approach might be tricky - won't this break in non-local environments, or if the port is different?

packages/core/src/server/service.ts
```diff
@@ -160,19 +161,34 @@ export class ServerService {
...
-      return graphqlMiddleware(req, res, next);
+      if (request.method === "GET") {
+        response.setHeader("Content-Type", "text/html");
+        response.send(graphiQLPage({ endpoint: `${SERVER_URL}/graphql` }));
+        return next();
+      }
+
+      if (request.method === "POST") {
+        return graphqlMiddleware(request, response, next);
+      }
```
0xOlias, line 175: Can get rid of this right?

packages/core/src/server/service.ts
```diff
@@ -160,19 +161,34 @@ export class ServerService {
...
     // This will be removed in a future release.
-    this.app?.use("/", (_, response) => {
-      response.setHeader("Content-Type", "text/html");
-      response.send(
-        graphiQLPage({ endpoint: "http://localhost:42069/graphql" })
-      );
+    this.app?.use("/", (request, response, next) => {
+      if (request.method === "POST") {
+        return graphqlMiddleware(request, response, next);
+      }
```
0xOlias, line 185: Same as above

0xOlias (commented):
Couple questions:
1. Any idea how difficult it would be to server render the HTML rather than using CDN scripts?
2. As I mentioned in comment, I think we'll need to determine the hostname somehow and embed that in the HTML rather than hardcode. Not really sure how.

packages/core/src/server/service.ts
o-az: it for sure will. What to use instead?

o-az: now passing dynamic server url

o-az:
@0xOlias 
> Couple questions:
> 
> 1. Any idea how difficult it would be to server render the HTML rather than using CDN scripts?

I've only ever used GraphiQL through CDN though I think it should be doable, you already have React as a dep so we can use that. The performance gain is definitely negligible and https://npm.im/graphiql is huuge 

> 2. As I mentioned in comment, I think we'll need to determine the hostname somehow and embed that in the HTML rather than hardcode. Not really sure how.

done

packages/core/src/server/service.ts
```diff
@@ -160,19 +167,46 @@ export class ServerService {
...
+          return graphqlMiddleware(request, response, next);
+        case "GET": {
+          const SERVER_URL = this.#getOwnUrlFromRequest(request);
+          return response
+            .status(200)
+            .setHeader("Content-Type", "text/html")
+            .send(graphiQLHtml({ endpoint: `${SERVER_URL}/graphql` }));
+        }
+        case "HEAD":
+          return response.status(200).send();
```
0xOlias, line 184: Any reason to specify a behavior for `"HEAD"` here rather than just falling back to the default behavior offered by express (via `next()`) ?

packages/core/src/server/service.ts
```diff
@@ -28,10 +28,13 @@ export class ServerService {
     this.port = this.common.options.port;
   }
 
+  #getOwnUrlFromRequest(request: express.Request) {
+    return `${request.protocol}://${request.get("host")}`;
+  }
```
0xOlias, line 33: This is a style thing, but TBH I'd probably inline this in both places rather than create a method.

o-az: done

packages/core/src/server/service.ts
o-az: `HEAD` request errors out when it's left out in both `/` and `/graphql` but not in the others (`/metrics`, etc.). I don't understand why but that's why I'm explicitly adding it here. My guess is it has something to do with `.use(...)` vs `.get`/`.<method>`? Not sure

0xOlias (approved):
LGTM

PR #377, opened 2023-10-05
convert codebase to ESM & add build `watch` support

o-az:
Convert to ESM & add watch support for build.

Converting to ESM required

- updating a number of `tsconfig` settings, including `moduleResolution` to be `Bundler`,
- postfixing local import with `.js` (in spite of [this](https://www.typescriptlang.org/tsconfig#moduleResolution)),
- replacing `__dirname` with ESM-safe option,
- leveraging `createRequire` for some imports,
- updating `type` to `module`,
- updating some dependencies to latest (each upgraded dependency had to be upgraded for this to work),
- small change to logic in `core/src/build/handlers.ts#buildRawHandlerFunctions`,
- small change to logic in `core/src/config/config.ts#buildConfig`,
- small tweak to `vitest` config needed for 2 tests (ens & art-gobblers) specifically,
- [import assertions](https://v8.dev/features/import-assertions) for json imports,
- switching to default export for `pg` (and updating the dep),
- small updates to types here and there,
- other things I can't remember atm :>


o-az:
@0xOlias all tests pass except for ens and art-gobblers in core

packages/core/src/_test/ens/app/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import { ponder } from "@/generated";
+import { ponder } from "../generated/index.js";
```
0xOlias, line 1: Curious why this was necessary. Does the alias still work in the example projects?

packages/core/src/build/handlers.ts
```diff
@@ -1,15 +1,19 @@
...
-import type { Log } from "@/types/log";
-import type { Transaction } from "@/types/transaction";
+import type { LogEventMetadata, LogFilter } from "@/config/logFilters.js";
+import type { Options } from "@/config/options.js";
+import type { Block } from "@/types/block.js";
+import type { Log } from "@/types/log.js";
+import type { Transaction } from "@/types/transaction.js";
+import { isRejected } from "@/utils/promise.js";
+
+const require = createRequire(import.meta.url);
```
0xOlias, line 16: Did you test that hot reloading still works using createRequire with the require cache invalidation trick?

packages/core/src/build/handlers.ts
```diff
@@ -148,23 +152,18 @@ export const buildRawHandlerFunctions = async ({
...
-      } catch (err) {
-        return err as Error;
-      }
-    })
-    .filter((err): err is Error => err !== undefined);
+  const importUserFiles = await Promise.allSettled(
+    outUserFilenames.map((filename) => import(filename))
+  );
+
+  const rejecteds = importUserFiles.filter(isRejected);
```
0xOlias, line 159: This change to allSettled makes sense, though I'd probably call this `rejectedImports`.

0xOlias (commented):
Couple small questions. Seems like most of the diff is updating imports to use the `.js` prefix. Any other areas you were unsure about / I should take a close at?

packages/core/src/build/handlers.ts
o-az: yup first thing. Sadly doesn't work

packages/core/src/build/handlers.ts
o-az: done

packages/core/src/_test/ens/app/src/index.ts
o-az: Actually not necessary after I added the vitest resolve. So reverting

o-az: done

packages/core/src/build/handlers.ts
o-az: m8 be relevant https://github.com/evanw/esbuild/issues/1828

o-az:
> Couple small questions. Seems like most of the diff is updating imports to use the `.js` prefix. Any other areas you were unsure about / I should take a close at?

If you could search globally for "await import" and review those files that would be great.

Besides that majority of it is adding .js

PR #380, opened 2023-10-09
Use typescript to define schema

### LINKED ISSUE #348: [Feature] Use TypeScript for schema definition
We're going to move away from `schema.graphql` for the schema definition langauge in favor of a `schema.ts` file. This will improve developer experience in a few ways:
- Static validation of schema definition (see errors directly in your editor)
- Autocomplete (see API directly in your editor)
- Use TS language server to inject entity types from `schema.ts` -> `src/index.ts` handler function files. This replaces a codegen process that the TS lang server often ignores, leading to briefly "stale" entity types in handler functions.

kyscott18:
# TypeScript Schema

## Overview

Move ponder schema from being described with GraphQL to being described with a custom TypeScript format.

This will benefit Ponder in a few ways:

- No-codegen type inference
- Instant, type-level schema validation
- Describe schema much closer to how it is structured in the database

## Example

A port of the existing [friendtech schema](https://github.com/0xOlias/ponder/blob/main/examples/friendtech/schema.graphql), but using an enum for `isBuy`

```ts
import { p } from "@ponder/core";

export const schema = p.createSchema({
  TradeType: p.createEnum(["BUY", "SELL"]),
  Share: p.createTable({
    id: p.bytes(),
    subjectId: p.bytes().references("Subject.id"),
    traderId: p.bytes().references("Trader.id"),
    shareAmount: p.bigint(),
  }),
  TradeEvent: p.createTable({
    id: p.bytes(),
    subjectId: p.bytes().references("Subject.id"),
    traderId: p.bytes().references("Trader.id"),
    shareAmount: p.bigint(),
    tradeType: p.enum("TradeType"),
    ethAmount: p.bigint(),
    protocolEthAmount: p.bigint(),
    subjectEthAmount: p.bigint(),
    traderAmount: p.bigint(),
    supply: p.bigint(),
    timestamp: p.int(),
  }),
  Subject: p.createTable({
    id: p.bytes(),
    totalShares: p.bigint(),
    totalTrades: p.bigint(),
    lastPrice: p.bigint(),
    earnings: p.bigint(),
    traderVolume: p.bigint(),
    protocolFeesGenerated: p.bigint(),
    shares: p.virtual("Share.subjectId"),
    trades: p.virtual("TradeEvent.subjectId"),
  }),
  Trader: p.createTable({
    id: p.bytes(),
    totalTrades: p.bigint(),
    spend: p.bigint(),
    earnings: p.bigint(),
    profit: p.bigint(),
    subjectFeesPaid: p.bigint(),
    protocolFeesPaid: p.bigint(),
    shares: p.virtual("Share.traderId"),
    trades: p.virtual("TradeEvent.traderId"),
  }),
  Protocol: p.createTable({
    id: p.int(),
    earnings: p.bigint(),
  }),
});

```

### createSchema

```ts
p.createSchema({
  // schema definition here
})
```

### createTable

```ts
p.createTable({
  // table definition here
})
```

### createEnum

```ts
p.createEnum(["ONE", "TWO", "THREE"])
```

### Scalar values

The types available are:

- `int`
- `float`
- `string`
- `bytes`
- `bigint`
- `boolean`

```ts
p.float()
```

### Modifiers

Modifiers alter the scalar type. Currently there are flags for `optional` and `list`

```ts
p.string().optional();

p.string().list();
```

### References

References are used to link entities to one another. This is done by storing the `id` of one table in the column of another. 

This is done by specifying the table name, followed by `.id`.

```ts
p.string().references("OtherTable.id");
```

**`OtherTable` must have an `id` column of type "string"**

The name of this column must end with "Id". GraphQL internally removes these characters, so the queryable field will not contain the characters "Id" at the end.

### Virtual Columns

Virtual columns are used only for querying. They create a queryable field that selects all the columns from another table that reference the current table. In the example below, all the `OwnerId` columns in the `Pet` table that reference this table would be returned. This is equivalent in functionality to the @derivedFrom(field) tag in GraphQL.

```ts
p.virtual("Pet.OwnerId")
```

**Table `Pet` must exist with a column `OwnerId` that references the table this column is in**

### Enum Columns

Enums are used to represent a set of values. Once an Enum is defined, it can be used as a type in a column.

```ts
p.enum("enum");
```

Closes #348


packages/core/src/user-store/postgres/store.ts
```diff
@@ -92,57 +93,21 @@ export class PostgresUserStore implements UserStore {
...
-                  gqlScalarToSqlType[field.relatedEntityIdType.name],
-                  (col) => {
-                    if (field.notNull) col = col.notNull();
-                    return col;
-                  }
-                );
-                break;
-              }
+          Object.keys(model.columns).forEach((key) => {
+            if (!model.columns[key].list) {
```
0xOlias, line 97: In general, I find it more readable to handle a truthy case first (even if it's less common/important). If you use `Object.entries`, this would read:

```ts
if (column.list) {
  // Handle scalar list columns.
  ...
} else {
  // Handle all other column types.
  ...
}
```

Which I find quite clear (comments optional 😅).

0xOlias (commented):
Looking great, very excited about this. Added some small suggestions.

packages/core/src/server/graphql/entity.ts
```diff
@@ -25,119 +23,58 @@ export const buildEntityType = ({
...
-            break;
-          }
-          case "RELATIONSHIP": {
-            const resolver: GraphQLFieldResolver<Source, Context> = async (
-              parent,
-              _args,
-              context
-            ) => {
-              const { store } = context;
+      Object.keys(entity.columns).forEach((key) => {
```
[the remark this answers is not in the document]
kyscott18: Agreed this is better

packages/core/src/user-store/postgres/store.ts
```diff
@@ -21,13 +21,14 @@ import {
...
-  Bytes: "text",
-  Float: "text",
+// TODO: should we have two different types for number and float
+const scalarToSqlType = {
+  boolean: "integer",
+  int: "integer",
+  float: "text",
+  string: "text",
+  bigint: sql`bytea`,
+  bytes: "text",
```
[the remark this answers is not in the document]
kyscott18: I think this is a legacy comment from when I only had a "number" type, will delete

packages/core/src/server/graphql/plural.ts
```diff
@@ -24,6 +25,8 @@ type PluralArgs = {
 };
 type PluralResolver = GraphQLFieldResolver<Source, Context, PluralArgs>;
 
+const tsTypeToGqlField = (key: string) => (key === "bigint" ? "bigInt" : key);
```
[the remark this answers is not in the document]
kyscott18: Removing this changes the query schema text from `bigInt` to `bigint`. I added this so querying would be compatible with current queries.

packages/core/src/server/graphql/entity.ts
```diff
@@ -25,119 +23,58 @@ export const buildEntityType = ({
...
+            type: !entity.columns[key].optional
+              ? new GraphQLNonNull(tsTypeToGqlScalar[entity.columns[key].type])
+              : tsTypeToGqlScalar[entity.columns[key].type],
+            // Convert bigints to strings for GraphQL responses.
+            resolve:
+              entity.columns[key].type === "bigint"
+                ? // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                  // @ts-ignore
+                  (parent) => (parent[key] as bigint).toString()
+                : undefined,
```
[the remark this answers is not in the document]
kyscott18: Yeah I agree after looking more closely, let's come back once all tests pass without having to comment any out

packages/core/src/build/service.ts
```diff
@@ -128,7 +130,8 @@ export class BuildService extends Emittery<BuildServiceEvents> {
       const schema = await buildSchema({
         schemaFile: this.common.options.schemaFile,
       });
-      this.emit("newSchema", { schema });
+      const graphqlSchema = buildGqlSchema(schema);
```
0xOlias, line 133: I think it would be clearer if this happens elsewhere (ServerService, maybe?). The BuildService should only be responsible for building & validating the new "app schema". Constructing the GraphQL schema from the "app schema" should be downstream.

packages/core/src/codegen/service.ts
```diff
@@ -88,22 +86,4 @@ export class CodegenService extends Emittery {
       msg: `Wrote new file at generated/index.ts`,
     });
   }
-
-  generateSchemaFile({ schema }: { schema: Schema }) {
```
0xOlias: I hadn't thought about this, but I kinda agree that we don't need to write the `generated/schema.graphql` file. I thought that it would be useful for stuff like `graphql-codegen`, but I'm not sure if anyone's using it.

I polled the telegram.

kyscott18: From what I remember you can use` graphql-codegen` with just the endpoint. If that's not the case then I think we should include it because I think that is pretty useful.

packages/core/src/build/service.ts
kyscott18: Hmmm it kinda makes sense to me to keep if here but open to change it. This is also the way it is done currently

packages/core/src/schema/build.ts
```diff
@@ -0,0 +1,61 @@
+import { build } from "esbuild";
```
0xOlias, line 1: This is more of a code build/loading concern, so I think we should move this file to `build/schema.ts`.

packages/core/src/schema/build.ts
```diff
@@ -0,0 +1,61 @@
...
+      );
+    }
+
+    let resolvedSchema: Schema;
+
+    if (typeof rawSchema === "function") {
+      resolvedSchema = await rawSchema();
+    } else {
+      resolvedSchema = await rawSchema;
+    }
```
0xOlias, line 54: Guessing this is stale 🍝 - I don't think we would ever allow passing an async function as the schema, or a function at all for that matter.

packages/core/src/schema/types.ts
```diff
@@ -1,83 +1,162 @@
...
+        Column
+      >)
+    | unknown = unknown
+> = {
+  table: Table<TTableName, TColumns>;
+  addColumn: <
+    TTables extends readonly Table[],
+    TName extends string,
+    TType extends Scalar,
+    TReferneces extends "id" extends TName
```
0xOlias, line 81: Typo, `TReferences`

packages/core/src/schema/schema.ts
```diff
@@ -1,341 +1,212 @@
...
-// Find all scalar types in the schema that were created by the user.
-const getUserDefinedScalarTypes = (schema: GraphQLSchema) => {
-  return Object.values(schema.getTypeMap()).filter(
-    (type) =>
-      !!type.astNode &&
-      type.astNode.kind === Kind.SCALAR_TYPE_DEFINITION &&
-      !["BigInt", "Bytes"].includes(type.name)
-  ) as GraphQLScalarType[];
-};
+const verifyKey = (key: string) => {
```
0xOlias, line 207: Being pedantic, but this could be called `validateTableOrColumnName`

packages/core/src/schema/schema.ts
```diff
@@ -1,341 +1,212 @@
...
+    >
+  >[]
+>(
+  schema: TSchema
+): {
+  entities: { [key in keyof TSchema]: TSchema[key]["table"] };
+} => {
+  const tables = schema.map((it) => it.table);
+
+  tables.forEach((t) => {
```
0xOlias, line 141: Love this runtime validation. At some point we'll need to test that the errors are propagating properly when making changes with the dev server active. Lmk if you want help with this, can point you in the right direction.

packages/core/src/codegen/service.ts
```diff
@@ -87,25 +86,4 @@ export class CodegenService extends Emittery {
       msg: `Wrote new file at generated/index.ts`,
     });
   }
-
-  generateSchemaFile({ graphqlSchema }: { graphqlSchema: GraphQLSchema }) {
```
0xOlias, line 91: Per results of the poll, let's add this back in. We can rename it `generateGraphqlSchemaFile` to make it more clear.

Also with that in mind, perhaps it does make more sense to call `buildGraphqlSchema` within the BuildService and emit both `schema` and `graphqlSchema` in the `newSchema` event.

packages/core/src/schema/types.ts
kyscott18: Nice catch

packages/core/src/schema/build.ts
kyscott18: yeah whoops

packages/core/src/schema/build.ts
kyscott18: Okay, gonna move the config build too

packages/core/src/server/graphql/plural.ts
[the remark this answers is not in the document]
kyscott18: Nvm you were definitely right

packages/core/src/schema/schema.test.ts
```diff
@@ -1,452 +1,66 @@
...
+    name: createColumn("id", "bigint").addColumn("age", "enum:enummm"),
+  });
+});
+
+test("references", () => {
+  createSchema({
+    Person: createColumn("id", "string").addColumn("age", "int"),
+    Dog: createColumn("id", "string").addColumn("owner", "string", {
+      references: "Person.id",
+    }),
```
0xOlias, line 64: Ooh, I quite like this style (use object properties as table/enum names). But, it feels weird that `Dog`, which is the name of a table, has `createColumn` next to it. I can see why, but it's still kinda weird.

Just spitballing again, but what do you think about this:

```ts
import { createSchema, createTable, createColumn } from "@ponder/core"

const schema = createSchema({
  Person: createTable({
    id: createColumn("string"),
    age: createColumn("int"),
  }),
  Dog: createTable({
    id: createColumn("string"),
    ownerId: createColumn("string", { references: "Person.id" }),
})
```

Feels maybe a bit verbose. Another spitball:

```ts
import { createSchema, table, enum, column } from "@ponder/core"

const schema = createSchema({
  PetKind: enum(["ONE", "TWO", "THREE"]),
  Person: table({
    id: column("string"),
    age: column("int"),
  }),
  Dog: table({
    id: column("string"),
    kind: column("PetKind"),
    ownerId: column("string", { references: "Person.id" }),
})
```

packages/core/src/user-store/postgres/store.ts
```diff
@@ -89,60 +90,45 @@ export class PostgresUserStore implements UserStore {
...
+                  col = col.check(
+                    sql`${sql.ref(columnName)} in (${sql.join(
+                      schema!.enums[column.type.slice(5)].map((v) => sql.lit(v))
+                    )})`
+                  );
+                  return col;
+                }
+              );
+            } else {
+              // Handle all other column types
```
0xOlias, line 123: It might be worth adding a foreign key constraint here for "references" fields. That would be the SQL thing to do. This way, if someone tries to insert a `Share` entity with a `traderId` value that is not found in the `Trader` table `id` column, they'll get an error.

We weren't doing this before, but it's worth thinking about.

0xOlias (approved):
LGTM 👑

PR #382, opened 2023-10-10
Move config build to build/

kyscott18:
Because this file is in the build step, it should be in the `build/` folder. Also sets the repo up for the adjacent function `buildSchema()`


0xOlias (approved):
LGTM

PR #383, opened 2023-10-12
fix http protocol when visiting graphiql

o-az:
the protocol was always set to `http`. Now it checks if `localhost` then `http` otherwise `https`

### DIFF
--- a/packages/core/src/server/service.ts
+++ b/packages/core/src/server/service.ts
@@ -170,14 +170,18 @@ export class ServerService {
         case "POST":
           return graphqlMiddleware(request, response, next);
         case "GET": {
+          const host = request.get("host");
+          if (!host) {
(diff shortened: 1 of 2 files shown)


packages/core/src/server/service.ts
```diff
@@ -195,14 +199,18 @@ export class ServerService {
         case "POST":
           return graphqlMiddleware(request, response, next);
         case "GET": {
+          const host = request.get("host");
+          if (!host) {
+            return response.status(400).send("No host header provided");
```
0xOlias, line 204: Do you know if there are any valid cases where this would happen?

o-az: I don't but `host` type is `string | undefined` so I added this check to make ts happy

PR #389, opened 2023-10-20
run `pnpm dedupe`

kyscott18:
`pnpm dedupe` deletes all unused deps from the lockfile that might solve some incompatibility issues with different versions of pnpm. This removed 93 packages according to the cli.

https://pnpm.io/cli/dedupe

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -106,13 +106,13 @@ importers:
     devDependencies:
       '@mdx-js/react':
         specifier: ^2.1.5
-        version: 2.1.5(react@18.2.0)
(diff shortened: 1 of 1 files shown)


0xOlias (approved):
TIL. Just for my understanding, were you running into issues around this? Haven't really spent the time to properly understand pnpm lockfile mechanics.

kyscott18:
Sometimes it would do this automatically for me when I would do a fresh `pnpm install` on main, updating the lockfile when I didn't want to.

Also, I suspect that it had something to do with random actions like [this](https://github.com/0xOlias/ponder/actions/runs/6587147159/job/17896850871?pr=388) failing

PR #390, opened 2023-10-20
Actions caching

kyscott18:
Utilize caching in order to speed up some actions


kyscott18:
After looking at some actions in practice, the extra caching seems like it sped up the setup step by about 15 seconds (25%). Not great

examples/with-docker/package.json
```diff
@@ -7,12 +7,12 @@
     "lint": "eslint ."
   },
   "dependencies": {
-    "@ponder/core": "latest"
```
kyscott18, line 10: @0xOlias Is this a mistake?

0xOlias: I think so. NGL, I haven't ran this docker example in a while so not certain it's working at all.

0xOlias (approved):
LGTM

PR #404, opened 2023-11-02
Handle custom topics for factory contracts internally

kyscott18:
Handle passing topics to factory contracts which includes: 

- updating logic surround `factories` and `factoryLogFilterIntervals` tables
- changing the type of the config
- utilities for building fragments from topics


packages/core/src/event-store/postgres/store.ts
```diff
@@ -373,46 +376,132 @@ export class PostgresEventStore implements EventStore {
...
+        baseJoin = join.on(({ or, cmpr }) =>
+          or([
+            cmpr("childAddressLocation", "is", null),
+            cmpr(
+              "fragmentChildAddressLocation",
+              "=",
+              sql.ref("childAddressLocation")
+            ),
+          ])
+        );
```
0xOlias, line 470: Unlike with log filters address + topics, I think these values will always be defined. So, we can probably replace this with:
```ts
let baseJoin = join.on(({ or, cmpr }) =>
  and([
    cmpr("fragmentAddress", "=", sql.ref("address")),
    cmpr("fragmentEventSelector", "=", sql.ref("eventSelector")),
    cmpr("fragmentChildAddressLocation", "=", sql.ref("childAddressLocation"))
  ])
);
```
The topics below are still optional, so that block looks good.

packages/core/src/event-store/postgres/store.ts
```diff
@@ -373,46 +376,132 @@ export class PostgresEventStore implements EventStore {
...
+    }, {} as Record<number, { startBlock: bigint; endBlock: bigint }[]>);
+
+    const fragmentIntervals = fragmentsWithIdx.map((f) => {
+      return (intervalsByFragment[f.idx] ?? []).map(
+        (r) =>
+          [Number(r.startBlock), Number(r.endBlock)] satisfies [number, number]
+      );
     });
+
+    return intervalIntersectionMany(fragmentIntervals);
```
0xOlias, line 504: This code path was written before fragment IDs were a thing - I think we might be able to simplify this a bit by replacing this notion of fragment indexes with fragment IDs. Then, in a single pass, construct the `fragmentIntervals` array. Would be able to copy the same change in 4 places across the two stores.

packages/core/src/event-store/store.test.ts
```diff
@@ -639,6 +639,36 @@ test("insertFactoryLogFilterInterval inserts and merges child contract intervals
...
+    block: blockOne,
+    transactions: blockOneTransactions,
+    logs: blockOneLogs,
+    interval: { startBlock: 0n, endBlock: 500n },
+  });
+
+  const intervals = await eventStore.getFactoryLogFilterIntervals({
+    chainId: 1,
+    factory: factoryCriteria,
+  });
```
0xOlias, line 667: This test is fine, but it could be worth testing the inclusion rules. Like, if you insert a factory criteria with `topics: [[1, 2]]` and then request the same factory criteria but with `topics: [1]` the block ranges initially inserted should returned (because the initial request includes all logs present in the second).

packages/core/src/historical-sync/service.ts
```diff
@@ -697,6 +697,7 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
     for await (const childContractAddressBatch of iterator) {
       const batchLogs = await this._eth_getLogs({
         address: childContractAddressBatch,
+        topics: factory.criteria.topics,
```
0xOlias, line 700: Pretty cool that this is the only diff in this file lmao

packages/core/src/utils/fragments.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/**
+ * Generates factory fragments from a factory.
+ *
+ * @param factory Factory to be decomposed into fragments.
+ * @returns A list of factory fragments.
+ */
+export function buildFactoryFragments({
+  address,
```
0xOlias, line 35: I find it _slightly_ misleading to put address and topics first like this, when here `address` means a static, singular factory contract address while in a log filter `address` means something different. Could be worth renaming either `FactoryCriteria.address` -> `FactoryCriteria.factoryAddress` or `FactoryCriteria.topics` -> `FactoryCriteria.childTopics`. Don't love either. Wdyt?

0xOlias (commented):
Couple ideas!

packages/core/src/event-store/postgres/store.ts
kyscott18: Nice catch, pretty sure the second line of your snippet is supposed to be `or`.

kyscott18: Never mind, I think the first line is supposed to have `and` as a function param instead of `or`

PR #406, opened 2023-11-03
Refactor config 

kyscott18:


packages/core/src/config/sources.ts
```diff
@@ -0,0 +1,185 @@
...
+            chainId: network.chainId,
+            events,
+            // optionally overridden properties
+            startBlock: networkContract.startBlock ?? contract.startBlock ?? 0,
+            endBlock: networkContract.endBlock ?? contract.endBlock,
+            maxBlockRange:
+              networkContract.maxBlockRange ?? contract.maxBlockRange,
+          } as const;
+
+          if ("factory" in contract) {
```
0xOlias, line 108: I think this won't "fall through" in the case where `factory` is not defined on the base contract but is defined per-network?

packages/core/src/config/sources.ts
```diff
@@ -0,0 +1,185 @@
...
+                ...buildFactoryCriteria(resolvedFactory),
+                topics,
+              },
+            } as const satisfies Factory;
+          } else {
+            // log filter
+
+            const resolvedAddress =
+              ("address" in networkContract && networkContract.address) ||
+              contract.address;
```
0xOlias, line 128: Shouldn't it be possible for `address` to be a list of addresses?

packages/core/src/config/sources.ts
```diff
@@ -0,0 +1,185 @@
...
+  >
+): Topics => {
+  if (Array.isArray(filter.event)) {
+    // List of event signatures
+    return [
+      filter.event
+        .map((event) =>
+          encodeEventTopics({
+            abi: [findAbiEvent(abi, event)],
+          })
```
0xOlias, line 160: I wonder if `encodeEventSelector` would be more readable here?

packages/core/src/config/sources.ts
```diff
@@ -0,0 +1,185 @@
...
+    // Single event with args
+    return encodeEventTopics({
+      abi: [findAbiEvent(abi, filter.event)],
+      args: filter.args,
+    });
+  }
+};
+
+/**
+ * Finds the abi event for the event string
```
0xOlias, line 174: Finds the event ABI item for the safe event name.

packages/core/src/event-store/postgres/migrations.ts
```diff
@@ -349,6 +349,31 @@ const migrations: Record<string, Migration> = {
         .execute();
     },
   },
+  ["2023_11_06_0_new_rpc_cache_design"]: {
+    async up(db: Kysely<any>) {
+      await db.schema.dropTable("contractReadResults").execute();
+
+      /**
+       * request format
```
0xOlias, line 357: Formatting for "request" field values:

0xOlias (approved):

packages/core/src/config/sources.ts
kyscott18: Yes

packages/core/src/config/sources.ts
kyscott18: Good catch

0xOlias (approved):
LGTM

PR #412, opened 2023-11-08
PonderApp type

kyscott18:
New `PonderApp` type strictly types the `ponder.on` property with strict names, events, contracts, networks, and clients.


0xOlias (approved):
LGTM

PR #415, opened 2023-11-09
Fix the runtime `context` object to match PonderApp

### LINKED ISSUE #394: Is ChainID exposed?

### LINKED ISSUE #400: [Feature] Add cache-enabled `PublicClient` + ABIs to `context`
We should replace `context.contracts` with a more versatile API for making cache-enabled contract read calls.

## Problems with the current approach
1. Today, `context.contracts` only contains an object for items in the `contracts` list in `ponder.config.ts` that have a stable address (AKA are not created by a factory contract). But, with factory contracts and custom event filters, you often want to read the contract that emitted the event currently being processed (dynamic address).
2. People are working around this by:
   1. Creating a new viem PublicClient instance (not cache-enabled, slows down indexing a lot)
   2. Grabbing ABI objects from `abis/` (unnecessary duplicated code & toil)
   3. Duplicating contract addresses in multiple places

Yuck!

## Ideas for a solution

Our solution should
1. Make it easy to understand which contract is being called
2. Encourage people to use cache-enabled calls whenever possible (should never need to set up your own viem Client)
3. Eliminate duplication of static contract addresses and ABIs

We're considering consolidating `contracts` and `filters` in `ponder.config.ts` into one list. I don't know what we would call that yet, but let's use `sources` as a placeholder here. The only properties that are guaranteed to be static & defined for every source is 1) source name and 2) ABI. If a source has a static/hard-coded `address`, that would be included too.

It's unlikely that you'd want to make a cross-chain contract read, and even if you did, it would not be possible to cache it. So, we could pass `context.client` or `context.publicClient` as a cache-enabled client _for the network of the event currently being processed_. This would be a departure from the idea that the `context` object is the same across all indexing functions. Further, we could limit the list of `sources` for a given event to those on the same network.

```ts
import { Abi, Address } from "abitype";

type SourceName = string;

type Context = {
  client: ReadOnlyPublicClient;
  sources: Record<
    SourceName,
    {
      name: SourceName;
      abi: Abi;
      address?: Address | undefined;
    }
  >;
};
```

### Contract read for a source with a static address
This is the case that is already handled by `context.contracts`.
```ts
import { ponder } from "@/generated";

ponder.on("FileStore:FileCreated", async ({ event, context }) => {
  const { filename, size } = event.params;

  const contents = await context.client.readContract({
    address: context.sources.FileStore.address,
    abi: context.sources.FileStore.abi,
    function: "readFile",
    args: [event.transaction.to as `0x{string}`, filename],
  });
  // OR
  const contents = await context.client.readContract({
    ...context.sources.FileStore,
    function: "readFile",
    args: [event.transaction.to as `0x{string}`, filename],
  });

  await context.entities.File.create({
    id: filename,
    data: {
      name: filename,
      size: Number(size),
      contents,
      createdAt: Number(event.block.timestamp),
    },
  });
});
```

### Contract read for a source with a dynamic address

This would handle both:
1. Contracts that are created by a factory
2. Custom event filters where the address is not specified up-front.

```ts
import { ponder } from "@/generated";

ponder.on("LlamaCore:Transfer", async ({ event, context }) => {
  
  const totalSupply = await context.client.readContract({
    address: event.log.address, // The address of the contract that emitted this event log.
    abi: context.sources.LlamaCore.abi,
    function: "totalSupply",
  })
  
  await context.entities.Token.create({
    id: event.params.actionId,
    data: { totalSupply }
  })
});
```

kyscott18:
The `context` object, passed into indexing functions, now matches the shaped of the data defined in `PonderApp`. This also updates the intermediate representation of config, performing validation in the `createConfig()` function. Fixes a bug with the strictness of types when filtering on a event with know args. 

Closes #400  and closes #394


0xOlias (approved):
LGTM

PR #418, opened 2023-11-10
Fix bugs in example repos

kyscott18:
Fix small bugs in example repos that occurred as a result of recent rapid changes. Also fixes linting.


examples/ethfs/.gitignore
```diff
@@ -3,4 +3,5 @@ node_modules/
 
 .env.local
 .ponder/
-generated/
```
0xOlias, line 6: Would be good to convert all the examples to use the .gitignore that I pasted into the create-ponder script

examples/ethfs/src/FileStore.ts
```diff
@@ -15,15 +18,17 @@ ponder.on("FileStore:FileCreated", async ({ event, context }) => {
...
-      contents: await context.contracts.FileStoreFrontend.read.readFile([
-        event.transaction.to as `0x{string}`,
-        filename,
-      ]),
+      contents: await context.client.readContract({
+        abi: FileStoreFrontendAbi,
+        functionName: "readFile",
+        address: "0xBc66C61BCF49Cc3fe4E321aeCEa307F61EC57C0b",
+        args: [event.transaction.to!, filename],
+      }),
```
0xOlias, line 31: I guess this works!! Seems like a fair enough pattern.

0xOlias (approved):

examples/ethfs/src/FileStore.ts
kyscott18: Gets tough when you want to use multichain but I wanted to get this moving.

examples/ethfs/.gitignore
kyscott18: Your example .gitignore doesn't include .ponder

PR #419, opened 2023-11-10
Update dependencies

kyscott18:


packages/core/src/server/graphql/entity.ts
```diff
@@ -70,8 +70,8 @@ export const buildEntityTypes = ({
             fieldConfigMap[columnName] = {
               type: new GraphQLNonNull(
                 new GraphQLList(
-                  new GraphQLNonNull(entityGqlTypes[column.referenceTable])
-                )
+                  new GraphQLNonNull(entityGqlTypes[column.referenceTable]),
```
0xOlias, line 73: Seems a bit strange to me that this line would have a comma. Is this what folks are doing with their lint rules these days?

0xOlias (commented):
Looks like create-ponder deps and tsconfig need to be updated as well. Looking good though!

I suspect we might run into issues with some dep breaking changes, notably `p-queue`.

PR #420, opened 2023-11-10
Move typescript and viem to peer dependencies

kyscott18:


packages/core/package.json
```diff
@@ -62,7 +66,6 @@
     "react": "^18.2.0",
     "retry": "^0.13.1",
     "stacktrace-parser": "^0.1.10",
-    "viem": "^1.18.9",
```
0xOlias: I'm still wrapping my head around peer dependencies. But, don't we need `viem` as a full dep? The interaction between `@ponder/core` and `viem` is not just in the user repository; it's also in `@ponder/core` itself? Genuinely not sure lol.

packages/core/package.json
```diff
@@ -82,7 +85,6 @@
     "rimraf": "^5.0.5",
     "supertest": "^6.3.3",
     "tsc-alias": "^1.8.8",
-    "typescript": "^5.2.2",
```
0xOlias, line 85: I noticed that the wagmi package.json still likes `typescript` (and `viem`) in dev deps. Not sure of the interaction between dev and peer deps but thought I'd mention. https://www.npmjs.com/package/wagmi?activeTab=code

0xOlias (commented):
Couple questions

PR #422, opened 2023-11-13
Update Topics type + mergeAbis utility

kyscott18:


packages/core/src/utils/mergeAbis.ts
```diff
@@ -0,0 +1,50 @@
...
+  TMerged extends Abi,
+  TImpls extends readonly Abi[],
+> = TImpls extends readonly [
+  infer First extends Abi,
+  ...infer Rest extends readonly Abi[],
+]
+  ? MergeAbis<MergeAbi<TMerged, First>, Rest>
+  : TMerged;
+
+const isAbiEqual = (a: AbiItem, b: AbiItem): boolean =>
```
0xOlias, line 27: `isAbiItemEqual`

packages/core/src/utils/mergeAbis.ts
```diff
@@ -0,0 +1,50 @@
...
+  formatAbiItem(a) === formatAbiItem(b);
+
+/**
+ * Build a single Abi from a proxy and its implementations
+ */
```
0xOlias, line 32: Combine multiple ABIs into one, removing duplicates if necessary.

packages/core/src/utils/mergeAbis.ts
```diff
@@ -0,0 +1,50 @@
...
+export const mergeAbis = <
+  const TProxy extends Abi,
+  const TImpl extends readonly Abi[],
```
0xOlias, line 35: Nit: This function could be useful in cases other than proxy/impl, so if possilbe would probably just have one generic parameter `TAbis extends readonly Abi[]` ?

0xOlias (approved):

0xOlias (approved):

PR #425, opened 2023-11-13
Benchmarks

kyscott18:


benchmarks/ponder/ponder.config.ts
```diff
@@ -1,24 +1,24 @@
...
     {
       name: "mainnet",
       chainId: 1,
-      transport: http(process.env.ANVIL_FORK_URL),
+      transport: http("http://127.0.0.1:8545"),
     },
   ],
   contracts: [
     {
       name: "RocketTokenRETH",
```
0xOlias, line 16: Will need to update these to use the new naming scheme

benchmarks/ponder/ponder.config.ts
```diff
@@ -1,24 +1,24 @@
...
-      network: "mainnet",
+      network: [{ name: "mainnet" }],
```
0xOlias, line 17: Can use the shortcut

benchmarks/ponder/ponder.schema.ts
```diff
@@ -0,0 +1,35 @@
+import { p } from "@ponder/core";
+
+export const schema = p.createSchema({
```
0xOlias, line 3: default export

benchmarks/subgraph/src/mapping.ts
```diff
@@ -1,73 +1,75 @@
...
 import {
   Account,
   Approval,
   ApprovalEvent,
   TransferEvent,
 } from "../generated/schema";
 
-export function handleTransfer(event: TTransferEvent): void {
-  const { params } = event;
+const delim = "kevin:+";
```
0xOlias, line 15: Lmao maybe change this

0xOlias (approved):
Looks good. It would be nice to add a quick `benchmarks/README.md` including steps for running the benchmarks. Can be super short, just a list of shell commands. Would be good to include the macOS build from source steps also.

0xOlias (approved):

PR #433, opened 2023-11-15
Fix bug in createConfig() requiring all networks in "networks" to be defined in "contract.network"

kyscott18:


packages/core/src/config/config.ts
```diff
@@ -243,6 +243,8 @@ export const createConfig = <
     } else {
       Object.entries(contract.network).forEach(
         ([networkName, contractOverride]) => {
+          if (!contractOverride) return;
```
0xOlias, line 246: If you're doing Object.entries, in what scenario would contractOverride be falsy?

0xOlias (approved):
LGTM but had a question

packages/core/src/config/config.ts
kyscott18:
```ts
network: {
  mainnet: undefined
}
```

Is technically allowed because of the Partial<> type in TypeScript. Not sure how to get around this.

PR #436, opened 2023-11-16
create-ponder overhaul

kyscott18:
Rewrite create ponder, allowing for starting from template repositories, and etherscan. Heavily borrowed from create-wagmi. Fixes problems with package versioning, because it no longer runs the codegen command on itself.


.eslintrc
```diff
@@ -27,5 +27,10 @@
     "@typescript-eslint/no-non-null-assertion": "off",
     "@typescript-eslint/ban-types": "off"
   },
-  "ignorePatterns": ["dist", "node_modules", "generated"]
+  "ignorePatterns": [
+    "dist",
+    "node_modules",
+    "generated",
+    "packages/create-ponder/templates/default/src/index.ts"
```
0xOlias, line 34: I think it's better to do this in a new eslintrc within the packages/create-ponder dir

packages/create-ponder/templates/etherscan/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "dependencies": {
+    "@ponder/core": "^0.0.95"
+  },
+  "devDependencies": {
+    "@types/node": "^20.9.0",
+    "abitype": "^0.10.2",
+    "eslint": "^8.53.0",
+    "eslint-config-ponder": "^0.0.95",
+    "typescript": "^5.2.2",
+    "viem": "^1.19.1"
```
0xOlias, line 20: Hmm doesn't viem need to be a dependency (not a dev dep)?

packages/create-ponder/src/index.ts
```diff
@@ -39,262 +50,387 @@ export type SerializableConfig = {
...
+          value: id,
+        })),
+      })
+    ).templateId;
+  }
 
-    const eventNamesToWrite = abiEvents.map((event) => event.name).slice(0, 2);
+  // Get template meta
+  const templateMeta = templates.find(({ id }) => id === templateId);
+  if (!templateMeta) return;
```
0xOlias, line 209: Seems like this should be a logged error?

packages/create-ponder/src/index.ts
```diff
@@ -39,262 +50,387 @@ export type SerializableConfig = {
...
+        })),
+      ).replaceAll(/"abi":"(.*?)"/g, "abi:$1")},
+    });
   `;
 
     writeFileSync(
-      path.join(rootDir, ".eslintrc.json"),
-      await prettier.format(eslintConfig, { parser: "json" }),
+      path.join(targetPath, "ponder.config.ts"),
+      await prettier.format(configContent, { parser: "typescript" }),
```
0xOlias, line 300: I wonder how much prettier slows things down

packages/create-ponder/src/index.ts
```diff
@@ -39,262 +50,387 @@ export type SerializableConfig = {
...
 
-  execSync(installCommand, {
-    cwd: rootDir,
-    stdio: "inherit",
-  });
+  const packageManager = await getPackageManager({ options });
+  if (packageManager === "npm") {
+    await fs.appendFile(
+      path.join(targetPath, ".npmrc"),
+      "\nlegacy-peer-deps = true",
```
0xOlias, line 360: Wonder if we need this

0xOlias (commented):
Some questions. Looking great

0xOlias (approved):
LGTM

PR #438, opened 2023-11-17
migrate build to tsup

kyscott18:
Migrate from tsc to tsup for bulid. Remove multiple packages previously supporting our build config. The end to end tests were also cleaned up a big because of the new tsconfig. The build step was separated from the actions setup workflow as we aim to be able to test before building.


.github/workflows/main.yml
```diff
@@ -25,12 +25,32 @@ jobs:
...
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+
+      - name: Setup
+        uses: ./.github/actions/setup
+
+      - name: Build
+        run: pnpm build
+
```
0xOlias, line 53: What's the purpose of this?

packages/core/src/_test/art-gobblers/app/src/index.ts
```diff
@@ -1,21 +1,38 @@
...
+// @ts-ignore
 ponder.on("setup", async ({ context }) => {
   const { SetupEntity } = context.db;
 
   await SetupEntity.upsert({
     id: "setup_id",
-    create: {},
-    update: {},
+    create: undefined,
+    update: undefined,
```
0xOlias, line 18: I think you can just omit these lines?

packages/core/tsconfig.json
```diff
@@ -1,7 +1,46 @@
...
+    "allowSyntheticDefaultImports": true,
+    "resolveJsonModule": true,
+
+    // Language and environment
+    "moduleResolution": "NodeNext",
+    "module": "NodeNext",
+    "target": "ESNext",
+    "lib": [
+      "ES2022", // By using ES2022 we get access to the `.cause` property on `Error` instances.
+      "DOM" // We are adding `DOM` here to get the `fetch`, etc. types. This should be removed once these types are available via DefinitelyTyped.
```
0xOlias, line 36: Wonder if we need this

packages/create-ponder/tsup.config.ts
```diff
@@ -7,6 +7,7 @@ import { defineConfig } from "tsup";
 import { dependencies } from "./package.json";
 
 export default defineConfig({
+  name: "create-ponder",
```
0xOlias, line 10: What does the name field do?

.github/workflows/main.yml
kyscott18: It is a test to see if the packages are able to be built successfully

packages/create-ponder/tsup.config.ts
kyscott18: Labels the information sent to the terminal

PR #440, opened 2023-11-17
Update tsconfig for examples and create-ponder templates

kyscott18:
Create a tsconfig specifically for examples/templates. Because the code is getting transpiled by esbuild, `"moduleResolution": "bundler"` should be used. This gets rid of the .js suffix. Organize the file with comments, and keep typechecking moderately strict to allow for user to move fast but safely.

### DIFF
--- a/examples/feature-factory/ponder.config.ts
+++ b/examples/feature-factory/ponder.config.ts
@@ -2,8 +2,8 @@ import { createConfig } from "@ponder/core";
 import { parseAbiItem } from "abitype";
 import { http } from "viem";
 
-import { LlamaCoreAbi } from "./abis/LlamaCoreAbi.js";
-import { LlamaPolicyAbi } from "./abis/LlamaPolicyAbi.js";
+import { LlamaCoreAbi } from "./abis/LlamaCoreAbi";
+import { LlamaPolicyAbi } from "./abis/LlamaPolicyAbi";
 
 const llamaFactoryEvent = parseAbiItem(
   "event LlamaInstanceCreated(address indexed deployer, string indexed name, address llamaCore, address llamaExecutor, address llamaPolicy, uint256 chainId)",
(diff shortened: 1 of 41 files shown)


examples/feature-factory/ponder.config.ts
```diff
@@ -2,8 +2,8 @@ import { createConfig } from "@ponder/core";
 import { parseAbiItem } from "abitype";
 import { http } from "viem";
 
-import { LlamaCoreAbi } from "./abis/LlamaCoreAbi";
-import { LlamaPolicyAbi } from "./abis/LlamaPolicyAbi";
+import { LlamaCoreAbi } from "./abis/LlamaCoreAbi.js";
+import { LlamaPolicyAbi } from "./abis/LlamaPolicyAbi.js";
```
0xOlias, line 6: Damn, I forgot that this will be necessary if we use `"type": "module"` in the project package.json. I'm a bit concerned that the `.js` file extension is going to confuse people quite a lot.

examples/feature-factory/tsconfig.json
```diff
@@ -1,12 +1,19 @@
...
     "strict": true,
-    "rootDir": "."
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noImplicitOverride": true,
+    "exactOptionalPropertyTypes": true,
+    "noUncheckedIndexedAccess": true,
+    "noFallthroughCasesInSwitch": true,
+    "skipLibCheck": true
   },
```
0xOlias, line 17: Might be nice to break these out into sections and add labels/comments, similar to how our internal tsconfigs look.

0xOlias (approved):
WDYT?

examples/feature-factory/ponder.config.ts
kyscott18: I think they can learn

kyscott18: I take this back, I think there is a way around it

0xOlias (approved):
LGTM, good call on `"moduleResolution": "bundler"`

PR #449, opened 2023-11-20
Developer experience upgrades for create-ponder

kyscott18:
Handles a few developer experience snags, namely:
- adding "engines" property to `package.json` to enforce node versions
- add an error message to both `@ponder/core` and `create-ponder` because not all package managers read the engines field
- change the template value of `.env.local` to be empty to a nice warning is thrown
- fix a bug in `project-friendtech` slowing down indexing
- throw a warning if rpc chainId doesn't match networks chainId
- support bun as a package manager in create-ponder


examples/feature-factory/package.json
```diff
@@ -19,5 +19,8 @@
     "eslint": "^8.53.0",
     "eslint-config-ponder": "workspace:*",
     "typescript": "^5.2.2"
+  },
+  "engines": {
+    "node": ">=18"
```
0xOlias, line 24: What's the experience like if this is violated? Do all package managers respect it?

packages/create-ponder/tsup.config.ts
```diff
@@ -36,5 +37,18 @@ export default defineConfig({
             : name.replace(/^\./, "_dot_"),
       },
     );
+
+    readdirSync(targetPath)
```
0xOlias, line 41: I think we'll also need to remove the `https://eth-sepolia.g.alchemy.com/v2/...` from the `.env.example` files in the `examples/` directory, right?

packages/core/src/sync-realtime/service.ts
```diff
@@ -80,6 +80,14 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
     const latestBlock = await this.getLatestBlock();
     const latestBlockNumber = hexToNumber(latestBlock.number);
 
+    // Fetch the chainId from the rpc, warn if not matching
+    const rpcChainId = await this.network.client.getChainId();
```
0xOlias, line 84: Would be best to do these requests concurrently, because this is in the critical path during setup.

packages/create-ponder/src/index.ts
```diff
@@ -419,6 +419,14 @@ export async function run({
     .option("--skip-git", "Skips initializing the project as a git repository")
     .help();
 
+  // Check Nodejs version
+  if (Number(process.version.split(".")[0].slice(1)) < 18)
+    throw Error(
+      pico.red(
+        `Node version:${process.version} does not meet the >=18 requirement`,
+      ),
+    );
```
0xOlias, line 428: Same question as above re: Bun.

packages/core/src/Ponder.ts
```diff
@@ -72,6 +72,12 @@ export class Ponder {
     this.common = { options, logger, errors, metrics, telemetry };
 
     this.buildService = new BuildService({ common: this.common });
+
+    // Check Nodejs version
+    if (Number(process.version.split(".")[0].slice(1)) < 18)
+      this.common.logger.fatal({
+        msg: `Node version:${process.version} does not meet the >=18 requirement`,
+      });
```
0xOlias, line 80: Two things:
1. Will this work on Bun?
2. I'm not sure that the process would ever make it to this line in practice, because the error could (probably does!) happen while executing dependency code. So, it might make sense to do this in `bin/ponder.ts` before doing basically anything else.

0xOlias (commented):
Couple more questions

packages/create-ponder/src/helpers/getPackageManager.ts
```diff
@@ -1,21 +1,26 @@
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
-import { detect } from "detect-package-manager";
```
kyscott18, line 3: This was consistently wrong and slow, the checks we have should cover what we need

examples/feature-factory/package.json
```diff
@@ -19,5 +19,11 @@
     "eslint": "^8.53.0",
     "eslint-config-ponder": "workspace:*",
     "typescript": "^5.2.2"
+  },
+  "trustedDependencies": [
```
0xOlias, line 23: Sorry for backtracking, but I think we actually don't want this. AFAIK Bun is the only package manager that respects `trustedDependencies`. And, once we actually support Bun as a runtime, we won't use `better-sqlite3` at runtime, we will use  `bun:sqlite`.

0xOlias (approved):
Nits

examples/feature-factory/package.json
kyscott18: The reason I included this is because this allows for `bunx create-ponder` to function properly. Then running `bun run dev` works because it by default uses node 20.8

0xOlias (approved):

PR #451, opened 2023-11-20
vitest typecheck

kyscott18:
Test types using vitest in ci

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,6 +28,9 @@ jobs:
       - name: Typecheck
         run: pnpm typecheck
 
+      - name: Vitest typecheck
+        run: cd packages/core && pnpm vitest typecheck
+
       - name: Lint
         run: pnpm lint
 


.github/workflows/main.yml
```diff
@@ -28,6 +28,9 @@ jobs:
       - name: Typecheck
         run: pnpm typecheck
 
+      - name: Vitest typecheck
+        run: cd packages/core && pnpm vitest typecheck
+
```
0xOlias, line 33: Is this running as expected? Looks like maybe it should be a flag instead of an option. https://vitest.dev/guide/testing-types#run-typechecking

Looks like CI is failing because vitest is still running the global setup script, which is really annoying and seems like a bug.

kyscott18: It's not running as expected because it passed on my local machine before I pushed. "--typecheck" is for the new version of vitest. I say we just wait until we upgrade the whole vite suite because these tests aren't super important. I believe they actually would cause "tsc" to fail.

PR #453, opened 2023-11-21
Reload sync and index state for a specific chain ID when developing

bankisan:
### Changes
- Adds an `/admin/reload` route to the server which will reload all the sync and index data for a specific chain ID
- Adds helper reset methods to services
- `app.use()` -> `app.all()` for custom routing behavior for a single path
- Adds more testing

### Steps to test
- Run an example app from `/examples` using `ponder dev`
- After historical sync has completed, reload the state using (replacing the `<CHAIN_ID>` placeholder for your specific chain:
 - `curl -X POST "localhost:42069/admin/reload?chainId=<CHAIN_ID>"`

### DIFF
--- a/packages/core/src/Ponder.ts
+++ b/packages/core/src/Ponder.ts
@@ -433,5 +433,10 @@ export class Ponder {
         this.serverService.setIsHistoricalIndexingComplete();
       }
     });
+
+    // Server listeners.
+    this.serverService.on("admin:reload", async () => {
+      // TODO: Reload only the specified chainIds.
+    });
   }
 }
--- a/packages/core/src/Ponder.ts
+++ b/packages/core/src/Ponder.ts
@@ -435,8 +435,36 @@ export class Ponder {
     });
 
     // Server listeners.
-    this.serverService.on("admin:reload", async () => {
-      // TODO: Reload only the specified chainIds.
+    this.serverService.on("admin:reload", async ({ chainId }) => {
+      const syncServiceForChainId = this.syncServices.find(
+        ({ network }) => network.chainId === chainId,
+      );
+      if (!syncServiceForChainId) {
+        this.common.logger.warn({
+          msg: `No network defined for chainId: ${chainId}`,
+        });
+        return;
+      }
+
+      // Reload the sync service by killing, setting up, and then starting again.
+      await syncServiceForChainId.realtime.kill();
+      await syncServiceForChainId.historical.kill();
+
+      await this.syncStore.deleteRealtimeData({
+        chainId,
+        fromBlock: BigInt(0),
+      });
+
+      const blockNumbers = await syncServiceForChainId.realtime.setup();
+      await syncServiceForChainId.historical.setup(blockNumbers);
+      //
+      await syncServiceForChainId.realtime.start();
+      syncServiceForChainId.historical.start();
+
+      // Reload the indexing service with existing schema. We use the exisiting schema as there is
+      // alternative resetting behavior for a schema change.
+      await this.indexingService.reset();
+      await this.indexingService.processEvents();
     });
   }
 }
--- a/packages/core/src/Ponder.ts
+++ b/packages/core/src/Ponder.ts
@@ -192,6 +192,7 @@ export class Ponder {
 
     // One-time setup for some services.
     await this.syncStore.migrateUp();
+    await this.serverService.start();
 
     // Finally, load the schema + indexing functions which will trigger
     // the indexing service to reload (for the first time).
@@ -208,7 +209,7 @@ export class Ponder {
         databaseKind: this.syncStore.kind,
       },
     });
-    await this.serverService.start({ mode: "development" });
+    this.serverService.registerDevRoutes();
 
     await Promise.all(
       this.syncServices.map(async ({ historical, realtime }) => {
@@ -231,7 +232,6 @@ export class Ponder {
       },
     });
 
-    await this.serverService.start({ mode: "production" });
     // If not using `dev`, can kill the build service here to avoid hot reloads.
     await this.buildService.kill();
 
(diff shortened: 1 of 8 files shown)


packages/core/src/server/service.test.ts
```diff
@@ -1575,3 +1576,74 @@ test.skip("serves derived entities versioned at provided timestamp", async (cont
 
   await service.kill();
 });
+
+// Admin routes.
+
+describe("/admin/reload", async () => {
+  it("emits chainIds in reload event", async (context) => {
```
0xOlias, line 1583: Style nit: We prefer top-level `test` statements with longer descriptions over `describe`/`it` blocks. I've found that nesting tests can get unwieldy very quick (see https://kentcdodds.com/blog/avoid-nesting-when-youre-testing).

So these would just be
```ts
test("/admin/reload emits chainIds in reload event", async (context) => {
  // ...
})
```

Totally fine if the first few words of the description are repeated many times.

packages/core/src/server/service.ts
```diff
@@ -124,6 +131,33 @@ export class ServerService {
       }
     });
 
+    // TODO: Only register in development mode.
+    this.app.post("/admin/reload", async (req, res) => {
+      try {
+        const chainIdsQuery = req.query.chainIds ?? "31337";
```
0xOlias, line 137: 1. What would this look like as a curl request?
2. I think it makes sense to response with a validation error if the user does not include a chain ID rather than default. Not specifying one seems like a "bad habit".

packages/core/src/server/service.ts
```diff
@@ -124,6 +131,33 @@ export class ServerService {
...
+        if (typeof chainIdsQuery !== "string") {
+          res.status(400).end("chainIds must be a string");
+        }
+
+        // TODO: Enforce number.
+        const chainIds: number[] = [];
```
0xOlias, line 143: Maybe you disagree, but I think it would be reasonable to only support one chain at a time (what would be the use case for many at once)?

packages/core/src/server/service.test.ts
```diff
@@ -1575,3 +1576,74 @@ test.skip("serves derived entities versioned at provided timestamp", async (cont
...
+    const { common, indexingStore } = context;
+    const { service } = await setup({
+      common,
+      indexingStore,
+      options: {
+        hasCompletedHistoricalIndexing: false,
+      },
+    });
+    const emitSpy = vi.spyOn(service, "emit");
```
0xOlias, line 1592: This looks good!

0xOlias (commented):
Looking good!

packages/core/src/server/service.ts
bankisan: 1. 
`curl -X POST "localhost:42069/admin/reload?chainId=31337"`

That would be to reload `anvil`'s default chain ID.

2. I agree, I ended up removing it :)

packages/core/src/server/service.test.ts
```diff
@@ -1575,3 +1576,74 @@ test.skip("serves derived entities versioned at provided timestamp", async (cont
...
+    .post("/admin/reload")
+    .query({ chainId: "badchainid" })
+    .expect(400);
+
+  expect(emitSpy).not.toHaveBeenCalled();
+
+  await service.kill();
+});
+
+test("/admin/reload does not exist if dev routes aren't registered", async (context) => {
```
0xOlias, line 1629: Nice!

packages/core/src/server/service.ts
```diff
@@ -158,7 +165,7 @@ export class ServerService {
      * GET /graphql -> returns graphiql page html
      * POST /graphql -> returns query result
      */
-    this.app?.use("/graphql", (request, response, next) => {
+    this.app?.all("/graphql", (req, res, next) => {
```
0xOlias, line 168: Just re-read the Express docs. This makes sense!

0xOlias (commented):
Looking good - see a test failure though

packages/core/src/Ponder.ts
```diff
@@ -427,5 +429,61 @@ export class Ponder {
...
+        chainId,
+        fromBlock: BigInt(0),
+      });
+
+      this.syncGatewayService.resetCheckpoints({ chainId });
+
+      // Reload the sync services for the specific chain by killing, setting up, and then starting again.
+      await syncServiceForChainId.realtime.kill();
+      await syncServiceForChainId.historical.kill();
+
```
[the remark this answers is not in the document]
bankisan: It's possible, but I wanted to keep all the sync service reload behavior in a single block.

packages/core/src/Ponder.ts
```diff
@@ -427,5 +429,61 @@ export class Ponder {
...
+      if (!syncServiceForChainId) {
+        this.common.logger.warn({
+          service: "server",
+          msg: `No network defined for chainId: ${chainId}`,
+        });
+        return;
+      }
+
+      // Clear all the metrics for the sources.
+      syncServiceForChainId.sources.forEach(({ networkName, contractName }) => {
```
0xOlias, line 447: It seems like the metric resets should happen within each service. I think you could do it in the `kill` method and that would be fine? Aside, I wonder if we'll run into issues with using the same method (`kill`) for two things:
1. Shutdown service, clean up all resources, get ready for process to exit
2. Clear service internal state during a reload

Don't have an immediate answer, just a thought.

0xOlias (approved):
LGTM

packages/core/src/Ponder.ts
bankisan: Yeah, I think `kill()` is a bit overloaded right now. Maybe something like `stop()` when we want to do these reloads makes most sense. I'll take a look into it today.

PR #455, opened 2023-11-22
Fix historical sync bugs

0xOlias:
- Fixes bug where progress interval calculations sometimes reported an incorrect checkpoint. This fixes a user-facing issue where the historical sync stalled out, failed to sync all events, but the indexing service continued processing an incomplete history of events.
- Fixes a bug where using `instanceof` to distinguish viem errors was not working properly (maybe due to viem being a peer dep)?

### DIFF
--- a/packages/core/src/utils/interval.test.ts
+++ b/packages/core/src/utils/interval.test.ts
@@ -173,6 +173,18 @@ test("intervalDifference handles full difference", () => {
   expect(result).toEqual([]);
 });
 
+test("intervalDifference does not mutate inputs", () => {
+  const initial = [6, 17] satisfies [number, number];
+  const remove = [
+    [0, 10],
+    [12, 15],
+  ] satisfies [number, number][];
+
+  intervalDifference([initial], remove);
+
+  expect(initial).toStrictEqual([6, 17]);
+});
+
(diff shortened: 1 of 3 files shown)


packages/core/src/utils/interval.test.ts
```diff
@@ -173,6 +173,18 @@ test("intervalDifference handles full difference", () => {
...
 });
 
+test("intervalDifference does not mutate inputs", () => {
+  const initial = [6, 17] satisfies [number, number];
+  const remove = [
+    [0, 10],
+    [12, 15],
+  ] satisfies [number, number][];
+
+  intervalDifference([initial], remove);
```
0xOlias, line 183: This was mutating `initial` 🤦 .

packages/core/src/utils/interval.ts
```diff
@@ -114,14 +114,18 @@ export function intervalDifference(
   initial: [number, number][],
   remove: [number, number][],
 ) {
+  // Create copies to avoid mutating the originals.
+  const initial_ = initial.map((interval) => [...interval] as [number, number]);
+  const remove_ = remove.map((interval) => [...interval] as [number, number]);
```
0xOlias, line 119: Might be a better way to do this, but the tests pass.

PR #458, opened 2023-11-23
Wrong API breaks the `create` command for Goerli Base

0xjjpa:
Updated to the right one


0xOlias (approved):
Thanks for contributing!

PR #461, opened 2023-11-24
Cleanup server logic

bankisan:


packages/core/src/server/service.ts
```diff
@@ -34,46 +35,31 @@ export class ServerService {
...
+
+    // GraphQL routes.
+    this.app?.all(
+      "/graphql",
+      this.handleGraphql({ shouldWaitForHistoricalSync: true }),
+    );
+    this.app?.all(
+      "/",
+      this.handleGraphql({ shouldWaitForHistoricalSync: false }),
+    );
```
0xOlias, line 59: Love this

packages/core/src/server/service.ts
```diff
@@ -105,30 +91,88 @@ export class ServerService {
...
+    this.isHistoricalIndexingComplete = true;
+
+    this.common.logger.info({
+      service: "server",
+      msg: `Started responding as healthy`,
+    });
+  }
+
+  // Middleware handlers.
+  private middlewareRequestTiming(): Handler {
```
0xOlias, line 121: Maybe could call this `metricsMiddleware` ?

0xOlias (approved):
LGTM

PR #467, opened 2023-11-27
Config type bug fixes

kyscott18:
- "parameter" property is strictly typed in "factory"
- ".on()" filters down events based on "filter"
- test file cleanups

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -67,6 +67,7 @@ export type ContractRequired<
   TNetworkNames extends string,
   TAbi extends Abi | unknown,
   TEventName extends string,
+  TFactoryEvent extends AbiEvent | undefined,
 > = {
   /** Contract application byte interface. */
   abi: Abi;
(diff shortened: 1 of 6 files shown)


packages/core/src/config/config.ts
```diff
@@ -67,36 +67,47 @@ export type ContractRequired<
...
 > = (
   | {
       address?: `0x${string}` | readonly `0x${string}`[];
+      factory?: never;
     }
   | {
+      address?: never;
       /** Factory contract configuration. */
-      factory: {
+      factory?: {
```
0xOlias, line 102: Interesting - were we previously requiring a `factory` field if `address` was not defined?

packages/core/src/config/sources.test.ts
```diff
@@ -1,10 +1,17 @@
...
-import { abiSimple, abiWithSameEvent } from "./config.test-d.js";
 import { buildSources } from "./sources.js";
 
+const event0 = parseAbiItem("event Event0(bytes32 indexed arg)");
+const event1 = parseAbiItem("event Event1()");
+const event1Overloaded = parseAbiItem("event Event1(bytes32)");
+
+const address1 = "0x0000000000000000000000000000000000000001";
+const bytes1 =
+  "0x0000000000000000000000000000000000000000000000000000000000000001";
```
0xOlias, line 13: This is great, much clearer and easier to follow these tests now.

0xOlias (approved):
LGTM

packages/core/src/config/config.ts
kyscott18: No because address was also optional. I tested this a bunch and this is the setup we want.

PR #469, opened 2023-11-27
Actions timeout

kyscott18:
Actions will timeout after 3 minutes

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,6 @@ jobs:
   lint:
     name: Lint
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -14,6 +14,7 @@ jobs:
   lint:
     name: Lint
     runs-on: ubuntu-latest
+    timeout-minutes: 3
```
0xOlias, line 17: I think the issue only happens on `test` and `test-create-ponder` right, because it's from `vitest`? Can probably remove from `lint` and `build` if so.

0xOlias (approved):

Issue #288, opened 2023-08-01
[Bug] Events are skipped on fast chains (Arbitrum)


0xOlias:
Because Ponder supports indexing multiple chains in the same app, it uses block timestamps to order events (rather than block numbers which are meaningless across chains). Ponder encounters a bug if a chain has several consecutive blocks with the same block timestamp (note that block timestamps are only granular to 1 second).

### Background

After ingesting a new head block for a given chain, Ponder's realtime sync service emits an event (`newRealtimeCheckpoint`) to downstream services announcing that logs are now available **up to and including** the block timestamp of the newly processed block.

Then, the user handler service runs a DB query to fetch these new logs from the store and runs them against the user-provided handler code. Internally, the user handler service stores a piece of mutable state `eventsProcessedToTimestamp`, again with a 1-second granularity.

### The problem

The bug occurs when a series of blocks is produced that looks like this.
```
block 10, timestamp 100
block 11, timestamp 100
------> ingest latest block (11) and emit newRealtimeCheckpoint(timestamp: 100)
  ----> user handlers service processes new logs, sets eventsProcessedToTimestamp: 100
block 12, timestamp 100 <-- logs in this block are never handled
block 13, timestamp 100 <-- logs in this block are never handled
block 14, timestamp 101
------> ingest latest block (14) and emit newRealtimeCheckpoint(timestamp: 101)
  ----> user handlers service processes new logs, sets eventsProcessedToTimestamp: 101
block 15, timestamp 101
...
```

### Potential solutions

One solution would be to update the realtime sync service to only emit `newRealtimeCheckpoint(timestamp)` once it has confirmed that no more blocks will be produced with the same timestamp.

A naive implementation would be `newRealtimeCheckpoint(timestamp: newHeadBlock.timestamp - 1)`. This effectively means: only process blocks once we've received the next block. On a fast chain like arbitrum this works pretty well - we're getting new blocks faster than 1/second. We're often ingesting & processing blocks in batches of ~4 anyway.

However, this doesn't work well on a chain like mainnet, because the service would always be one entire block (~12s) behind latest.

AFDudley:
Two quick solutioms come to mind:
1. Can you just append the block number to the end of the time stamp?
2. If you know the endpoint is synced, them you can use the time you received the block instead of the timestamp.

[...]

PR #470, opened 2023-11-27
[WIP] Granular event checkpoints

0xOlias:
This PR updates Ponder's internal checkpointing system from simple block timestamps to a much more granular value:
1. Block timestamp
2. Chain ID
3. Block number
4. Execution index (log index)

This PR should also solve the issue where events are sometimes skipped on chains with sub-second block times (#288)


packages/core/src/utils/checkpoint.ts
```diff
@@ -0,0 +1,86 @@
...
+) => {
+  if (a.blockTimestamp !== b.blockTimestamp)
+    return a.blockTimestamp > b.blockTimestamp;
+  if (a.chainId !== b.chainId) return a.chainId > b.chainId;
+  if (a.blockNumber !== b.blockNumber) return a.blockNumber > b.blockNumber;
+  return true;
+};
+
+export const checkpointMax = (...checkpoints: SyncCheckpoint[]) =>
+  checkpoints.reduce((max, checkpoint) => {
```
kyscott18, line 73: I think you could use the helper function you wrote above for the min and max

packages/core/src/indexing/service.ts
```diff
@@ -403,8 +418,8 @@ export class IndexingService extends Emittery<IndexingEvents> {
             });
           }
 
-          // Decode events, dropping any that cannot be decoded using the provided ABI item.
-          const decodedEvents = events.reduce<LogEvent[]>((acc, event) => {
+          // Decode events and add them to the queue.
+          events.forEach((event) => {
```
kyscott18, line 422: Nice cleanup!

kyscott18 (approved):
This looks really awesome!

kyscott18 (commented):
This looks really awesome!

PR #473, opened 2023-11-29
[WIP] Example project-foundry

bankisan:


examples/project-foundry/wagmi.config.ts
```diff
@@ -0,0 +1,16 @@
+import { defineConfig } from "@wagmi/cli";
```
0xOlias, line 1: Some ideas for how to structure this example using a more realistic monorepo setup:

```
examples/
  with-foundry/
    contracts/ <- this would be the foundry app, what's currently called "foundry"
    ponder/ or backend/ <- ponder app
    README.md <- top-level readme explaining the ponder + foundry integration 
    wagmi.config.ts <- Maybe this also makes sense at the top level?
    abis/ <- Maybe this also makes sense at the top level?
```

Also - we could use the `with-` prefix to label examples of integrations, eg later add `with-nextjs`.

examples/project-foundry/foundry/script/Deploy.s.sol
```diff
@@ -0,0 +1,17 @@
...
+
+contract CounterScript is Script {
+    function setUp() public {}
+
+    function run() public {
+        vm.broadcast();
+        Counter counter = new Counter();
+
+        console.log("Deployments:");
+        console.log("Counter deployed at: ", address(counter));
```
0xOlias, line 15: Could we also add the ponder reload cURL command here, like the docs suggest?

examples/project-foundry/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "codegen": "ponder codegen",
+    "deploy": "cd foundry && forge script script/Deploy.s.sol --broadcast --fork-url http://localhost:8545 --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
+    "dev": "ponder dev",
+    "generate:abi": "wagmi generate",
+    "increment": "cd foundry &&  cast send 0x5FbDB2315678afecb367f032d93F642f64180aa3 \"increment()\" --rpc-url http://localhost:8545 --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
+    "lint": "eslint .",
+    "start": "ponder start",
+    "start:anvil": "anvil --block-time 10",
+    "typecheck": "tsc"
+  },
```
0xOlias, line 15: See monorepo comment first - after moving the ponder and foundry-specific scripts to their own directories, it might still make sense to have a top-level package.json here with certain scripts. Lmk what you think.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "examples/project-foundry/foundry/lib/forge-std"]
```
0xOlias, line 1: Nw, fine to add this to the top-level of the repo.

0xOlias (commented):
Looking great. Had a few ideas!

examples/project-foundry/foundry/script/Deploy.s.sol
bankisan: I added it to the README. Let me know if that works better for you.

bankisan:
@kyscott18 Should we wait for your nextjs example before merging this in?

PR #474, opened 2023-11-30
Historical sync retry and queue logic

kyscott18:
Fix several bugs with sync-historical retry, timeout, and queue logic. Should increase reliability and sync speeds.

The results of the bugs were:
- No timeout on long running tasks
- Not aborting failed tasks
- Opaque retry logic
- Error messages not printed to the terminal
- Process not immediately exiting after kill

The main changes implemented were:
- Use p-queue internal timeout
- Use abort controller to abort long rpc calls
- Use our own retry logic instead of viem
- Historical sync requests are done with helper functions instead of viem client.request

### DIFF
--- a/packages/core/src/config/networks.ts
+++ b/packages/core/src/config/networks.ts
@@ -1,20 +1,32 @@
-import { type Client, createPublicClient, type PublicClient } from "viem";
+import {
+  type Client,
+  createPublicClient,
+  type PublicClient,
+  type Transport,
+} from "viem";
+import { rpc } from "viem/utils";
 
 import type { Config } from "@/config/config.js";
 import type { Common } from "@/Ponder.js";
 import { chains } from "@/utils/chains.js";
 
+type Request = (
+  options: Parameters<typeof rpc.webSocketAsync>[1] &
+    Parameters<typeof rpc.http>[1],
+) => any;
+
 export type Network = {
   name: string;
   chainId: number;
   client: PublicClient;
+  request: Request;
   pollingInterval: number;
   defaultMaxBlockRange: number;
   maxRpcRequestConcurrency: number;
   finalityBlockCount: number;
 };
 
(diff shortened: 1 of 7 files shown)


packages/core/src/config/networks.ts
```diff
@@ -161,29 +181,25 @@ function getFinalityBlockCount({ chainId }: { chainId: number }) {
...
         return [transport.url as string | undefined];
       }
       case "webSocket": {
-        // TODO: Enable this codepath once we can make this function async.
-        // This will happen when we make the config file reloadable during
-        // https://github.com/0xOlias/ponder/issues/322.
-        // try {
-        //   const socket = await transport.getSocket();
-        //   return [socket.url];
-        // } catch (e) {
```
kyscott18, line 177: @0xOlias What does this catch logic do?

packages/core/src/sync-historical/service.ts
```diff
@@ -898,17 +952,33 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
...
+      const rawRequest = await this.network.request({
+        body: {
+          method: "eth_getLogs",
+          params: [options],
+        },
+        fetchOptions: { signal },
       });
+
+      if (rawRequest.error)
+        throw new RpcRequestError({
```
0xOlias, line 972: I'm a little worried about this, because we take advantage of viem error formats in the _eth_getLogs helper. This probably results in different errors.

0xOlias (commented):
Looks good, one question

packages/core/src/sync-historical/service.ts
kyscott18: I think this should be fine because this error is the exact same as the one viem is throwing

0xOlias: Ok, got it.

PR #476, opened 2023-11-30
Update client.readContract

kyscott18:
- use request helper function instead of viem client.request()
- retry logic

Note: this is lacking abort logic in readContract. Not sure how this would look without big, annoying changes

### DIFF
--- a/packages/core/src/indexing/service.test.ts
+++ b/packages/core/src/indexing/service.test.ts
@@ -146,7 +146,7 @@ test("processEvents() calls getEvents with sequential timestamp ranges", async (
   await service.kill();
 });
 
-test.only("processEvents() calls indexing functions with correct arguments", async (context) => {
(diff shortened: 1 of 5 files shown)


packages/core/src/indexing/service.test.ts
```diff
@@ -48,13 +57,39 @@ const transferIndexingFunction = vi.fn(async ({ event, context }) => {
...
+
+      await context.db.Supply.create({
+        id: event.log.id,
+        data: {
+          supply: totalSupply,
+        },
+      });
+    } catch (err) {
+      console.log(err);
+    }
```
0xOlias, line 77: Necessary to include this?

0xOlias (approved):

packages/core/src/indexing/service.test.ts
kyscott18: Oops, this was leftover from debugging.

PR #482, opened 2023-12-06
Indexing service performance bug

kyscott18:
Fixed a bug in the indexing service where the event loop was being surrendered too frequently, causing large performance loses. Also fixed many event metrics that were previously not specific enough to what they were supposed to be testing.


0xOlias (approved):
LGTM

PR #483, opened 2023-12-06
Fix zero downtime deployments

bankisan:
**Changes**
- Uses schema versions to separate new instances of the indexer
- Readonly behavior on the index store for web server responses

**Tests**
- Tested on railway with uniswap example


0xOlias:
Closing in favor of #498

PR #486, opened 2023-12-07
Fix a bug when some gas related transaction field are sometimes null

kyscott18:
This now handles "maxFeePerGas" and "maxPriorityFeePerGas" being null. This was happening on optimism testnet.


packages/core/src/types/transaction.ts
```diff
@@ -71,9 +71,9 @@ export type Transaction = Prettify<
         accessList?: never;
         gasPrice?: never;
         /** Total fee per gas in wei (gasPrice/baseFeePerGas + maxPriorityFeePerGas). Only present in EIP-1559 transactions. */
-        maxFeePerGas: bigint;
+        maxFeePerGas: bigint | undefined;
```
0xOlias, line 74: Noticing that other fields in this type use the `?:` syntax, we should try to be consistent.

PR #493, opened 2023-12-08
Bytes and Address case sensivity

### LINKED ISSUE #194: [Bug] Consider normalizing (lowercase) bytes values
If I write bytes (address or otherwise), I would expect that querying for those wouldn't be impacted by checksummed bytes.

This could be done with `bytea` column type on postgres databases, but may need some finagling in sqlite

### LINKED ISSUE #478: Address case sensitivity
Address string comparisons can return incorrect results when the casing is not correct. Many viem function return a checksummed address, while ponder internally uses lowerCase in many places. The solution should be to use checksum addresses internally.

kyscott18:
Two main case sensitivity problems were addressed:
- viem decoding returned checksum addresses while ponder was using lowercase
- bytes values were case sensitive

These issues were fixed by:
- internally representing addresses with checksum
- case insensitive comparisons for "bytes" columns

This closes #478 and closes #194

### DIFF
--- a/packages/core/src/_test/constants.ts
+++ b/packages/core/src/_test/constants.ts
@@ -997,7 +997,7 @@ export const blockOneTransactions: RpcTransaction[] = [
 
 export const blockOneLogs: RpcLog[] = [
   {
-    address: "0x15D4c048F83bd7e37d49eA4C83a07267Ec4203dA",
+    address: "0x15d4c048f83bd7e37d49ea4c83a07267ec4203da",
     blockHash:
       "0xebc3644804e4040c0a74c5a5bbbc6b46a71a5d4010fe0c92ebb2fdf4a43ea5dd",
     blockNumber: "0xec6fc6",
@@ -1014,7 +1014,7 @@ export const blockOneLogs: RpcLog[] = [
     transactionIndex: "0x45",
   },
   {
-    address: "0x72D4C048f83bd7E37D49eA4C83a07267ec4203dA",
+    address: "0x72d4c048f83bd7e37d49ea4c83a07267ec4203da",
     blockHash:
       "0xebc3644804e4040c0a74c5a5bbbc6b46a71a5d4010fe0c92ebb2fdf4a43ea5dd",
     blockNumber: "0xec6fc6",
--- a/packages/core/src/_test/constants.ts
+++ b/packages/core/src/_test/constants.ts
@@ -1,4 +1,5 @@
 import {
+  checksumAddress,
   parseAbiItem,
   type RpcBlock,
   type RpcLog,
@@ -247,7 +248,7 @@ const usdcContractAbi = [
 
 export const usdcContractConfig = {
   chainId: 1,
-  address: "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
+  address: checksumAddress("0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"),
   abi: usdcContractAbi,
   events: getEvents({ abi: usdcContractAbi }),
 } as const;
@@ -996,7 +997,7 @@ export const blockOneTransactions: RpcTransaction[] = [
 
 export const blockOneLogs: RpcLog[] = [
   {
-    address: "0x15d4c048f83bd7e37d49ea4c83a07267ec4203da",
+    address: "0x15D4c048F83bd7e37d49eA4C83a07267Ec4203dA",
     blockHash:
       "0xebc3644804e4040c0a74c5a5bbbc6b46a71a5d4010fe0c92ebb2fdf4a43ea5dd",
     blockNumber: "0xec6fc6",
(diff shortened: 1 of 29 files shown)


packages/core/src/_test/constants.ts
```diff
@@ -247,7 +248,7 @@ const usdcContractAbi = [
 
 export const usdcContractConfig = {
   chainId: 1,
-  address: "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
+  address: checksumAddress("0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"),
```
0xOlias, line 251: Checking my understanding - this is necessary because normally, these values have already passed through our validator/formatter?

packages/core/src/_test/constants.ts
```diff
@@ -996,7 +997,7 @@ export const blockOneTransactions: RpcTransaction[] = [
 
 export const blockOneLogs: RpcLog[] = [
   {
-    address: "0x15d4c048f83bd7e37d49ea4c83a07267ec4203da",
+    address: "0x15D4c048F83bd7e37d49eA4C83a07267Ec4203dA",
```
0xOlias, line 1000: Was this necessary? My expectation is that now, it doesn't matter whether RPC responses are lowercase or checksummed.

packages/core/src/indexing-store/sqlite/store.ts
```diff
@@ -763,4 +807,12 @@ export class SqliteIndexingStore implements IndexingStore {
...
     return result;
   };
+
+  private idColumnComparator = ({
+    tableName,
+    schema,
+  }: {
+    tableName: string;
+    schema: Schema | undefined;
+  }) => (schema?.tables[tableName]?.id.type === "bytes" ? "like" : "=");
```
0xOlias, line 817: Was curious about this - found that like Postgres has the `ILIKE` comparator which is case-insensitive, but for SQLite `LIKE` is case insensitive, and you can configure it (https://sqlite.org/forum/info/af7978156ce072e7e1060f0741dc10d33dc7ccaf702f6d67c5ab692bffa25174). I'm guess you found the same.

packages/core/src/indexing-store/store.test.ts
```diff
@@ -929,3 +936,126 @@ test("revert() updates versions that only existed during the safe timestamp to l
   expect(pets.length).toBe(1);
   expect(pets[0].name).toBe("Skip");
 });
+
+test("findUnique() works with bytes case sensitivity", async (context) => {
```
0xOlias, line 940: Awesome tests

packages/core/src/sync-store/postgres/format.ts
```diff
@@ -46,7 +45,7 @@ export function rpcToPostgresBlock(
     gasUsed: BigInt(block.gasUsed),
     hash: block.hash!,
```
0xOlias, line 46: Hmm, I wonder if we'll ever run into issues with inconsistent casing in other hex strings returned by various RPC providers.

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -350,7 +356,7 @@ export class PostgresSyncStore implements SyncStore {
       }
 
       if (batch.length > 0) {
-        yield batch.map((a) => a.childAddress);
+        yield batch.map((a) => checksumAddress(a.childAddress));
```
0xOlias, line 359: Again just confirming my understanding - this is required because the `selectChildAddressExpression` returns lowercase addresses, right?

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -1303,7 +1309,7 @@ export class PostgresSyncStore implements SyncStore {
 
     exprs.push(
       eb(
-        "logs.address",
+        sql`lower(logs.address)`,
```
0xOlias, line 1312: If we don't also lowercase the generated `childAddress` before we compare, I think we're counting on the log `topic1,2,3` and `data` columns to be stored as lowercase.

0xOlias (approved):
LGTM, awesome work on this.

packages/core/src/_test/constants.ts
kyscott18: Yes

packages/core/src/_test/constants.ts
kyscott18: In some cases we are using this value as the return from the rpc and in some cases we are using it as the return from the db. In the return from db case, the value is always checksummed because we checksum the value before we add it to the db. I changed this back and explicitly checksummed in the few cases when we expected this value to be checksummed.

packages/core/src/indexing-store/sqlite/store.ts
kyscott18: Yes, there is basically a mirror of this function in the postgres folder

packages/core/src/sync-store/postgres/store.ts
kyscott18: yes

packages/core/src/sync-store/postgres/store.ts
kyscott18: Yes, I think there are many spots where we are counting on return values to be lowercase. Let's talk about this more because it might make sense to make this a separate pr.

0xOlias (approved):
LGTM

PR #495, opened 2023-12-09
Config validation and reloading

### LINKED ISSUE #479: Config validation
Config should be validated at runtime. Specifically, we should check for:

- named abi parameters
- matching network names
- valid addresses
- valid filters
- valid factories

kyscott18:
Fix several bugs around config validation:

- Explicitly validate both config and schema in the build pipeline
- Stricter typing on config, with more tests
- Gracefully handle ABIs with unnamed parameters
- Reload ponder app on both config and schema changes, handling errors without shutting down the dev server
- Handle invalid and default rpc urls
- Realtime sync setup sequence retry logic (occurs with free rpc providers)

Closes #479


packages/core/src/bin/ponder.ts
```diff
@@ -65,8 +66,9 @@ cli
     const ponder = new Ponder({ options: startOptions });
     registerKilledProcessListener(() => ponder.kill());
 
-    await ponder.setup();
-    await ponder.start();
+    if (await ponder.setup()) {
+      await ponder.start();
+    }
```
0xOlias, line 71: I might structure this as
```
const isSetupSuccessful = await ponder.setup()
if (!isSetupSuccessful) return
await ponder.start()
```

packages/core/src/Ponder.ts
```diff
@@ -74,14 +74,7 @@ export class Ponder {
...
 
-  async setup({
-    syncStore,
-    indexingStore,
-  }: {
-    // These options are only used for testing.
-    syncStore?: SyncStore;
-    indexingStore?: IndexingStore;
-  } = {}) {
+  async setupInitial() {
```
0xOlias, line 77: Could we make this private, and maybe rename it `setupBuildService` ?

packages/core/src/Ponder.ts
```diff
@@ -98,10 +91,26 @@ export class Ponder {
...
+        service: "app",
+        msg: `Config validation failed, killing app`,
+      });
       await this.buildService.kill();
       return;
     }
+    return config;
+  }
 
+  async setupPrepare({
```
0xOlias, line 104: Could call this `setupCoreServices`

packages/core/src/Ponder.ts
```diff
@@ -244,15 +264,30 @@ export class Ponder {
...
-        const blockNumbers = await realtime.setup();
-        await historical.setup(blockNumbers);
+    await this.syncServiceStart();
+  }
 
-        historical.start();
-        await realtime.start();
-      }),
-    );
+  private async syncServiceStart() {
```
0xOlias, line 270: And this could be `startSyncServices`

packages/core/src/Ponder.ts
```diff
@@ -359,43 +397,79 @@ export class Ponder {
...
-      await this.setup();
-      // NOTE: We know we are in dev mode if the build service is receiving events. Build events are disabled in production.
-      await this.dev();
+        // Clear all listeners. Will be added back in setup.
+        await this.reload();
+
+        await this.setupPrepare({ config });
+        // NOTE: We know we are in dev mode if the build service is receiving events. Build events are disabled in production.
+        await this.dev();
+      } else {
```
0xOlias, line 446: Might be nice to add a comment here:
```
If the build service emits the "newConfig" event with undefined, it means there was an error
while building or validating the config on a hot reload.
```

Also - do we need to reset metrics here like we're doing in the `admin:reload` handler? If so, might make sense to handle metric resets within the `kill` function of the sync and indexing services.

packages/core/src/build/service.ts
```diff
@@ -168,10 +170,22 @@ export class BuildService extends Emittery<BuildServiceEvents> {
 
     const config = result.exports.default as Config;
 
-    // TODO: Validate config lol
+    try {
+      await validateConfig({ config });
```
0xOlias, line 174: Up to you, but I might consider having this function return `Error | undefined` rather than throw.

0xOlias (approved):
All nits. Looks great!

PR #501, opened 2023-12-12
Rpc request cache modifiers

kyscott18:
Add the ability to ignore the block of a ponder rpc action. This uses the latest block instead of the block w.r.t. indexing progress. All requests that use this tag are equivalent (will be a cache hit).


0xOlias (approved):
LGTM

PR #502, opened 2023-12-12
Address case normalization

kyscott18:
- All sync-store values are lowercase.
- All addresses passed to indexing functions are checksummed.


0xOlias (approved):
LGTM!

PR #504, opened 2023-12-13
Indexing function retries

kyscott18:
Retry indexing functions three times before stopping the service and displaying the error to the terminal. Internally, uses the same revert() function that reorgs use to reset the state of the indexing-store to what it was before the failed indexing function ran. 

Ponder actions on client no longer need retry logic and instead use this.

### DIFF
--- a/packages/core/src/indexing-store/sqlite/store.ts
+++ b/packages/core/src/indexing-store/sqlite/store.ts
@@ -189,7 +189,6 @@ export class SqliteIndexingStore implements IndexingStore {
               .updateTable(table)
               .set({ effectiveToCheckpoint: "latest" })
               .where("effectiveToCheckpoint", ">=", encodedCheckpoint)
-              // .where("effectiveToCheckpoint", "!=", "latest")
               .execute();
           }),
         );
--- a/packages/core/src/indexing-store/sqlite/store.ts
+++ b/packages/core/src/indexing-store/sqlite/store.ts
@@ -166,36 +166,34 @@ export class SqliteIndexingStore implements IndexingStore {
     });
   };
 
-  revert = async ({ safeCheckpoint }: { safeCheckpoint: Checkpoint }) => {
+  /**
+   * Revert any changes that occurred during or after the specified checkpoint.
+   */
+  revert = async ({ checkpoint }: { checkpoint: Checkpoint }) => {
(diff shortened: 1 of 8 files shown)


packages/core/src/indexing-store/sqlite/store.ts
```diff
@@ -166,26 +166,30 @@ export class SqliteIndexingStore implements IndexingStore {
...
               .execute();
 
             // Now, any versions with effectiveToCheckpoint greater than or equal
             // to the safe checkpoint are the new latest version.
             await tx
               .updateTable(table)
-              .where("effectiveToCheckpoint", ">=", encodedCheckpoint)
               .set({ effectiveToCheckpoint: "latest" })
+              .where("effectiveToCheckpoint", ">=", encodedCheckpoint)
+              // .where("effectiveToCheckpoint", "!=", "latest")
```
0xOlias, line 192: Can rm commented line

packages/core/src/indexing/service.test.ts
```diff
@@ -371,6 +371,68 @@ test("processEvents() client.readContract handles errors", async (context) => {
...
+  syncGatewayService.checkpoint = checkpoint10;
+  await service.processEvents();
+
+  expect(transferIndexingFunction).toHaveBeenCalledTimes(2);
+  expect(indexingStoreRevertSpy).toHaveBeenCalledOnce();
+
+  await service.kill();
+});
+
+test("processEvents() handles errors", async (context) => {
```
0xOlias, line 404: This is great!

packages/core/src/indexing/service.ts
```diff
@@ -613,46 +626,57 @@ export class IndexingService extends Emittery<IndexingEvents> {
...
+                service: "indexing",
+                msg: `Indexing function failed, retrying... (event=${fullEventName}, block=${Number(
+                  event.block.number,
+                )}, error=${getErrorMessage(error)})`,
+              });
+              await this.indexingStore.revert({
+                checkpoint: this.currentIndexingCheckpoint,
+              });
+              queue.addTask(task, { priority: 1, retry: true });
+              queue.start();
```
0xOlias, line 676: LFG! So stoked about this.

0xOlias (approved):
LGTM! Long time coming.

Issue #228, opened 2023-06-21
[Feature] Alternative realtime sync strategy for fast chains


0xOlias:
The new realtime indexing strategy calls `eth_getBlockByNumber` for every new block on the chain. This works just fine for the majority of chains that have a block time >=2 seconds. But, this approach does not work well for chains like Arbitrum that have a very short block time (Arbitrum targets 0.25 seconds). The reorg-safe approach also (generally) consumes more RPC usage.

It likely makes sense to offer two approaches:
1. **Reorg safe**. This is the approach currently used in main. In a nutshell, it fetches every block using `eth_getBlockByNumber`, checks if any of the registered log filters match `block.logsBloom`. If yes, call `eth_getLogs`, add the matched logs, transactions, and block to the store, then process the now-ready events.
2. **Reorg unsafe, RPC-usage optimized**. Ponder previously used this approach. On an interval, fetch the latest block, then call `eth_getLogs` from the previous latest block number to the new one. This could be further optimized by only calling `eth_getLogs` with `fromBlock` equal to some known-to-be-handled block number in the past and `toBlock: "latest"`, then de-duping any repeated events.

Approach 2 would be appropriate for any chains where it is reasonable to assume that there will never be a reorg (which, AFAIK, includes Arbitrum).

This option could be exposed to the user as a field in `networks` in `ponder.config.ts`.

EDIT: See also #506 for an additional request on this.

PR #506, opened 2023-12-13
filter realtime logs by source address

holic:
but only if all sources filter by address and none use factory contracts


0xOlias:
Hi @holic thanks for opening! For posterity, wanted to note that this was proposed as a workaround for RPC providers that require specifying an `address` for eth_getLogs requests. Today, Ponder's realtime sync does not specify one.

We plan to solve this, but would rather do it alongside a full review of our realtime sync algorithm (https://github.com/ponder-sh/ponder/issues/228) rather than patch it in. Nonetheless thanks for raising this!

PR #508, opened 2023-12-13
Fix versioning race condition

0xOlias:
Also fixes small bug where the app would crash if there was a contract with no indexing functions registered.


kyscott18 (approved):
Looks good to me. Only note is that async/await is used in a lot of spots where it would be very easy to use .then(). Probably a small performance gain that we will revisit later.

0xOlias:
> Looks good to me. Only note is that async/await is used in a lot of spots where it would be very easy to use .then(). Probably a small performance gain that we will revisit later.

Oh, I didn't think there was a perf difference between the two.

Related - using Promises for the SQLite stores at all is just overhead. The SQLite drivers (better-sqlite3, bun:sqlite) are sync, kysely just makes them async so the APIs are more compatible with other drivers.

PR #526, opened 2023-12-15
Schema metadata table for ponder indexing

bankisan:
**Changes**
- Adds metadata table for ponder indexing (in Postgres) to enable zero downtime indexing reads
- Adds `findXPublic` methods that will consistently read from the `public` namespace
- Registers an event listener on `NOTIFY` events in postgres when a publish has happened

**N.B.**: Public index methods are implicitly tested in the integration tests for the server


0xOlias:
Plan to address in a different PR

PR #529, opened 2023-12-16
Local test suite

### LINKED ISSUE #488: Local test suite
Three services have tests that rely on data from a forked node. In contrast to a entirely local test, this is slower, less reliable, and less sandboxed.

Instead of reading data from a node like alchemy, we can create mock data using Viem's test utilities and then read from Anvil.

We can then easily extend this to benchmarks that don't rely on network requests. Overall the goal is to shorten the feedback loop, relying on less external factors.

kyscott18:
Move the entire test suite to run locally, without relying on fork testing. 

Still several todos to make this a viable solution:
- [x] CI seems to either be running against a fork or has a bug
- [x] setupAnvil not fully resetting chain
- [x] e2e tests config addresses

Closes #488


kyscott18:
@0xOlias Do you want to add the `pnpm wagmi generate` command as a postinstall script or add it to the contributing docs?

packages/core/src/sync-store/store.test.ts
```diff
@@ -1446,152 +1404,173 @@ test("getLogEvents filters on fromBlock", async (context) => {
...
     log: {
-      topics: blockTwoLogs[0].topics,
+      topics: rawEvents[1].log.topics,
     },
   });
 });
 
-test("getLogEvents filters on fromCheckpoint (inclusive)", async (context) => {
-  const { syncStore } = context;
+test("getLogEvents filters on fromCheckpoint (inclusive)", async ({
```
kyscott18, line 1481: @0xOlias Can you check that these inclusivity rules are still correct?

PR #544, opened 2023-12-20
Fix contract reads and improve validation

### LINKED ISSUE #554: p.one() resolving a non-reference should be an error
```ts
    proposalId: p.int(),
    proposalRelationship: p.string().references('Proposal.id'),
    proposal: p.one('proposalId'),
 ```

### PR DESCRIPTION
**Pull request by 0xOlias**
This PR adds new validation rules for the config, indexing functions, and schema.

Fixes #554

### CODE REVIEW

**Comment by kyscott18** [comment]
I think this pr should handle the case when no events to index are found. This was a common problem that a few of our users were running into and they didn't know what to do. Right now all that happens in this case is that the indexing function progress bar is at zero, which makes it seem like indexing is just frozen and hasn't started.

File: packages/core/src/build/config/config.test.ts [hunk 1]
```diff
@@ -0,0 +1,394 @@
...
+const event0 = parseAbiItem("event Event0(bytes32 indexed arg)");
+const event1 = parseAbiItem("event Event1()");
+const event1Overloaded = parseAbiItem("event Event1(bytes32 indexed)");
+const eventFactory = parseAbiItem("event EventFactory(address indexed child)");
+
+const address1 = "0x0000000000000000000000000000000000000001";
+const address2 = "0x0000000000000000000000000000000000000001";
+const bytes1 =
+  "0x0000000000000000000000000000000000000000000000000000000000000001";
+const bytes2 =
```
kyscott18, line 19: bytes 2 is the same as bytes 1

kyscott18 (approved):

PR #549, opened 2023-12-21
Fixes TypeScript compatibility + refactor config types

### LINKED ISSUE #541: Node v.18.0.0 is unsupported
image

**Problem**

Attempting to create a project with Node v18.0.0 gives me the following error `os.AvailableParallelism is not a function`.

**Expectation**

Either Node 18 should be supported, or an error message should be provided letting me know that it's not supported.

**Fix**

Upgrading to latest Node version worked.

kyscott18:
Fixes peer dependency status of TypeScript and viem.

Refactor config for more verbosity and compliance with older TypeScript versions

Also, closes #541


package.json
```diff
@@ -30,7 +30,8 @@
     "lint-staged": "^15.1.0",
     "prettier": "^3.1.0",
     "simple-git-hooks": "^2.9.0",
-    "typescript": "^5.3.3"
+    "typescript": "^5.3.3",
+    "viem": "^1.20.3"
```
0xOlias, line 34: Do we need viem here (at the top level)?

packages/core/src/config/config.ts
```diff
@@ -193,7 +193,7 @@ export type Network = {
    * }
    * ```
    */
-  transport: Transport;
+  transport: Transport | unknown;
```
0xOlias, line 196: What did this do?

package.json
kyscott18: I believe it helps with resolving the specific version instead of letting pnpm pick the version. I was using wagmi + viem as a reference for all of this, and they do the same.

packages/core/src/config/config.ts
kyscott18: Solves a very common problem with mismatched types in viem. `transport` is a unique case because the user is often importing viem, then using `http()` and passing it to the config.

packages/core/src/config/config.ts
```diff
@@ -200,70 +55,89 @@ export type Network = {
...
-  contracts: TContracts &
-    Record<string, Contract<string, Abi, string, AbiEvent>>;
-  options?: Option;
-}): {
-  database?: Database;
-  networks: TNetworks;
-  contracts: TContracts;
-  options?: Option;
-} => config;
+export type CreateConfigReturnType<networks, contracts> = {
```
0xOlias, line 138: What's the purpose of this type?

packages/core/src/config/utilityTypes.ts
```diff
@@ -0,0 +1,51 @@
+import type { Abi, AbiEvent, FormatAbiItem } from "abitype";
```
0xOlias, line 1: I wonder if this file should go into the `types/` folder. Either way, we should come up with a convention for what to name and where to put type-only files.

packages/core/src/config/utilityTypes.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export type ExtractAbiEvents<
+  abi extends Abi,
+  events = Extract<abi[number], { type: "event" }>,
+> = [events] extends [never] ? AbiEvent : events;
+
+export type ParseAbiEvent<
+  abi extends Abi,
+  signature extends string,
+  ///
```
0xOlias, line 16: ?

packages/core/src/types/ponder.ts
```diff
@@ -1,67 +1,59 @@
...
+  [name in keyof contracts]: `${name & string}:${
+    | (SafeEventNames<contracts[name]["abi"]> extends infer events extends
+        string
+        ? contracts[name]["filter"] extends {
+            event: infer filterEvents extends string | readonly string[];
           }
           ? Extract<
               events,
-              filterEvents extends string[]
+              filterEvents extends readonly string[]
```
0xOlias, line 24: Lol at this diff

packages/create-ponder/package.json
```diff
@@ -41,10 +41,9 @@
     "abitype": "^0.10.2",
     "dotenv": "^16.3.1",
     "tsup": "^8.0.1",
-    "typescript": "^5.3.2",
```
0xOlias, line 44: Is this because we're using the top-level monorepo typescript?

packages/create-ponder/src/index.ts
```diff
@@ -426,10 +426,16 @@ export async function run({
     .help();
 
   // Check Nodejs version
-  if (Number(process.version.split(".")[0].slice(1)) < 18)
+  const _nodeVersion = process.version.split(".");
```
0xOlias, line 429: Is there anywhere in `create-ponder` where we're generating the package.json (not just copying from templates) that needs to be updated?

0xOlias (approved):
LGTM

packages/create-ponder/package.json
kyscott18: Yes

packages/core/src/config/address.ts
```diff
@@ -0,0 +1,50 @@
+import type { AbiEvent } from "abitype";
+
+// TODO: support unnamed parameter
```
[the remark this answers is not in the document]
kyscott18: Note to make sure that unnamed event arguments are behaving as expected

packages/core/src/config/utilityTypes.ts
kyscott18: This is a convention I took from abitype/viem. Basically, the type arguments under the "///" are actually just variables. These arguments should never be passed in and instead the type falls back to the default value assigned by the "=".

packages/core/src/bin/ponder.ts
```diff
@@ -127,7 +127,13 @@ function registerKilledProcessListener(fn: () => Promise<unknown>) {
 }
 
 function validateNodeVersion() {
```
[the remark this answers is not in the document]
kyscott18: The "engine" field of package.json only throws a warning when you install the package, so it's pretty easy to miss.

packages/core/src/config/config.ts
kyscott18: It enforces maximum strictness. The type "Config" is not as strict because we want the nice language server features when hovering over a property.

packages/create-ponder/src/index.ts
kyscott18: Good thought but no.

0xOlias (approved):
LGTM

PR #552, opened 2023-12-24
Implement global request throttling

kyscott18:
## Summary

Requests from historical sync, realtime sync, and indexing services are all passed through a global request queue, to make sure not to overload the rpc.

## Goals
- Make ponder more compatible with less capable RPCs
- Remove AbortController logic
- Separate the logic for requests and db writes in the historical sync service
- Support all viem transports

## Solution

- requestQueue per network that limits maxRequestsPerSecond, depending on config value, default = 100
- use viem transport straight from config, it now handles retry and timeout logic
- kill request queue by simultaneously rejecting all promises, no more abort controller logic

## Other minor updates

- refactor ponderTransport
- finalized block can be undefined, representing no finalized blocks
- rpc evaluation script
- local historical sync benchmark with vitest bench
- sync store retry logic handled in `transaction()` helper function
- remove extra logic from old queue
- metrics for request duration moved to request queue


packages/core/src/sync-historical/service.ts
```diff
@@ -463,60 +518,21 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
...
+        t.isComplete(),
+      ) &&
+      Object.values(this.factoryChildAddressProgressTrackers).every((t) =>
+        t.isComplete(),
+      ) &&
+      Object.values(this.factoryLogFilterProgressTrackers).every((t) =>
+        t.isComplete(),
+      ) &&
+      this.blockProgressTracker.isComplete()
+    ) {
```
0xOlias, line 535: Could perhaps extract this into a `private isSyncComplete()` function which could also be used at the end of `start()`.

packages/core/src/sync-historical/service.ts
```diff
@@ -529,164 +545,231 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
...
+  private runTask = async (task: HistoricalSyncTask) => {
+    try {
+      if (task.kind === "LOG_FILTER") await this.logFilterTaskWorker(task);
+      else if (task.kind === "FACTORY_LOG_FILTER")
+        await this.factoryLogFilterTaskWorker(task);
+      else if (task.kind === "FACTORY_CHILD_ADDRESS")
+        await this.factoryChildAddressTaskWorker(task);
+      else if (task.kind === "BLOCK") await this.blockTaskWorker(task);
+    } catch (error_) {
+      if (error_ === undefined) return;
```
0xOlias, line 560: When would this happen?

packages/core/src/sync-realtime/service.test.ts
```diff
@@ -39,6 +37,7 @@ test("setup() returns block numbers", async (context) => {
   expect(finalizedBlockNumber).toBe(blockNumbers.finalizedBlockNumber);
 
   await service.kill();
+  networks[0].requestQueue.clear();
```
0xOlias, line 40: Why is this required?

0xOlias (approved):
LGTM

packages/core/src/sync-historical/service.ts
```diff
@@ -445,16 +487,29 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
...
+      ) &&
+      Object.values(this.factoryChildAddressProgressTrackers).every((t) =>
+        t.isComplete(),
+      ) &&
+      Object.values(this.factoryLogFilterProgressTrackers).every((t) =>
+        t.isComplete(),
+      ) &&
+      this.blockProgressTracker.isComplete()
+    ) {
+      if (this.finalizedBlockNumber !== undefined) {
```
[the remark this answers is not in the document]
kyscott18: It could be undefined in the case when the network finality is 32 and the chain is only 16 blocks long.

packages/core/src/sync-historical/service.ts
kyscott18: When killing the requestQueue, all promises are immediately rejected, leading to the error being undefined.

packages/core/src/sync-realtime/service.test.ts
kyscott18: Resets the timing on the queue so the next test doesn't have to wait for the time buffer

packages/core/src/sync-historical/service.ts
```diff
@@ -148,23 +160,280 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
...
-    finalizedBlockNumber: number;
+    finalizedBlockNumber: number | undefined;
   }) {
     // Initialize state variables. Required when restarting the service.
-    this.isKilling = false;
     this.blockTasksEnqueuedCheckpoint = 0;
-
     this.finalizedBlockNumber = finalizedBlockNumber;
 
+    this.network.requestQueue.pause();
```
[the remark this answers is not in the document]
kyscott18: Had the same thought, but waiting to do this when we do some work on realtime sync service, because that will also change this behavior.

packages/core/src/sync-historical/service.ts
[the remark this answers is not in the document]
0xOlias: Got it. I wonder if it should get set to 0 in that case.

PR #563, opened 2024-01-08
docs: Fix foundry link + import ABI the right way 

carletex:
Hey there!

I've been playing with ponder these last few days: love it, it's a great tool! Thanks for all the effort.

Creating this PR with a couple of _issues_ I found in the Docs:

1. On the `Advanced` page, the foundry link was broken. I also moved it to the bottom (as in the submenu)
2. The getting started guide `ponder.config.ts` example use a JSON path directly on the `abi` prop. Updated it to use the import as suggested in `Add Contracts` page.

Happy to tweak it further if needed.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,8 @@ Ponder fetches event logs for the contracts added to `ponder.config.ts`, and pas
 // ponder.config.ts
 import { http } from "viem";
 
+import { BaseRegistrarAbi } from "./abis/BaseRegistrar";
+
 export const config = {
   networks: [
     {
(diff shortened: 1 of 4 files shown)


0xOlias (approved):
This is awesome, thanks so much for contributing.

I just noticed that the code snippets in `README.md` and `packages/core/README.md` had the same outdated ABI file path code - would you be down to fix those too?

carletex:
> I just noticed that the code snippets in README.md and packages/core/README.md had the same outdated ABI file path code - would you be down to fix those too?

Done in https://github.com/ponder-sh/ponder/pull/563/commits/e9b25b564427dc3cd0019d2db0e16b9e03791080!

I also noticed that both READMEs were still using `schema.graphql` (instead of `ponder.schema.ts`) so I tweaked it on https://github.com/ponder-sh/ponder/pull/563/commits/fe91ed6fe0521a5994018448856dc6602ff0b4cb

Feel free to tweak as you see fit.

Thanks!

PR #568, opened 2024-01-10
Add heartbeat to telemetry service; improve telemetry flush handling

### LINKED ISSUE #560: [Bug] Telemetry service network error
image

grayleonard:
To avoid bugs and complications of getting the path for the script to run in detached process, I write it to tmp and call it there. Also can bump up heartbeat rate from 60sec to like 10min, depending on what you expect average runtime to be (forever?).

The bug preventing the flush from running was that the path was sometimes incorrect depending on how ponder was being called, and it was failing silently.

Fixes #560


packages/core/src/telemetry/service.ts
```diff
@@ -16,8 +15,6 @@ import process from "process";
 import type { Options } from "@/config/options.js";
 import { getGitRemoteUrl } from "@/telemetry/remote.js";
 
-const __dirname = url.fileURLToPath(new URL(".", import.meta.url));
```
0xOlias, line 19: A quick search yielded this solution for ESM:
```
import { fileURLToPath } from "node:url";
import path from "node:path";

const __filename = fileURLToPath(import.meta.url);
const __dirname = path.dirname(__filename);
```
I'm not sure where I found the `new URL(".", ...)` bit. Perhaps that was causing the issue/inconsistency.

My preference would be to use the file rather than the string to temp file approach if we can. Do you know in which scenarios this old code was working and not working?

packages/core/src/telemetry/service.ts
```diff
@@ -100,7 +109,7 @@ export class TelemetryService {
       });
     } catch (e) {
       const error = e as { name: string };
-      if (error.name === "AbortError") {
+      if (error.name === "AbortError" || error.name === "TimeoutError") {
```
0xOlias, line 112: When might a `TimeoutError` happen here?

packages/core/src/telemetry/service.ts
```diff
@@ -146,13 +199,18 @@ export class TelemetryService {
...
+    //Write remaining events and flush script to tmp files
     fs.writeFileSync(telemetryEventsFilePath, serializedEvents);
+    fs.writeFileSync(detachedFlushScriptPath, this.detachedFlushScript());
 
+    //Spawn child
     child_process.spawn(process.execPath, [
-      path.join(__dirname, "detached-flush.js"),
+      detachedFlushScriptPath,
       this.options.telemetryUrl,
       telemetryEventsFilePath,
```
0xOlias, line 215: Not blocking, but I am interesting in a solution where we just pass the `serializedEvents` string as a CLI arg to the child process rather than write + read from a file.

0xOlias (commented):
Looking good, some quick ideas to try. Also, our convention now (for better or for worse) is for file names to use camelCase, so would be great to rename the `detached-flush.js` script here.

packages/core/src/telemetry/service.ts
grayleonard: It seems to break after build & minification, which causes `telemetry/detached-flush.js` to no longer be in the same path as `telemetry/service.ts`. So calling `import.meta.url` technically returns the correct path for where it's being called from, but we can't rely on it because we can't guarantee the two files will always be in the same dir. 

Old code seems to be failing more times than not due to this.

The ideal solution would be using `import.meta.resolve('./detached-script.js')` (https://nodejs.org/api/esm.html#importmetaresolvespecifier) but only stable in node 20+.

packages/core/src/telemetry/service.ts
```diff
@@ -69,6 +68,16 @@ export class TelemetryService {
...
 
+  private establishHeartbeat() {
+    this.heartbeatIntervalId = setInterval(() => {
+      this.record({ event: "Heartbeat" });
+    }, 60_000);
+  }
+
+  private killHeartbeat() {
+    clearInterval(this.heartbeatIntervalId);
+  }
```
0xOlias, line 79: Style nit: I'd probably just inline these (create the interval in the constructor, clear it in `kill()`)

0xOlias (approved):
LGTM! Couple nits and requests.

1. I pulled the branch down and opened up the file, looks like there's now an unused import for `import child_process from "node:child_process";`
2. While we're here, let's change the notify message to: `Ponder collects anonymous telemetry data to identify issues and prioritize features. See https://ponder.sh/advanced/telemetry for more information.`

0xOlias:
Two more things:
1. Should add "Fixes https://github.com/ponder-sh/ponder/issues/560" to the PR comment to get that sweet issue link
2. Should add a changeset, see guide here https://github.com/ponder-sh/ponder/issues/560

0xOlias (approved):
LGTM

PR #574, opened 2024-01-12
Global queue for all network requests

kyscott18:


packages/core/src/sync-historical/getLogFilterRetryRanges.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  );
+
+  const retryRanges = getLogFilterRetryRanges(error, "0x1", "0x3");
+
+  expect(retryRanges).toHaveLength(2);
+  expect(retryRanges[0]).toMatchObject(["0x1", "0x2"]);
+  expect(retryRanges[1]).toMatchObject(["0x3", "0x3"]);
+});
+
+test("start() handles Quicknode 'eth_getLogs and eth_newFilter are limited to a 10,000 blocks range' error", async () => {
```
0xOlias, line 19:
```diff
-start()
+getLogFilterRetryRanges
```

packages/core/src/sync-historical/getLogFilterRetryRanges.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    numberToHex(1),
+    numberToHex(20_000),
+  );
+
+  expect(retryRanges).toHaveLength(2);
+  expect(retryRanges[0]).toMatchObject([numberToHex(1), numberToHex(10000)]);
+  expect(retryRanges[1]).toMatchObject([
+    numberToHex(10_001),
+    numberToHex(20000),
+  ]);
```
0xOlias, line 37: Would be nice to consistently use (or not use) `_` in numbers

packages/core/src/sync-historical/service.bench.ts
```diff
@@ -0,0 +1,53 @@
...
+const getBlockNumbers = () =>
+  publicClient.getBlockNumber().then((b) => ({
+    latestBlockNumber: Number(b) + 5,
+    finalizedBlockNumber: Number(b),
+  }));
+
+bench(
+  "Historical sync benchmark",
+  async () => {
+    const teardownSync = await setupSyncStore(context);
```
0xOlias, line 29: So, you have to do it this way because `vitest bench` doesn't allow `beforeEach` ?

packages/core/src/sync-historical/service.ts
```diff
@@ -620,48 +625,115 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
...
+    const logs = await Promise.all(
+      childAddresses.map(async (c) =>
+        this._eth_getLogs({
+          address: c,
+          topics: factory.criteria.topics,
+          fromBlock: numberToHex(fromBlock),
+          toBlock: numberToHex(toBlock),
+        }),
+      ),
+    ).then((l) => l.flat());
```
0xOlias, line 687: I wonder if this should be properly batched, where we query a batch of addresses from the DB, make an eth_getLogs call, add a block callback, then repeat.

packages/core/src/sync-historical/service.ts
```diff
@@ -952,115 +939,97 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
...
-          Number(options.fromBlock),
-      );
-      retryRanges.push([toHex(options.fromBlock), toHex(midpoint)]);
-      retryRanges.push([toHex(midpoint + 1), options.toBlock]);
-    } else {
-      // Throw any unrecognized errors.
-      throw error;
+      return Promise.all(
+        retryRanges.map(([from, to]) =>
+          this._eth_getLogs({
```
0xOlias, line 978: Should be fine to `Promise.all` these because concurrency is managed by the queue, right?

packages/core/src/sync-historical/service.ts
```diff
@@ -952,115 +939,97 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
...
+        method: "eth_getBlockByNumber",
+        params: [numberToHex(params.blockNumber), true],
+      })
+      .then((block) => {
+        if (!block)
+          throw new BlockNotFoundError({
+            blockNumber: BigInt(params.blockNumber),
+          });
+        return block as HistoricalBlock;
       });
```
0xOlias, line 1006: What happens if this throws an error? Trying to get acquainted with `.then` syntax.

packages/core/src/utils/requestQueue.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  expect(await queue.size()).toBe(1);
+  expect(await queue.pending()).toBe(0);
+
+  await r2;
+
+  expect(await queue.size()).toBe(0);
+  expect(await queue.pending()).toBe(0);
+});
+
+test("add() returns promise", async ({ common }) => {
```
0xOlias, line 77: Typo (should be `request()`?)

packages/core/src/utils/requestQueue.ts
```diff
@@ -0,0 +1,168 @@
...
+        processQueue();
+      }, interval - timeSinceLastRequest);
+    }
+  };
+
+  return {
+    request: <TParameters extends EIP1193Parameters<PublicRpcSchema>>(
+      params: TParameters,
+    ): RequestReturnType<TParameters["method"]> => {
+      const stopClockLag = startClock();
```
0xOlias, line 136: If we're not using this `startClock` method anywhere else, I think it makes sense to remove it and use the proper `node` std perf/timer function.

packages/core/src/utils/requestQueue.ts
```diff
@@ -0,0 +1,168 @@
...
+            id += 1;
+            pending -= 1;
+
+            metrics.ponder_rpc_request_duration.observe(
+              { method: params.method, network: networkName },
+              stopClock(),
+            );
+          });
+
+        if (queue.length === 0) break;
```
0xOlias, line 117: This line can run before line 103 in the same loop iteration, right?

packages/core/src/build/config/config.ts
```diff
@@ -10,14 +11,18 @@ import {
...
   Factory,
   LogFilter,
   Source,
   Topics,
 } from "../../config/sources.js";
 
-export async function buildNetworksAndSources({ config }: { config: Config }) {
+export async function buildNetworksAndSources({
+  config,
+  common,
```
0xOlias, line 24: Don't have a great solution, but it's smelly to me that this function now needs `common`/`metrics`.

packages/core/src/sync-historical/service.bench.ts
kyscott18: yes

packages/core/src/sync-historical/service.ts
kyscott18: Sounds good, I was a little weary of this spot which is why I left the note

packages/core/src/utils/requestQueue.test.ts
kyscott18: Yes

packages/core/src/sync-historical/service.ts
kyscott18: If line 1002 throws an error, it gets passed up the stack because there is no `.catch()` callback. Same behavior as async/await.

packages/core/src/utils/requestQueue.ts
kyscott18: We still use `startClock` for measuring server responses.

packages/core/src/utils/requestQueue.ts
```diff
@@ -150,6 +178,14 @@ export const createRequestQueue = ({
       queue = new Array();
       lastRequestTime = 0;
     },
+    kill: () => {
+      // NOTE: Should this go in clear or pause instead?
+      clearTimeout(timeout);
+      on = false;
+      queue = new Array();
+      lastRequestTime = 0;
+      abortController.abort();
```
kyscott18, line 187: Does this work after it runs once? Basically does the abort controller need to be reset?

PR #575, opened 2024-01-12
fix 404 in foundry docs

jrfrantz:


0xOlias (approved):
LGTM thanks for the fix

PR #582, opened 2024-01-17
create-ponder subgraph ID support on parity with etherscan

grayleonard:
Pulls subgraph info from ipfs using subgraph ID and generates ponder config and populates ABIs. Doesn't generate schema based on events yet, and doesn't support subgraph factory templates because the factory's child creation logic is in assemblyscript, and not the subgraph configuration itself.


0xOlias (approved):
LGTM! Great to get this fixed.

PR #588, opened 2024-01-19
Concurrent indexing

kyscott18:


packages/core/src/sync-store/postgres/store.ts
```diff
@@ -994,137 +980,192 @@ export class PostgresSyncStore implements SyncStore {
...
+    // Get total count of matching logs, grouped by log filter and event selector.
+    const eventCountsQuery = baseQuery
+      .clearSelect()
+      .select([
+        "source_id",
+        "logs.topic0",
+        this.db.fn.count("logs.id").as("count"),
+      ])
+      .where((eb) => {
+        // NOTE: Not adding the includeEventSelectors clause here.
```
0xOlias: rm

packages/core/src/sync-gateway/service.test.ts
```diff
@@ -354,13 +351,13 @@ test("handleNewFinalityCheckpoint emits newFinalityCheckpoint if subsequent even
...
 
-  const { networks, sources } = getMultichainNetworksAndSources(context);
+  const { networks } = getMultichainNetworksAndSources(context);
   const [mainnet, optimism] = networks;
 
   const service = new SyncGateway({
     common,
     syncStore,
-    sources,
+
```
0xOlias, line 360: rm newline?

packages/core/src/indexing/service.ts
```diff
@@ -527,264 +712,277 @@ export class IndexingService extends Emittery<IndexingEvents> {
...
-          address: address ? checksumAddress(address) : address,
-          startBlock: source.startBlock,
-          endBlock: source.endBlock,
-          maxBlockRange: source.maxBlockRange,
-        },
-      },
-    };
-  });
+    for (const count of counts) {
+      const event =
```
0xOlias, line 955: Would consider doing this here:
`const source = this.sourceById[count.sourceId]!`

packages/core/src/indexing/service.ts
```diff
@@ -527,264 +712,277 @@ export class IndexingService extends Emittery<IndexingEvents> {
...
+        // Note: Assumption is that all sources with the same contract name have the same abi.
+        const i = this.sources.findIndex(
+          (s) =>
+            s.contractName === contractName &&
+            s.abiEvents.bySafeName[eventName] !== undefined,
+        );
+
+        const abiEvent = this.sources[i].abiEvents.bySafeName[eventName]!.item;
+        const eventSelector =
+          this.sources[i].abiEvents.bySafeName[eventName]!.selector;
```
0xOlias, line 915:
```ts
const abiEvent = this.sources[i].abiEvents.bySafeName[eventName]!
const abiItem = abiEvent.item;
const eventSelector = abiEvent.selector;
```
Or just inline the 2nd and 3rd lines here on 923/924

packages/core/src/indexing/service.ts
```diff
@@ -527,264 +712,277 @@ export class IndexingService extends Emittery<IndexingEvents> {
...
+  };
+
+  private buildSourceById = () => {
+    for (const source of this.sources) {
+      this.sourceById[source.id] = source;
     }
-  > = {};
+  };
+
+  private buildIndexingFunctionMap = () => {
```
0xOlias, line 854: I wonder if it's worth combining `this.indexingFunctions` and the top half of `this.indexingFunctionMap` into one `Record` that's keyed by event key (eg `"Contract:Event"`). This object could also be created in the build service.

Then, you could have another record keyed the same way calling smth like `indexingFunctionState`. This would enforce some degree of separation from the immutable and mutable pieces of state in the class which could be desirable. Low conviction comment though, take it or leave it.

packages/core/src/indexing/service.ts
```diff
@@ -527,264 +712,277 @@ export class IndexingService extends Emittery<IndexingEvents> {
...
+    const contractName = keyHandler.contractName;
+    const eventName = keyHandler.eventName;
+
+    for (const event of events) {
+      try {
+        const decodedLog = decodeEventLog({
+          abi,
+          data: event.log.data,
+          topics: event.log.topics,
+        });
```
0xOlias, line 793: Random thought - I wonder if there's any perf gain to be had by decoding all `events` first, then pushing them to the queue. Decoding could also happen in a Worker 😈

packages/core/src/indexing/service.ts
```diff
@@ -1,103 +1,134 @@
...
     networkName: string;
     contractName: string;
-    chainId: number;
-    blockNumber: number;
+    checkpoint: Checkpoint;
+  };
+};
+type LogEventTask = {
+  kind: "LOG";
+  event: {
```
0xOlias, line 59: Maybe call this `data`?

packages/core/src/indexing/service.ts
```diff
@@ -1,103 +1,134 @@
...
 
-  private eventProcessingMutex: Mutex;
-  private queue?: IndexingFunctionQueue;
+  queue?: IndexingFunctionQueue;
 
-  private eventsProcessedToCheckpoint: Checkpoint = zeroCheckpoint;
+  private getNetwork: (checkpoint: Checkpoint) => Context["network"] =
+    undefined!;
+  private getClient: (checkpoint: Checkpoint) => Context["client"] = undefined!;
+  private getDB: (checkpoint: Checkpoint) => Context["db"] = undefined!;
```
0xOlias, line 99: `getDb` for consistency (sorry)

packages/core/src/indexing/service.ts
```diff
@@ -1,103 +1,134 @@
...
+  private getContracts: (checkpoint: Checkpoint) => Context["contracts"] =
+    undefined!;
 
-  private currentIndexingCheckpoint: Checkpoint = zeroCheckpoint;
-  private isPaused = false;
-  /** If true, failed tasks should not log errors or be retried. */
-  private isShuttingDown = false;
+  private indexingFunctionMap: Record<
+    /* Indexing function key: "{ContractName}:{EventName}" */
+    string,
```
0xOlias, line 105: Wonder if it would be worth using branded types for some of these keys (rather than the comment). Would also be useful for self-documenting the `parents` field below, for example.

packages/core/src/indexing/service.ts
```diff
@@ -1,103 +1,134 @@
...
+    {
+      contractName: string;
+      eventName: string;
+      /* Indexing function keys that write to tables that this indexing function key reads from. */
+      parents: string[];
+      /* True if this key is a parent of itself. */
+      selfReliance: boolean;
```
0xOlias, line 112: Maybe `isSelfDependent`?

packages/core/src/indexing/service.ts
```diff
@@ -1,103 +1,134 @@
...
+      /* Checkpoint of max completed task. */
+      checkpoint: Checkpoint;
+      /* Checkpoint of the most recent task loaded from db. */
+      maxTaskCheckpoint: Checkpoint;
+      /* Buffer of in memory tasks that haven't been enqueued yet. */
+      indexingFunctionTasks: LogEventTask[];
+      /* Mutex ensuring tasks are loaded one at a time. */
+      dbMutex: Mutex;
+      /** True if a task has been enqueued with itself as the most recent checkpoint. */
+      serialQueue: boolean;
```
0xOlias, line 127:
```
      /* Checkpoint of max completed task. */
      tasksProcessedToCheckpoint: Checkpoint;
      /* Checkpoint of the most recent task loaded from db. */
      tasksLoadedToCheckpoint: Checkpoint;
      /* Buffer of in memory tasks that haven't been enqueued yet. */
      loadedTasks: LogEventTask[];
      /* Mutex ensuring tasks are not loaded twice. */
      loadingMutex: Mutex;
      /** Lock to prevent tasks from running concurrently. */
      selfDependentLock: boolean;
```

0xOlias (approved):
LGTM 👑

Issue #329, opened 2023-08-25
[Feature] Pagination for large data sets


Willyham:
We are building a system which indexes deposit events on the beacon chain from specific contracts. This will easily grow into the 10-50k+  events range. Ponder has a limit of `5000` for the `skip` parameter.

Is there any way to paginate through this data? Forgive the ignorance about the underlying system but this doesn't seem like too much of a stretch for the underlying postgres database if we could query it in batches of 1000.

0xOlias:
This is very reasonable (the limit/offset pagination was inherited from subgraphs, but no reason keep it around). The purpose of the `1000`/`5000` validation is to discourage using limit/offset pagination for large result sets, which can be very bad for performance.

We should probably support cursor pagination for this purpose, which would be more performant. Would that work for your use case?

I think it'd be a relatively light lift to add cursor pagination to the `findMany` store method that backs the plural GraphQL fields.

Willyham:
Cursor pagination would be perfect. We have hacked our way around the issue with a crude implementation of that from the client side:

```
query GetFoos($upToBlock: Int!, $previousCursor: Int!, $perBatch: Int!) {
  foos(first: $perBatch, where: { block_lte: $upToBlock, block_gte: $previousCursor }, orderBy: "block", orderDirection: "asc") {
    id
    block
    foo
  }
}
```

0xOlias:
@eliobricenov and I chatted through a design and implementation notes. We would need to introduce a breaking change to replace the current limit/offset pagination with cursor-based pagination.

Before
```graphql
{
  transferEvents(first: 5, skip: 10) {
    id
    amount
  }
}
```

After
```graphql
{
  transferEvents(after: "ACX4YS=") {
    items {
      id
      amount
    }
    before
    after
  }
}
```

In the new design, all top-level plural GraphQL types will return a Page object rather than the raw list. The Page object will look like:

```graphql
type TransferEventPage {
  before: String
  after: String
  totalCount: Int! # Maybe include this?
  items: [TransferEvent!]!
}
```

where the `items` field includes the original/raw list that Ponder currently returns.

### Details

For cursor pagination to work, the underlying dataset must have a consistent sort order enforced. We can do this by relying on the unique `id` field that all Ponder entities already have.

If the user does not specify an `orderBy`/`orderDirection`, a default of `orderBy: "id", orderDirection: "asc"` will be applied (this is already the case).

[...]

PR #589, opened 2024-01-22
GraphQL Cursor-based Pagination

grayleonard:
This PR introduces breaking changes to the GraphQL server a la #329, which specs out cursor-based querying instead of skip-based. This greatly improves querying performance and enables searching through whole tables instead of only up to 5000 rows (what it was before).

Now, GraphQL queries return a generated Page type per table; for table `TransferEvent` the Page type will be
```js
type TransferEventPage {
  items: [TransferEvent!]!
  before: String
  after: String
}
```

`before` and `after` will either be a cursor string or `null`. If `before` or `after` are `null`, you can infer that there is either no next page or no previous page.

To use these cursors in a query, you now include either a `before` or `after` argument and an optional `limit` argument. If no cursor string is included, then the query will return results starting from the first row.

```
query Query {
  transferEvents(
    limit: 10
    after: "MHgyMmY0MjhmNDM0NDNkZTNjODg2ZjRkOWQwZTU0ODFjYjdjNmY3MzNkYzA1OTVkZGViMTM5M2ZkYTI3NjE5ZWQ3LTB4MTAw"
  ) {
    items {
      id
    }
    after
    before
  }
}
```

Comments / concerns welcome


packages/core/src/indexing-store/sqlite/store.ts
```diff
@@ -338,6 +342,196 @@ export class SqliteIndexingStore implements IndexingStore {
...
+   * ]
+   */
+
+  findManyPaginated = async ({
+    tableName,
+    checkpoint = "latest",
+    where,
+    before,
+    after,
+    take,
```
0xOlias, line 391: I think we should pick one of `take`, `limit`, and `first` and use it everywhere. Personally, I'm partial to `limit`. WDYT?

packages/core/src/indexing-store/sqlite/store.ts
```diff
@@ -338,6 +342,196 @@ export class SqliteIndexingStore implements IndexingStore {
...
+   *    id: 0
+   * ]
+   * 2) reverse()
+   * [
+   *    id: 0,
+   *    id: 1
+   * ]
+   */
+
+  findManyPaginated = async ({
```
0xOlias, line 385: Rather than add a new method, I think we should just change the API of `findMany` to use cursor pagination.

packages/core/src/indexing-store/sqlite/store.ts
```diff
@@ -338,6 +342,196 @@ export class SqliteIndexingStore implements IndexingStore {
...
+        query = query.limit(limit + 1);
+      }
+
+      let orderDirection = "asc";
+
+      if (orderBy) {
+        // Get reversed order conditions, for finding the -1 row
+        const orderByConditions = !before
+          ? buildSqlOrderByConditions({ orderBy })
+          : buildSqlOrderByConditionsReversed({ orderBy });
```
0xOlias, line 443: Could we use the existing `buildSqlOrderByConditions` and then reverse the direction(s) here?

packages/core/src/indexing-store/sqlite/store.ts
```diff
@@ -338,6 +342,196 @@ export class SqliteIndexingStore implements IndexingStore {
...
+
+        orderDirection = orderByConditions[0][1];
+
+        for (const orderByCondition of orderByConditions) {
+          query = query.orderBy(...orderByCondition);
+        }
+        //query = query.orderBy(`id ${orderDirection}`);
```
0xOlias, line 450: rm

packages/core/src/indexing-store/sqlite/store.ts
```diff
@@ -338,6 +342,196 @@ export class SqliteIndexingStore implements IndexingStore {
...
+          const currentRowQuery = this.db
+            .selectFrom(table)
+            .selectAll()
+            .limit(1)
+            .where(
+              "id",
+              "=",
+              Buffer.from(after || before || "", "base64").toString(),
+            );
+          const res = await tx.executeQuery(currentRowQuery);
```
0xOlias, line 464: You can build and execute the query in one go like this:
```ts
const result = await tx.selectFrom(table)
  .selectAll()
  // ...
```

packages/core/src/indexing-store/sqlite/store.ts
```diff
@@ -338,6 +342,196 @@ export class SqliteIndexingStore implements IndexingStore {
...
+      });
+
+      let deserializedRows = rows.map((row) =>
+        this.deserializeRow({ tableName, row }),
+      );
+      if (before) {
+        deserializedRows.reverse();
+      }
+
+      const hasAfter = rows.length > (take || 1000);
```
0xOlias, line 505: If I'm reading correctly, this represents "has either before or after", correct?

Considering how different the logic is, and that they're mutually exclusive, I wonder if it would be more readable to write a branch for `after`, `before`, and neither.

0xOlias (commented):
Looking good!

packages/core/src/indexing-store/utils/filter.ts
```diff
@@ -0,0 +1,153 @@
...
+        throw new Error(
+          `Invalid filter condition for column '${columnName}'. Got '${condition}', expected one of [${allowedConditions
+            .map((c) => `'${c}'`)
+            .join(", ")}]`,
+        );
+      }
+
+      const filterEncodingFn = filterEncodingMap[condition];
+
+      console.log(typeof value);
```
kyscott18, line 134: Get rid of this

packages/core/src/indexing-store/utils/filter.ts
```diff
@@ -0,0 +1,153 @@
...
+import type { ComparisonOperatorExpression } from "kysely";
+import type { Table, WhereInput } from "../store.js";
+import { encodeValue } from "./encoding.js";
+
+const filterValidityMap = {
+  boolean: {
+    singular: ["equals", "not", "in", "notIn"],
+    list: ["equals", "not", "has", "notHas"],
+  },
+  string: {
```
kyscott18, line 12: What are enums encoded as?

packages/core/src/indexing-store/utils/cursor.ts
```diff
@@ -0,0 +1,71 @@
...
+export function encodeCursor(
+  record: Row,
+  orderByConditions: OrderByConditions,
+) {
+  const cursorValues = orderByConditions.map(([columnName]) => [
+    columnName,
+    record[columnName],
+  ]);
+
+  return Buffer.from(serialize(cursorValues)).toString("base64");
```
kyscott18, line 15: What is the purpose of encoding as a buffer instead of just using the string returned from `serialize`?

packages/core/src/indexing-store/utils/sort.ts
```diff
@@ -0,0 +1,59 @@
...
+        .join(", ")}]`,
+    );
+  }
+  if (column._type === "m" || column._type === "o") {
+    throw Error(
+      `Invalid sort. Cannot filter on virtual column '${columnName}'.`,
+    );
+  }
+
+  if (orderDirection === undefined || !["asc", "desc"].includes(orderDirection))
```
kyscott18, line 45: First expression isn't needed

packages/core/src/indexing-store/sqlite/store.ts
```diff
@@ -301,37 +310,154 @@ export class SqliteIndexingStore implements IndexingStore {
...
+      for (const [column, direction] of orderByConditions) {
+        query = query.orderBy(column, direction);
       }
 
-      if (take) {
-        const limit = validateTake(take);
-        query = query.limit(limit);
+      if (limit > MAX_LIMIT) {
+        throw new Error(
+          `Invalid limit. Got <=${MAX_LIMIT}, received ${limit}.`,
```
kyscott18, line 329: `Invalid limit. Got ${limit}, but expected <=${MAX_LIMIT}`

packages/core/src/types/model.ts
```diff
@@ -84,10 +84,16 @@ export type DatabaseModel<T extends { id: string | number | bigint }> = {
 
   findMany: (options?: {
     where?: Prettify<WhereInput<T>>;
-    skip?: number;
-    take?: number;
-    orderBy?: OrderByInput<T>; // TODO: Prettify/flatten this type. Couldn't get it working.
-  }) => Promise<Prettify<T>[]>;
+    // TODO: Prettify/flatten this type. Couldn't get it working.
```
kyscott18, line 87: I gotchu

packages/core/src/server/graphql/entity.ts
```diff
@@ -88,41 +86,66 @@ export const buildEntityTypes = ({ schema }: { schema: Schema }) => {
...
+                where: whereObject,
+                orderBy: orderByObject,
+                limit,
+                before,
+                after,
               });
             };
+            const pageType = new GraphQLObjectType({
+              name: `${tableName}PageChild`,
+              fields: () => ({
```
kyscott18, line 128: Why callback?

kyscott18 (approved):

packages/core/src/indexing-store/utils/cursor.ts
0xOlias: Cursors should not be human/client-readable. You don't want to encourage clients of the API to use the cursor value for anything other than passing it back to the server. So, it's best to encode it as an opaque string.

packages/core/src/indexing-store/utils/sort.ts
0xOlias: Is convenient for type checking on line 50

packages/core/src/indexing-store/utils/filter.ts
0xOlias: Ended up treating them the exact same way as strings.

PR #591, opened 2024-01-23
Prettify types of log, block, transaction

kyscott18:
ponder is a work of art, this is another stroke on the canvas


0xOlias (approved):
LGTM

PR #598, opened 2024-01-25
p.hex()

kyscott18:
Implement p.hex() as a replacement to p.bytes(). p.hex() encodes hex strings as the "blob" type in SQLite and "bytea" type in Postgres. p.hex() is case insensitive and returns lowercase values.

Internally, refactored the indexing store encoding and decoding logic, such that encoding validates columns against the respective column type.


kyscott18:
Still need to determine the proper "filtering" behavior in the server.

packages/core/src/indexing-store/postgres/store.ts
```diff
@@ -449,36 +448,42 @@ export class PostgresIndexingStore implements IndexingStore {
...
-        let updateRow: ReturnType<typeof formatRow>;
+        let updateRow: ReturnType<typeof encodeRow>;
         if (typeof data === "function") {
-          const current = this.deserializeRow({
-            tableName,
-            row: latestRow,
-          });
+          const current = decodeRow(
+            latestRow,
+            this.schema!.tables[tableName],
```
0xOlias, line 472: For the functions that use this many times, I'd consider doing a `const table = this.schema!.tables[tableName]` above.

packages/core/src/indexing-store/postgres/store.ts
```diff
@@ -510,11 +508,7 @@ export class PostgresIndexingStore implements IndexingStore {
         // we need to update the latest version AND insert a new version.
         await tx
           .updateTable(table)
-          .where(
-            "id",
-            this.idColumnComparator({ tableName, schema: this.schema }),
-            formattedId,
-          )
+          .where("id", "=", formattedId)
```
0xOlias, line 511: This is such a nice change.

packages/core/src/indexing-store/postgres/store.ts
```diff
@@ -883,12 +819,4 @@ export class PostgresIndexingStore implements IndexingStore {
     );
     return result;
   };
-
-  private idColumnComparator = ({
```
0xOlias, line 887: LFG love that we can kill this.

packages/core/src/indexing-store/store.test.ts
```diff
@@ -978,59 +977,59 @@ test("revert() updates versions that only existed during the safe timestamp to l
...
 
   const instance = await indexingStore.findUnique({
     tableName: "table",
     checkpoint: createCheckpoint(25),
-    id: "0xA",
+    id: "0x0A",
   });
-  expect(instance).toMatchObject({ id: "0xa", n: 1 });
+  expect(instance).toMatchObject({ id: "0x0a", n: 1 });
 });
```
0xOlias, line 997: Great test!

packages/core/src/indexing-store/utils/format.ts
```diff
@@ -1,43 +1,182 @@
-import { encodeAsText } from "@/utils/encoding.js";
-
+import type {
```
0xOlias, line 1: Maybe rename this file to `encoding.ts`? Could also move it up a level and get rid of the `utils` folder.

packages/core/src/utils/pg.ts
```diff
@@ -1,20 +1,19 @@
...
 // https://stackoverflow.com/a/70601114
 const originalClientQuery = pg.Client.prototype.query;
 // @ts-ignore
-pg.Client.prototype.query = async function query(
+pg.Client.prototype.query = function query(
   ...args: [queryText: string, values: any[], callback: () => void]
 ) {
   try {
-    return await originalClientQuery.apply(this, args);
+    return originalClientQuery.apply(this, args);
```
0xOlias, line 16: Will verify that this still works. IIRC I tried this before and it broke.

0xOlias (approved):
This is awesome. Mostly nits.

Issue #593, opened 2024-01-24
Column `chainId` type should be `bigint`


wuminzhe:
postgres type `integer` is 4 bytes length with a range between -2147483648 and +2147483647.

1. But,some evm's chain ids are bigger than this. https://chainid.network/chains_mini.json
2. And, the geth use bigint. https://github.com/ethereum/go-ethereum/blob/6b0de79935110fb5f63a60288191848dd98980ea/params/config.go#L42

0xOlias:
Thanks for opening. Did you run into an error related to this?

wuminzhe:
> Thanks for opening. Did you run into an error related to this?

yes, I did.

PR #611, opened 2024-02-01
Use `int8` for chain IDs in Postgres

0xOlias:
Adds a migration for Postgres chain ID columns from `int4` -> `int8`. Now, Postgres should behave the same as SQLite and can safely store chain IDs <= `Number.MAX_SAFE_INTEGER`. This also aligns Postgres with our checkpoint encoding logic, which also allows 16 digits.

Couple notes:
- Postgres' `bigint` is the same as `int8`.
- I added a pg-types parser for `bigint` <-> `Number`.
- We store our concept of "bigints" (EVM uint256) using `numeric(78,0)`, not `bigint`.

Thanks a bunch @wuminzhe! Fixes https://github.com/ponder-sh/ponder/issues/593.


kyscott18 (approved):
Looks great

PR #614, opened 2024-02-01
Cached indexing

0xOlias:


packages/core/src/build/static/ids.ts
```diff
@@ -1,45 +1,46 @@
...
-import { isEnumColumn, isManyColumn, isOneColumn } from "@/schema/utils.js";
+import { isBaseColumn, isEnumColumn } from "@/schema/utils.js";
 import { dedupe } from "@/utils/dedupe.js";
 import type { TableAccess } from "./parseAst.js";
 
 export type FunctionIds = { [func: string]: string };
 export type TableIds = { [table: string]: string };
 
+type FunctionInputs = {
+  [key: string]: {
```
0xOlias, line 12: Might be a matter of preference, but I think it's nice to name the key here like:
```ts
type FunctionInputs = {
  [functionId: string]: {
    astHash: string;
    sources: object[];
  }
};
```
Could even do something like this (not sure I like it):
```ts
type FunctionId = string

type FunctionInputs = {
  [id: FunctionId]: {
    astHash: string;
    sources: object[];
  }
};
```

packages/core/src/build/static/ids.ts
```diff
@@ -60,68 +61,135 @@ export const getIds = ({
...
+    functionIds,
+  });
+
   return { tableIds, functionIds };
 };
 
+/**
+ * Deterministically resolve all dependencies of a given function or table.
+ */
+const resolveDependencies = ({
```
0xOlias, line 117: Bit strange to me that this helper handles both functions and tables - seems like the logic is pretty different, might make sense to separate into two helpers?

PR #619, opened 2024-02-03
Efficient Realtime Sync

### LINKED ISSUE #228: [Feature] Alternative realtime sync strategy for fast chains

kyscott18:
Refactor realtime sync service to use less rpc credits under all scenarios.

Closes #228.

### DIFF
--- a/packages/core/src/sync-realtime/service.ts
+++ b/packages/core/src/sync-realtime/service.ts
@@ -280,19 +280,25 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
       return;
     }
 
-    // Note: should this be after the re-org
     const sync = this.determineSyncPath(newBlock);
-    if (sync === "traverse") await this.syncTraverse(newBlock);
-    else await this.syncBatch(newBlock);
+    const syncedData =
+      sync === "traverse"
+        ? await this.syncTraverse(newBlock)
+        : await this.syncBatch(newBlock);
 
-    const latestBlockNumber = hexToNumber(newBlock.number);
+    if (syncedData) {
+      this.logs.push(...syncedData.logs.map(realtimeLogToLightLog));
+      this.blocks.push(...syncedData.blocks.map(realtimeBlockToLightBlock));
+    }
 
     // If this block moves the finality checkpoint, remove now-finalized blocks from the local chain
     // and mark data as cached in the store.
-    if (
+
+    const latestBlockNumber = hexToNumber(newBlock.number);
+    const blockMovesFinality =
       latestBlockNumber >=
-      this.finalizedBlockNumber + 2 * this.network.finalityBlockCount
-    ) {
+      this.finalizedBlockNumber + 2 * this.network.finalityBlockCount;
+    if (blockMovesFinality) {
       if (!this.isLocalChainConsistent(latestBlockNumber)) {
         await this.detectReorg(latestBlockNumber);
       }
@@ -337,6 +343,27 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
       });
     }
 
+    if (syncedData)
+      await this.insertRealtimeBlocks(syncedData.logs, syncedData.blocks);
+
+    const newBlockNumber = hexToNumber(newBlock.number);
+    const newBlockTimestamp = hexToNumber(newBlock.timestamp);
+
+    this.emit("realtimeCheckpoint", {
+      blockTimestamp: newBlockTimestamp,
+      chainId: this.network.chainId,
+      blockNumber: newBlockNumber,
+    });
+
+    this.common.metrics.ponder_realtime_latest_block_number.set(
+      { network: this.network.name },
+      newBlockNumber,
+    );
+    this.common.metrics.ponder_realtime_latest_block_timestamp.set(
+      { network: this.network.name },
+      newBlockTimestamp,
+    );
+
     this.common.logger.debug({
       service: "realtime",
       msg: `Finished syncing new head block ${hexToNumber(
@@ -372,14 +399,15 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
     return batchCost > traverseCost ? "traverse" : "batch";
   };
 
-  private syncTraverse = async (newBlock: RealtimeBlock) => {
+  private syncTraverse = async (
+    newBlock: RealtimeBlock,
+  ): Promise<{ blocks: RealtimeBlock[]; logs: RealtimeLog[] } | undefined> => {
     const latestBlock = this.getLatestLocalBlock();
     const latestBlockNumber = latestBlock
       ? latestBlock.number
       : this.finalizedBlockNumber;
 
     const newBlockNumber = hexToNumber(newBlock.number);
-    const newBlockTimestamp = hexToNumber(newBlock.timestamp);
 
     const missingBlockRange = range(latestBlockNumber + 1, newBlockNumber);
     const newBlocks = await Promise.all(
@@ -412,34 +440,19 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
         true,
       );
 
-      // Insert logs + blocks + transactions
-      await this.insertRealtimeBlocks(matchedLogs, newBlocks);
-    }
-
-    this.emit("realtimeCheckpoint", {
-      blockTimestamp: newBlockTimestamp,
-      chainId: this.network.chainId,
-      blockNumber: newBlockNumber,
-    });
-
-    this.common.metrics.ponder_realtime_latest_block_number.set(
-      { network: this.network.name },
-      newBlockNumber,
-    );
-    this.common.metrics.ponder_realtime_latest_block_timestamp.set(
-      { network: this.network.name },
-      newBlockTimestamp,
-    );
+      return { blocks: newBlocks, logs: matchedLogs };
+    } else return undefined;
   };
 
-  private syncBatch = async (newBlock: RealtimeBlock) => {
+  private syncBatch = async (
+    newBlock: RealtimeBlock,
+  ): Promise<{ blocks: RealtimeBlock[]; logs: RealtimeLog[] }> => {
     const latestBlock = this.getLatestLocalBlock();
     const latestBlockNumber = latestBlock
       ? latestBlock.number
       : this.finalizedBlockNumber;
 
     const newBlockNumber = hexToNumber(newBlock.number);
-    const newBlockTimestamp = hexToNumber(newBlock.timestamp);
 
     // Get logs
     const logs = await this._eth_getLogs({
@@ -463,23 +476,7 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
     );
     blocks.push(newBlock);
 
-    // Insert logs + blocks + transactions
-    await this.insertRealtimeBlocks(matchedLogs, blocks);
-
-    this.emit("realtimeCheckpoint", {
-      blockTimestamp: newBlockTimestamp,
-      chainId: this.network.chainId,
-      blockNumber: newBlockNumber,
-    });
-
-    this.common.metrics.ponder_realtime_latest_block_number.set(
-      { network: this.network.name },
-      newBlockNumber,
-    );
-    this.common.metrics.ponder_realtime_latest_block_timestamp.set(
-      { network: this.network.name },
-      newBlockTimestamp,
-    );
+    return { blocks: blocks, logs: matchedLogs };
   };
 
   /**
@@ -623,9 +620,6 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
       });
     }
 
-    this.logs.push(...logs.map(realtimeLogToLightLog));
-    this.blocks.push(...blocks.map(realtimeBlockToLightBlock));
-
     this.lastLogsPerBlock = this.logs.length / this.blocks.length;
   };
 
@@ -692,7 +686,6 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
 
   /** Returns true if "blocks" has a valid chain of block.parentHash to block.hash. */
   private isLocalChainConsistent = (latestBlockNumber: number): boolean => {
-    if (this.blocks.length === 0) return false;
     if (this.blocks[this.blocks.length - 1].number !== latestBlockNumber)
       return false;
 
--- a/packages/core/src/sync-realtime/service.ts
+++ b/packages/core/src/sync-realtime/service.ts
@@ -54,7 +54,7 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
   private logFilterSources: LogFilter[];
   private factorySources: Factory[];
   private address: Address[] | undefined;
-  private topics: Hex[];
+  private eventSelectors: Hex[];
 
   private isProcessingBlock = false;
   private isProcessBlockQueued = false;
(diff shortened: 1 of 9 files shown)


packages/core/src/sync-realtime/service.ts
```diff
@@ -79,19 +94,48 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
-    let latestBlock: BlockWithTransactions;
+    let latestBlock: RealtimeBlock;
+    let finalizedBlock: RealtimeBlock;
     let rpcChainId: number;
     try {
-      [latestBlock, rpcChainId] = await Promise.all([
-        this.getLatestBlock(),
+      [latestBlock, finalizedBlock, rpcChainId] = await Promise.all([
+        this._eth_getBlockByNumber("latest"),
+        this._eth_getBlockByNumber("finalized"),
```
0xOlias, line 138: Not all networks support this, and I suspect the behavior might be inconsistent across providers (some might error, others might return `"latest"` instead, etc.

We could still consider using it if we think there's a real benefit, but would need to handle these cases and fall back to the old behavior.

packages/core/src/sync-realtime/service.ts
```diff
@@ -150,50 +193,24 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
-    // If the latest block was not added to the queue, setup was not completed successfully.
-    if (this.queue.size === 0) {
-      throw new Error(
-        "Unable to start. Must call setup() method before start().",
-      );
+      return;
     }
 
-    // Fetch the block at the finalized block number.
-    const finalizedBlock = await this.requestQueue.request({
```
0xOlias, line 172: Lol, made no sense to fetch latest in `setup` and finalized in `start`.

packages/core/src/sync-realtime/service.ts
```diff
@@ -202,444 +219,506 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
+    if (sync === "traverse") await this.syncTraverse(newBlock);
+    else await this.syncBatch(newBlock);
+
+    // If this block moves the finality checkpoint, remove now-finalized blocks from the local chain
+    // and mark data as cached in the store.
     if (
-      newBlock.number === previousHeadBlock.number + 1 &&
-      newBlock.parentHash === previousHeadBlock.hash
+      hexToNumber(newBlock.number) >
+      this.finalizedBlockNumber + 2 * this.network.finalityBlockCount
```
0xOlias, line 280: If we do choose to use the `"finalized"` tag, the `finalityBlockCount` logic would also need to be dynamic (I think).

packages/core/src/sync-realtime/service.ts
```diff
@@ -202,444 +219,506 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
 
-        matchedLogs = filterLogs({
-          logs,
-          logFilters: [
-            ...this.sources.filter(sourceIsLogFilter).map((l) => l.criteria),
-            ...factoryLogFilters,
-          ],
-        });
-      }
+    if (costBatch > costTraverse) return "traverse";
```
0xOlias, line 357: Might call these `batchCost` and `traverseCost`

packages/core/src/sync-realtime/service.ts
```diff
@@ -202,444 +219,506 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
+    else return "batch";
```
0xOlias, line 358: Ternary?

packages/core/src/sync-realtime/service.ts
```diff
@@ -202,444 +219,506 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
-        }
-      } else {
-        // The app has factory contracts.
-        // Don't attempt to skip calling eth_getLogs, just call it every time.
-        logs = await this.requestQueue.request({
-          method: "eth_getLogs",
-          params: [{ blockHash: newBlock.hash }],
-        });
+    // Probability of a log in a block
+    const pLog = Math.min(this.lastLogsPerBlock, 1);
```
0xOlias, line 349: Ohhh now he's cooking

packages/core/src/sync-realtime/service.ts
```diff
@@ -202,444 +219,506 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
-            }
-            return {
-              address: childContractAddresses,
-              topics: factory.criteria.topics,
-            };
-          }),
-        );
+    // Probability of no logs in the range of blocks
+    const pNoLogs = (1 - pLog) ** numBlocks;
+    const costTraverse = 16 * numBlocks + 75 * (1 - pNoLogs);
```
0xOlias, line 355: Is this double negation intentional?

packages/core/src/sync-realtime/service.ts
```diff
@@ -202,444 +219,506 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
-      if (matchedLogCount > 0) {
-        // Filter transactions down to those that are required by the matched logs.
-        const requiredTransactionHashes = new Set(
-          matchedLogs.map((l) => l.transactionHash),
-        );
-        const filteredTransactions =
-          newBlockWithTransactions.transactions.filter((t) =>
-            requiredTransactionHashes.has(t.hash),
+      if (newBlocks[0].parentHash !== this.mostRecentBlock.hash) {
+        await this.reorgTraverse(newBlocks[0], hexToNumber(newBlock.number));
```
0xOlias, line 376: Do you need to return here after this?

packages/core/src/sync-realtime/service.ts
```diff
@@ -202,444 +219,506 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
 
-        // TODO: Maybe rename or at least document behavior
-        await this.syncStore.insertRealtimeBlock({
-          chainId: this.network.chainId,
-          block: newBlockWithTransactions,
-          transactions: filteredTransactions,
-          logs: matchedLogs,
+      if (requestLogs) {
+        // Get logs
+        const logs = await this._eth_getLogs({
```
0xOlias, line 393: This is awesome!

packages/core/src/sync-realtime/service.ts
```diff
@@ -202,444 +219,506 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
-          const block = await this.requestQueue.request({
-            method: "eth_getBlockByNumber",
-            params: [numberToHex(number), true],
-          });
-          if (!block) {
-            throw new Error(`Failed to fetch block number: ${number}`);
-          }
-          return block as BlockWithTransactions;
-        });
+    this.mostRecentBlock = realtimeBlockToLightBlock(newBlock);
```
0xOlias, line 473: Any reason not to call this `latestBlock`?

packages/core/src/sync-realtime/service.ts
```diff
@@ -202,444 +219,506 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
+    newBlock: RealtimeBlock,
+    latestBlockNumber: number,
+  ) => {
+    const commonAncestor = this.blocks.findLast(
+      (parent) => parent.hash === newBlock.parentHash,
+    );
+
+    if (commonAncestor !== undefined) {
+      // Found parent block in local blocks
+      this.emit("shallowReorg", {
```
0xOlias, line 489: I think this should only be emitted after all the store operations are done

packages/core/src/sync-realtime/service.ts
```diff
@@ -202,444 +219,506 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
-      detectedAtBlockNumber: newBlock.number,
-      minimumDepth: depth,
-    });
-
-    this.common.logger.warn({
-      service: "realtime",
-      msg: `Unable to reconcile >${depth}-block reorg (network=${this.network.name})`,
-    });
+  /** Returns true if "blocks" is not missing any blocks from tip to "finalizedBlockNumber". */
+  private isBlocksComplete = (): boolean => {
```
0xOlias, line 717: `isLocalChainConsistent` or something similar? Was not sure what this meant when reading it above.

0xOlias (commented):
This looks great, makes a ton of sense. Mostly nits and questions!

packages/core/src/sync-realtime/service.ts
kyscott18: yes

packages/core/src/sync-realtime/service.ts
kyscott18: Haven't figured it out yet, not sure you do because right above you just fetched a bunch of new blocks and they might still be useful.

packages/core/src/sync-realtime/service.ts
kyscott18: At first it was confusing because there was also a function called "getLatestBlock". But now I think it makes sense.

packages/core/src/sync-realtime/service.ts
kyscott18: Yeah this is a way better name

0xOlias (commented):
Questions and nits

packages/core/src/sync-realtime/service.ts
```diff
@@ -79,19 +92,47 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
+
+    const isAddressDefined = this.logFilterSources.every(
+      (source) => !!source.criteria.address,
+    );
+    this.address =
+      !this.hasFactorySource && isAddressDefined
+        ? (sources.flatMap((source) => source.criteria.address) as Address[])
+        : undefined;
+
+    this.topics = sources.flatMap((source) => {
```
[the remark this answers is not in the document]
kyscott18: I'm worried that requesting a range of blocks with no log filters or addresses is going to be over the maximum data size for some rpc providers

packages/core/src/sync-realtime/service.ts
```diff
@@ -150,496 +190,511 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
-        this.common.logger.info({
-          service: "realtime",
-          msg: `Reconciled ${depth}-block reorg with common ancestor block ${commonAncestorBlock.number} (network=${this.network.name})`,
+          blockNumber: commonAncestor.number,
         });
+      }
+    };
 
+    let i = 0;
+    for (; i < localLogs.length && i < matchedLogs.length; i++) {
```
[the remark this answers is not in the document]
kyscott18: No, its used after the for loops goes out of scope

packages/core/src/sync-realtime/service.ts
```diff
@@ -150,496 +190,511 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
-
-        // Default to a retry (uses the retry options passed to the queue).
-        queue.addTask({ ...task });
-      },
-    });
-
-    return queue;
+      if (this.isProcessBlockQueued) {
+        this.isProcessBlockQueued = false;
+        await this.processBlock();
```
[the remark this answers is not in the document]
kyscott18: There are three cases for the realtime sync process. 

1. Nothing is running.
2. Something is running and after it should wait for the next polling interval.
3. Something is running and before it finished, the polling interval finished. We shouldn't run two at once. The running process should finish and then immediately re-run.

packages/core/src/sync-realtime/service.ts
```diff
@@ -150,496 +190,511 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
+    // Don't attempt to skip "eth_getLogs" if a factory source is present.
+    // Note: this may not be a possible path depending on the implementation of "determineSyncPath".
+    const requestLogs = this.hasFactorySource
+      ? true
+      : newBlocks.some((block) =>
+          isMatchedLogInBloomFilter({
+            bloom: block.logsBloom,
+            logFilters: criteria,
+          }),
+        );
```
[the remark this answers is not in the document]
kyscott18: Just making sure this should actually be:

```ts
const canSkipGetLogs =
  !this.hasFactorySource &&
  newBlocks.every((block) =>
    !isMatchedLogInBloomFilter({
      bloom: block.logsBloom,
      logFilters: criteria,
    }))
```

packages/core/src/sync-realtime/service.ts
```diff
@@ -150,496 +190,511 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
+    const logs = await this._eth_getLogs({
+      fromBlock: numberToHex(latestBlockNumber + 1),
+      toBlock: numberToHex(newBlockNumber),
+    });
 
-    // Store the block objects as we fetch them.
-    // Once we find the common ancestor, we will add these blocks to the queue.
-    const canonicalBlocksWithTransactions = [newBlockWithTransactions];
+    const matchedLogs = await this.getMatchedLogs(
+      logs,
```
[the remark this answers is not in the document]
kyscott18: Doesn't the call to `eth_getLogs` with endBlock being newBlock do that?

```ts
const logs = await this._eth_getLogs({
  fromBlock: numberToHex(latestLocalBlockNumber + 1),
  toBlock: newBlock.number,
});
```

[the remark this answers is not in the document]
0xOlias: Yes, that makes sense

packages/core/src/sync-realtime/service.ts
```diff
@@ -486,160 +374,393 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
+          blockNumber: commonAncestor.number,
         });
+      }
+    };
+
+    let i = 0;
+    for (; i < localLogs.length && i < matchedLogs.length; i++) {
+      const lightMatchedLog = realtimeLogToLightLog(matchedLogs[i]);
+      if (lightMatchedLog.blockHash !== localLogs[i].blockHash) {
+        handleReorg(i);
```
0xOlias, line 584: If our local chain has multiple now-invalid blocks, won't this run multiple times (unnecessarily)?

packages/core/src/sync-realtime/service.ts
```diff
@@ -486,160 +374,393 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
+
+    return { blocks: blocks, logs: matchedLogs };
+  };
+
+  /**
+   * Check if a re-org occurred by comparing remote logs to local.
+   *
+   * @returns True if a re-org has occurred.
+   */
+  detectReorg = async (latestBlockNumber: number) => {
```
0xOlias, line 511: Still makes sense to me that this would be called `reconcileReorg` or something. The current function signature kinda suggests to me that it's a pure function.

0xOlias (approved):

packages/core/src/sync-realtime/service.ts
kyscott18: Doesn't it return right after this line?

PR #621, opened 2024-02-07
Improve Request Queue

kyscott18:


packages/core/src/utils/requestQueue.test.ts
```diff
@@ -119,6 +119,7 @@ test("onIdle() resolves once idle", async ({ networks, common }) => {
   queue.request({ method: "eth_chainId" });
   queue.request({ method: "eth_chainId" });
 
+  queue.clear();
```
0xOlias, line 122: Why did you need to add these?

packages/core/src/utils/requestQueue.ts
```diff
@@ -94,44 +92,31 @@ export const createRequestQueue = ({
...
 
-        network.transport
+        const p = network.transport
           .request(params)
-          .then((a) => {
-            resolve(a);
-          })
+          .then(resolve)
           .catch(reject)
           .finally(async () => {
```
0xOlias, line 99: Does this need to be async?

0xOlias (approved):

packages/core/src/utils/requestQueue.test.ts
kyscott18: More accurately reflects how it is used in the actual codebase

packages/core/src/utils/requestQueue.ts
kyscott18: Nice catch

PR #625, opened 2024-02-07
Detect re-orgs in traversal algorithm in realtime service

kyscott18:


packages/core/src/sync-realtime/service.ts
```diff
@@ -441,6 +448,11 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
     );
     newBlocks.push(newBlock);
 
+    // Detect re-org
+    if (newBlocks[0].parentHash !== latestLocalBlock.hash) {
```
0xOlias, line 452: Do we also need to check that the missing range doesn't have a forked block in it? Eg if `newBlocks` has length > 1?

packages/core/src/sync-realtime/service.ts
```diff
@@ -302,8 +302,10 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
         ? await this.syncTraverse(newBlock)
         : await this.syncBatch(newBlock);
 
-    this.logs.push(...syncedData.logs.map(realtimeLogToLightLog));
-    this.blocks.push(...syncedData.blocks.map(realtimeBlockToLightBlock));
+    if (!syncedData.reorg) {
```
0xOlias, line 305: Maybe `syncedData` should be called `syncResult` ?

packages/core/src/sync-realtime/service.ts
```diff
@@ -509,23 +525,23 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
-    const newFinalizedBlockNumber =
-      latestBlockNumber - this.network.finalityBlockCount;
+    // const newFinalizedBlockNumber =
+    //   latestBlockNumber - this.network.finalityBlockCount;
 
     // Note: toBlock could be latestBlock
     const logs = await this._eth_getLogs({
       fromBlock: numberToHex(this.finalizedBlock.number + 1),
-      toBlock: numberToHex(newFinalizedBlockNumber),
+      toBlock: numberToHex(latestBlockNumber),
```
0xOlias, line 534: So now, reconcileReorg is a larger range?

0xOlias (approved):
LGTM, let's test it live

PR #626, opened 2024-02-09
Fatal Realtime Service Error

kyscott18:


packages/core/src/sync-realtime/service.ts
```diff
@@ -251,15 +252,16 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
-        msg: `Realtime sync task failed (network=${
-          this.network.name
-        }, error=${`${error.name}: ${error.message}`})`,
+        error,
+        msg: `Realtime sync task failed (network=${this.network.name})`,
         network: this.network.name,
       });
-      // Note: Good spot to throw a fatal error.
+
+      this.emit("fatal", error);
```
0xOlias, line 264: Are there any retries before this happens?

packages/core/src/Ponder.ts
```diff
@@ -549,6 +549,16 @@ export class Ponder {
       realtime.on("shallowReorg", (checkpoint) => {
         this.syncGatewayService.handleReorg(checkpoint);
       });
+
+      realtime.on("fatal", async (error) => {
+        error.stack = undefined;
+        this.common.logger.fatal({
+          service: "app",
+          msg: "Failed to fetch initial realtime data",
```
0xOlias, line 557: Fix msg

packages/core/src/sync-realtime/service.ts
kyscott18: Right now just the internal viem transport retries

packages/core/src/sync-realtime/service.ts
```diff
@@ -245,31 +246,35 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
+          }, error=${`${error.name}: ${error.message}`})`,
+          network: this.network.name,
+        });
 
-      if (this.isProcessBlockQueued) {
-        this.isProcessBlockQueued = false;
-        await this.process();
-      } else {
-        this.emit("idle");
+        if (i === 3) this.emit("fatal");
```
0xOlias, line 266: Looks great

0xOlias (approved):

PR #628, opened 2024-02-09
Uses the correct language parser in static analysis

kyscott18:


0xOlias (approved):

0xOlias (commented):
Should probably add a bugfix changeset

PR #644, opened 2024-02-15
Improve create-ponder behavior for unverified contrats

0xOlias:


packages/create-ponder/src/etherscan.ts
```diff
@@ -278,12 +270,36 @@ const getContractAbiAndName = async (
...
+    if (rawAbi === "Contract source code not verified") {
+      warnings.push(
+        `Contract ${contractAddress} is unverified or has an empty ABI.`,
+      );
+      abi = [];
+    } else {
+      abi = JSON.parse(rawAbi);
+    }
+
+    contractName = data.result[0].ContractName;
```
kyscott18, line 291: What if this is undefined

kyscott18 (approved):

packages/create-ponder/src/etherscan.ts
0xOlias: Added fallback. And this is in a block that gets caught so should be OK if it's something weird / an error.

PR #648, opened 2024-02-21
Fixes issue with parsing class methods as helper functions

kyscott18:
```ts
class HelperClass {
  async helper({ context }: { event: Event; context: Context }) {
    await context.db.Table.upsert({
      id: "kyle",
    });
  }
}

ponder.on("C:Event1", async ({ event, context }) => {
  const h = new HelperClass();
  await h.helper({ context, event });
});
```
is now handled by Ponder static analysis.


0xOlias (approved):
LGTM

PR #650, opened 2024-02-22
feat: enable allowImportingTsExtensions in tsconfig

o-az:
> `--allowImportingTsExtensions` allows TypeScript files to import each other with a TypeScript-specific extension like `.ts`, `.mts`, or `.tsx`

[sauce](https://www.typescriptlang.org/tsconfig#allowImportingTsExtensions)

is enabling here sufficient or should I also add the config to each template?

### DIFF
--- a/examples/reference-erc20/tsconfig.json
+++ b/examples/reference-erc20/tsconfig.json
@@ -17,6 +17,7 @@
     "noEmit": true,
     "lib": ["ES2022"],
     "target": "ES2022",
+    "allowImportingTsExtensions": true,
 
     // Skip type checking for node modules
     "skipLibCheck": true


0xOlias:
Thanks for opening!

I'm open to adding this to the examples and templates, but could you provide a bit more context as to why we want this?

o-az:
> Thanks for opening!
> 
> 
> 
> I'm open to adding this to the examples and templates, but could you provide a bit more context as to why we want this?

Working in a codebase where the convention happens to be .ts imports and settings is configured to auto append.ts. When I added ponder as a workspace in monorepo it errored when imports had .ts
So I checked if enabling this has any other side effects and it looks like there are none.

0xOlias:
> > Thanks for opening!
> > I'm open to adding this to the examples and templates, but could you provide a bit more context as to why we want this?
> 
> Working in a codebase where the convention happens to be .ts imports and settings is configured to auto append.ts. When I added ponder as a workspace in monorepo it errored when imports had .ts So I checked if enabling this has any other side effects and it looks like there are none.

Got it. So, you changed this setting in your ponder project tsconfig and it worked? Basically, if that worked for you, I'm not sure it makes sense to update every Ponder tsconfig. It doesn't look like the `create-next` tsconfig includes this flag (which I often use as reference).

0xOlias:
Closing - will revisit if others report!

PR #655, opened 2024-02-23
Improve Static Analysis

kyscott18:
Proactively this time


packages/core/src/indexing/service.ts
```diff
@@ -998,4 +1011,22 @@ export class IndexingService extends Emittery<IndexingEvents> {
...
+      tableName,
+    }: { storeMethod: StoreMethod; tableName: string }) => {
+      const matchedAccess = this.tableAccess?.[indexingFunctionKey]?.find(
+        (t) => t.storeMethod === storeMethod && t.tableName === tableName,
+      );
+
+      if (matchedAccess === undefined) {
+        this.common.logger.warn({
+          service: "indexing",
+          msg: `Unexpected table access in indexing function "${indexingFunctionKey}". Events could potentially be run out of order. Please report this bug to "http://github.com/ponder-sh/ponder/issues".`,
```
0xOlias, line 1028: Would include the unexpected table/method in here too.

`Unexpected table access "${tableName}.${storeMethod}()" in indexing function "${indexingFunctionKey}". This may cause event ordering errors. Please open an issue: https://github.com/ponder-sh/ponder/issues`

0xOlias (approved):

PR #662, opened 2024-02-26
Kjs/gle

kyscott18:


packages/core/src/sync-historical/service.ts
```diff
@@ -137,6 +140,7 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
   private blockTasksEnqueuedCheckpoint = 0;
 
   private queue: Queue<HistoricalSyncTask>;
+  private debouncedEmitCheckpoint: (checkpoint: Checkpoint) => void;
```
0xOlias, line 143: Any reason to define it here rather than as a private method near the bottom?

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -824,82 +814,65 @@ export class PostgresSyncStore implements SyncStore {
...
+          id: string;
+          chainId: number;
+          criteria: FactoryCriteria;
+          fromBlock?: number;
+          toBlock?: number;
+          includeEventSelector: Hex;
+        }[];
+      }
+  )) {
+    let stopClock = startClock();
```
0xOlias, line 847: I think we should pick a timing approach (either startClock or `performance.now()`) and use it everywhere in the file. And, we should use the same approach in the SQLite sync store. Wdyt?

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -1047,16 +1021,39 @@ export class PostgresSyncStore implements SyncStore {
...
         block: Block;
         transaction: Transaction;
       };
     });
 
+    stopClock = startClock();
+
     // Query for the checkpoint of the last event in the requested range (ignore the batch limit)
-    const lastCheckpointRows = await baseQuery
+    const lastCheckpointRows = await this.db
```
0xOlias, line 1034: It makes sense that these two queries don't need to be in a transaction, correct?

Should we consider running them at the same time?

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -824,82 +814,65 @@ export class PostgresSyncStore implements SyncStore {
...
-    const start = performance.now();
+  } & (
+    | {
+        logFilters: {
+          id: string;
+          chainId: number;
+          criteria: LogFilterCriteria;
+          fromBlock?: number;
+          toBlock?: number;
+          includeEventSelector: Hex;
```
0xOlias, line 831: I noticed that we're now always including the event selector in the SQL filter. Maybe we should rename this?

packages/core/src/sync-store/sqlite/store.ts
```diff
@@ -1220,7 +1229,7 @@ export class SqliteSyncStore implements SyncStore {
   }) => {
     const exprs = [];
 
-    exprs.push(eb("source_id", "=", logFilter.id));
+    // exprs.push(eb("source_id", "=", logFilter.id));
```
0xOlias, line 1232: rm

packages/core/src/utils/debounce.test.ts
```diff
@@ -0,0 +1,83 @@
...
+});
+
+test("sets timeout to run after interval", async () => {
+  let i = 0;
+  const increment = (_i: number) => {
+    i = _i;
+  };
+  const d = debounce(1, increment);
+
+  d(1);
```
0xOlias, line 38: Should you be able to `expect(i).toBe(1)` here?

packages/core/src/utils/debounce.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  d();
+
+  d();
+  d();
+  d();
+  d();
+
+  await wait(1);
+
+  expect(i).toBe(2);
```
0xOlias, line 82: Could use `toHaveBeenCalledTimes` here

packages/core/src/utils/debounce.ts
```diff
@@ -1,18 +1,37 @@
...
-export function debounce<T extends Fn>(ms: number, fun: T) {
-  let toId: any;
-  return (...args: Parameters<T>) => {
-    clearTimeout(toId);
-    toId = setTimeout(() => fun(...args), ms);
+export function debounce<param extends unknown[], returnType>(
+  ms: number,
+  fun: (...x: param) => returnType,
+) {
+  let lastTimestamp = 0;
```
0xOlias, line 12: Should this be initialized to `0`?

kyscott18: Depends on whether or not the first call to debounce should have lag or not

packages/core/src/sync-historical/service.ts
kyscott18: Debounce returns a callback which has extra context about the previous invocations of that callback. You can't just call `debounce()` in a helper function, because then it doesn't have any information about when it was called last and what timers are set.

0xOlias (approved):

PR #667, opened 2024-02-27
upgrade ast-grep to fix docker compatibility

### LINKED ISSUE #666: Node Alpine Support
**Issue**
I've been trying to setup Ponder inside of docker using node alpine and it's been crashing on startup with this error `Cannot find module '@ast-grep/napi-linux-x64-musl'` it's an issue with `ast-grep`. Workaround for the time being is to just use an OS that uses glibc over msul libc.

**Fix**
This was fixed in this PR https://github.com/ast-grep/ast-grep/pull/944 and should be available for use in `0.19.2` and then I think alpine would work no problem if Ponder can update to it 🙏

kyscott18:
Closes #666

### DIFF
--- a/.changeset/long-cups-grow.md
+++ b/.changeset/long-cups-grow.md
@@ -1,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
+Fixed Docker compatibility issues by upgrading ast-grep.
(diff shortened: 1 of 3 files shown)


.changeset/long-cups-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
+Fixed Docker compatibility issues by upgrading ast-grep.
```
0xOlias, line 5: Updated `ast-grep` to `0.19.3` which fixed support for running Ponder on Alpine Linux-based Docker images.

PR #679, opened 2024-03-01
fix: sqlite `filename` config value was ignored

jaylmiller:
Screen Shot 2024-03-01 at 11 27 47 AM
Screen Shot 2024-03-01 at 11 27 57 AM

### DIFF
--- a/packages/core/src/config/database.ts
+++ b/packages/core/src/config/database.ts
@@ -28,10 +28,15 @@ export const buildDatabase = ({
   config: Config;
 }): DatabaseConfig => {
   const { ponderDir, rootDir } = common.options;
-  const defaultStorePath = path.join(ponderDir, "store");
-  const defaultSyncFilePath = path.join(defaultStorePath, "sync.db");
-  const defaultIndexingFilePath = path.join(defaultStorePath, "indexing.db");
-  const sqlitePrintPath = path.relative(rootDir, defaultStorePath);
+  const storePath = (() => {
+    if (config.database?.kind === "sqlite" && config.database.filename) {
+      return path.join(rootDir, config.database.filename);
+    }
+    return path.join(ponderDir, "store");
+  })();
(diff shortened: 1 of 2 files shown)


0xOlias:
Nice catch. Taking a look, I realize that we broke this a little while back when we separated the indexing store and sync store into separate database files (before it was just one filename). So now, the `filename` option is kind of a misnomer; there's no way for it to behave as expected.

IMO the fix is to rename `filename` to `directory`, but that's technically a breaking change. So we're going to include it in `0.3` (tagged you in a comment on that PR).

TL;DR you're totally right but I don't think it makes sense to fix in this way. Feel free to push back!

packages/core/src/config/database.ts
```diff
@@ -28,10 +28,15 @@ export const buildDatabase = ({
...
-  const defaultStorePath = path.join(ponderDir, "store");
-  const defaultSyncFilePath = path.join(defaultStorePath, "sync.db");
-  const defaultIndexingFilePath = path.join(defaultStorePath, "indexing.db");
-  const sqlitePrintPath = path.relative(rootDir, defaultStorePath);
+  const storePath = (() => {
+    if (config.database?.kind === "sqlite" && config.database.filename) {
+      return path.join(rootDir, config.database.filename);
+    }
+    return path.join(ponderDir, "store");
+  })();
```
0xOlias, line 36: IMO a ternary here is probably more idiomatic for TS:
```ts
const storePath = config.database?.kind === "sqlite" && config.database.filename
  ? path.join(rootDir, config.database.filename);
  : path.join(ponderDir, "store");
```

0xOlias (commented):
Looks good! Only thing missing is a changeset.

packages/core/src/config/database.ts
jaylmiller: good point

0xOlias (approved):
LGTM, gz on first contribution!

PR #685, opened 2024-03-02
Fixes race condition in debounce

kyscott18:


packages/core/src/utils/debounce.test.ts
```diff
@@ -1,83 +1,87 @@
...
-});
+  await wait(20);
 
-test("sets timeout to run after interval", async () => {
-  let i = 0;
-  const increment = (_i: number) => {
-    i = _i;
-  };
-  const d = debounce(1, increment);
+  expect(fun).toHaveBeenCalledTimes(2);
```
0xOlias, line 25: Good test

0xOlias (approved):

PR #688, opened 2024-03-02
Utilities package

kyscott18:
primarily to flex

### DIFF
--- a/packages/utils/.gitignore
+++ b/packages/utils/.gitignore
@@ -1,0 +1,4 @@
+dist/
+.ponder/
+generated/
+.env.local
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -1,0 +1,20 @@
+{
+  "name": "@ponder/utils",
+  "version": "0.0.0",
+  "description": "",
+  "license": "MIT",
+  "type": "module",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ponder-sh/ponder",
+    "directory": "packages/utils"
+  },
+  "scripts": {
+    "build": "tsup",
+    "typecheck": "tsc --noEmit"
+  },
+  "dependencies": {
+    "tsup": "^8.0.1",
+    "vitest": "^1.0.2"
+  }
+}
--- a/packages/utils/README.md
+++ b/packages/utils/README.md
@@ -1,0 +1,1 @@
+# @ponder/utils
--- a/packages/utils/src/index.ts
+++ b/packages/utils/src/index.ts
@@ -1,0 +1,4 @@
+export {
+  promiseWithResolvers,
+  type PromiseWithResolvers,
+} from "./promiseWithResolvers.js";
--- a/packages/utils/src/promiseWithResolvers.test.ts
+++ b/packages/utils/src/promiseWithResolvers.test.ts
@@ -1,0 +1,36 @@
+import { assertType, expect, test } from "vitest";
+import { promiseWithResolvers } from "./promiseWithResolvers.js";
+
+test("resolves", async () => {
+  const { promise, resolve } = promiseWithResolvers<number>();
+
+  resolve(1);
+
+  const value = await promise;
+
+  expect(value).toBe(1);
+});
+
+test("rejects", async () => {
+  let rejected = false;
+
+  const { promise, reject } = promiseWithResolvers();
+
+  promise.catch(() => {
+    rejected = true;
+  });
+
+  await Promise.reject().catch(reject);
+
+  expect(rejected).toBe(true);
+});
+
+test("resolve type", () => {
+  const { resolve } = promiseWithResolvers<number>();
+  assertType<(arg: number) => void>(resolve);
+});
+
+test("promise type", () => {
+  const { promise } = promiseWithResolvers<number>();
+  assertType<Promise<number>>(promise);
+});
--- a/packages/utils/src/promiseWithResolvers.ts
+++ b/packages/utils/src/promiseWithResolvers.ts
@@ -1,0 +1,20 @@
+export type PromiseWithResolvers<TPromise> = {
+  resolve: (arg: TPromise) => void;
+  reject: (error: Error) => void;
+  promise: Promise<TPromise>;
+};
+
+/**
+ * @description Application level polyfill.
+ */
+export const promiseWithResolvers = <
+  TPromise,
+>(): PromiseWithResolvers<TPromise> => {
+  let resolve: (arg: TPromise) => void;
+  let reject: (error: Error) => void;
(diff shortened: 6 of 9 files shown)


packages/utils/.gitignore
```diff
@@ -0,0 +1,4 @@
+dist/
+.ponder/
+generated/
```
0xOlias, line 3: Don't need?

packages/utils/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@ponder/utils",
+  "version": "0.0.0",
```
0xOlias, line 3: We should think about how we want to version this package. My instinct is that we should lock it with the other packages, but could be convinced otherwise.

0xOlias (approved):

packages/common/src/debounce.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Creates a debounced function that waits "ms" milliseconds before being called.
```
0xOlias, line 2: Creates a debounced function that waits `ms` milliseconds between invocations. If the function is called multiple times between invocations, the latest arguments passed to the function will be used.

packages/common/src/debounce.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+export function debounce<param extends unknown[], returnType>(
+  ms: number,
+  fun: (...x: param) => returnType,
+) {
+  let args: param;
```
0xOlias, line 8: This type name is a little weird

packages/common/src/dedupe.ts
```diff
@@ -0,0 +1,13 @@
+export const dedupe = <item, id>(
```
0xOlias, line 1: Maybe use a function expression to be consistent? Also a docstring would be helpful considering there's some complexity around behavior of objects vs primitives and the `getId` option.

packages/common/src/queue.test.ts
```diff
@@ -0,0 +1,349 @@
...
+});
+
+test("onIdle short loop", async () => {
+  const queue = createQueue({
+    concurrency: 1,
+    frequency: 1,
+    worker: () => Promise.resolve(),
+  });
+
+  await queue.onIdle();
```
0xOlias, line 166: No `expect` here? Also a few other tests in this file

packages/common/src/queue.test.ts
```diff
@@ -0,0 +1,349 @@
...
+  expect(queue.size()).toBe(2);
+  expect(func).toHaveBeenCalledTimes(2);
+
+  await new Promise((resolve) => setTimeout(resolve, 1_010));
+
+  expect(queue.size()).toBe(0);
+  expect(func).toHaveBeenCalledTimes(4);
+});
+
+test("event loop", async () => {
```
0xOlias, line 304: Not sure what this is testing - could you write a bit more in the test description?

packages/common/src/queue.ts
```diff
@@ -0,0 +1,167 @@
...
+  worker,
+}: {
+  worker: (task: taskType) => Promise<returnType>;
+} & (
+  | {
+      concurrency: number;
+      frequency: number;
+    }
+  | { concurrency: number; frequency?: never }
+  | { concurrency?: never; frequency: number }
```
0xOlias, line 36: Maybe a runtime validation for this type?

packages/utils/src/parseGetLogsError/parseGetLogsError.ts
```diff
@@ -0,0 +1,213 @@
...
+  InvalidParamsRpcError,
+  LimitExceededRpcError,
+  type LogTopic,
+  RpcError,
+  RpcRequestError,
+  hexToBigInt,
+  numberToHex,
+} from "viem";
+
+export type ParseGetLogsErrorParameters = {
```
0xOlias, line 14: I think we should rename this `getLogsRetryHelper`.

packages/core/src/sync-historical/service.ts
```diff
@@ -832,7 +833,7 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
     });
 
     if (newBlockCheckpoint) {
-      this.debouncedEmitCheckpoint({
+      this.debouncedEmitCheckpoint.callback({
```
0xOlias, line 836: `callback` looks a bit weird here, maybe `invoke` or `call`?

packages/core/src/sync-historical/service.ts
```diff
@@ -952,37 +953,40 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
     fromBlock: Hex;
     toBlock: Hex;
   }): Promise<RpcLog[]> => {
+    const _params: ParseGetLogsErrorParameters["params"] = [
```
0xOlias, line 956: I feel like there's probably a Viem type that's better here? Maybe not though.

packages/core/src/sync-historical/service.ts
```diff
@@ -952,37 +953,40 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
...
-        err as LogFilterError,
-        params.fromBlock,
-        params.toBlock,
-      );
+      const getLogsErrorResponse = parseGetLogsError({
+        params: _params,
+        error: err as RpcError,
+      });
+
+      if (!getLogsErrorResponse.shouldRetry) throw err;
```
0xOlias, line 981: Should we add a debug-level log here? It might be useful.

packages/core/src/sync-realtime/service.test.ts
```diff
@@ -112,7 +112,7 @@ test("start() sync realtime data with batch method", async (context) => {
   await service.setup();
   service.start();
 
-  await testClient.mine({ blocks: 8 });
+  await testClient.mine({ blocks: 10 });
```
0xOlias, line 115: Why did you have to make changes in this file?

packages/common/src/queue.test.ts
kyscott18: These tests are testing that the promise resolves. If the test were to fail, it would basically just timeout

packages/core/src/sync-realtime/service.test.ts
kyscott18: Because we changed the behavior of poll to run the "process" function immediately instead of wait the interval to start processing.

packages/core/src/sync-historical/service.ts
kyscott18: Nice idea

packages/core/src/sync-historical/service.ts
kyscott18: Couldn't find one

PR #692, opened 2024-03-06
New viem transports

kyscott18:


packages/common/src/queue.ts
```diff
@@ -182,5 +236,9 @@ export const createQueue = <returnType, taskType = void>({
       }
       return emptyPromiseWithResolvers.promise;
     },
+    setParameters: (_parameters) => {
+      validateParameters(_parameters);
+      parameters = _parameters;
+    },
```
0xOlias, line 242: If you initialize with both parameters defined, then call `setParameters` with only one defined, I think it will overwrite the other one here (as `undefined`). Just want to double check that's the expected behavior.

packages/core/src/index.ts
```diff
@@ -2,4 +2,9 @@ export { createConfig } from "@/config/config.js";
 export { createSchema } from "@/schema/schema.js";
 export type { Block, Log, Transaction } from "@/types/eth.js";
 export type { Virtual } from "@/types/virtual.js";
-export { type MergeAbis, mergeAbis } from "@ponder/utils";
+export {
+  type MergeAbis,
+  mergeAbis,
+  loadBalancedTransport,
+  rateLimitedTransport,
```
0xOlias, line 9: Viem transport naming is pretty terse `http`, `fallback`. Maybe we should follow suit with `loadBalance`, `rateLimit`.

packages/utils/src/loadBalancedTransport.ts
```diff
@@ -0,0 +1,30 @@
+import { type Transport, type TransportConfig, createTransport } from "viem";
+
+/**
+ * @description Creates a load balanced transport that spreads requests between child transports using a round robin algorithm.
```
0xOlias, line 4: Nice description

packages/utils/src/rateLimitedTransport.ts
```diff
@@ -0,0 +1,43 @@
...
+    browser,
+  }: { requestsPerSecond: number; browser?: boolean },
+): Transport => {
+  return ({ chain, retryCount, timeout }) => {
+    const transport =
+      chain === undefined
+        ? _transport({ retryCount: 0, timeout })
+        : _transport({ chain, retryCount: 0, timeout });
+
+    const queue = createQueue({
```
0xOlias, line 20: I wonder if there are any memory leak issues to consider here.

packages/utils/src/rateLimitedTransport.ts
```diff
@@ -0,0 +1,43 @@
...
+      worker: (body: {
+        method: string;
+        params?: unknown;
+      }) => {
+        return transport.request(body);
+      },
+    });
+
+    return createTransport({
+      key: "rate",
```
0xOlias, line 34: `"rateLimit"` assuming we go with that name

0xOlias (approved):
Looks great, excited for this

packages/common/src/queue.ts
kyscott18: Nice call - will make it so that it only updates named keys

PR #693, opened 2024-03-07
fix: edge case when computing last checkpoint when result set is empty

jaylmiller:
A regression was introduced when `getLogEvents` query was split into 2 separate queries (the 2nd one being just to get the `lastCheckpoint`). The 2nd one did not include the lower checkpoint bounds and just queried for the last event in the range sorted descending. This worked fine, except in an edge case where no rows are returned in the main result set. Without the lower bounds on the secondary query, you end up using an event from outside the main result set to define the `lastCheckpoint` value that gets returned


.changeset/good-walls-film.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
+Fix regression in sync-store getLogEvents `lastCheckpoint` that was introduced after splitting out the query into 2.
```
0xOlias, line 5: You and I know what this means, but it's IMO a bit too low-level for the public changelog. It's a hard changeset to write because we're not certain that it caused any bugs. How about:
```
Fixed a regression that may have caused bugs related to indexing progress or event ordering.
```

0xOlias (changes requested):
LGTM, awesome find and thanks for opening. Suggested a different changeset message.

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -961,6 +961,7 @@ export class PostgresSyncStore implements SyncStore {
           "blocks.number as block_number",
           "logs.logIndex as log_logIndex",
         ])
+        .where((eb) => this.buildCheckpointCmprs(eb, ">", fromCheckpoint))
```
jaylmiller, line 964: update: i still need to dig into why adding this lower bound breaks this query. by break i mean it makes it return zero rows. 

the corresponding one in sqlite works as expected

jaylmiller: by works as expected i mean it returns a single row containing the last element in the filter/checkpoint combinations result set

PR #709, opened 2024-03-14
Make `blocks.sha3uncles` column nullable

d-mooers:
Makes sha3uncles column nullable, enabling support for Celo.


kyscott18 (approved):
Looks great, much appreciated

PR #712, opened 2024-03-14
Adds retry delay with exponential backoff

kyscott18:
Also fixes logs for block not found and fixed a potential bug with reorg reconciliation

### DIFF
--- a/.changeset/stupid-rockets-provide.md
+++ b/.changeset/stupid-rockets-provide.md
@@ -1,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
+Added retry delay with exponential backoff to the realtime service.
--- a/.changeset/stupid-rockets-provide.md
+++ b/.changeset/stupid-rockets-provide.md
@@ -1,5 +1,0 @@
----
-"@ponder/core": patch
----
-
-Added retry delay with exponential backoff to the realtime service.
--- a/packages/core/src/sync-realtime/service.ts
+++ b/packages/core/src/sync-realtime/service.ts
@@ -12,6 +12,7 @@ import { type Checkpoint } from "@/utils/checkpoint.js";
 import { Emittery } from "@/utils/emittery.js";
 import { range } from "@/utils/range.js";
 import type { RequestQueue } from "@/utils/requestQueue.js";
+import { wait } from "@/utils/wait.js";
 import { dedupe, poll } from "@ponder/common";
 import {
   type GetLogsRetryHelperParameters,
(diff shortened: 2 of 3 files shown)


packages/core/src/sync-realtime/service.ts
```diff
@@ -704,13 +709,17 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
         lightMatchedLog.blockHash !== localLogs[i].blockHash ||
         lightMatchedLog.logIndex !== localLogs[i].logIndex
       ) {
-        await handleReorg(i);
+        await handleReorg(localLogs[i].blockNumber - 1);
```
0xOlias, line 712: Maybe add a debug/trace level log here so we can confirm that we're hitting this branch, and the updated branch below?

packages/core/src/sync-realtime/service.ts
```diff
@@ -773,9 +782,19 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
         method: "eth_getBlockByNumber",
         params: [typeof block === "number" ? numberToHex(block) : block, true],
       })
-      .then((block) => {
-        if (!block) throw new BlockNotFoundError({});
-        return block as RealtimeBlock;
+      .then((_block) => {
+        if (!_block)
+          throw new BlockNotFoundError(
```
0xOlias, line 787: We could probably just pass `blockNumber: "latest"` here because it's only user in a string interpolation?
```ts
import type { Hash } from '../types/misc.js'

import { BaseError } from './base.js'

export type BlockNotFoundErrorType = BlockNotFoundError & {
  name: 'BlockNotFoundError'
}
export class BlockNotFoundError extends BaseError {
  override name = 'BlockNotFoundError'
  constructor({
    blockHash,
    blockNumber,
  }: {
    blockHash?: Hash
    blockNumber?: bigint
  }) {
    let identifier = 'Block'
    if (blockHash) identifier = `Block at hash "${blockHash}"`
    if (blockNumber) identifier = `Block at number "${blockNumber}"`
    super(`${identifier} could not be found.`)
  }
}
```

0xOlias (approved):

0xOlias (approved):

.changeset/stupid-rockets-provide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
+Added retry delay with exponential backoff to the realtime service.
```
0xOlias, line 5: I actually don't think we need a changeset here, because the retry behavior is not yet documented / part of the public API.

PR #716, opened 2024-03-17
skip update and upsert queries if rows are equal

kyscott18:


packages/core/src/indexing-store/utils/isRowEqual.ts
```diff
@@ -0,0 +1,31 @@
...
+export const isRowEqual = ({
+  originalRow,
+  updateRow,
+}: { originalRow: EncodedRow; updateRow: EncodedRow }) => {
+  for (const columnName of Object.keys(updateRow)) {
+    if (Buffer.isBuffer(updateRow[columnName])) {
+      if (
+        (updateRow[columnName] as Buffer).equals(
+          originalRow[columnName] as Buffer,
+        ) === false
```
0xOlias, line 20: Should probably handle/test the case where the updated field is a Buffer but the original field is not.

packages/core/src/indexing-store/sqlite/store.ts
```diff
@@ -556,6 +562,11 @@ export class SqliteIndexingStore implements IndexingStore {
                 `Cannot update ${tableName} record with ID ${encodedId} at checkpoint ${encodedCheckpoint} because there is a newer version of the record at checkpoint ${latestRow.effective_from}. Hint: Did you forget to await the promise returned by a store method?`,
               );
 
+            // Skip queries if no changes to the original row were made.
+            if (isRowEqual({ originalRow: latestRow, updateRow: updateRow })) {
```
0xOlias, line 566: Could object shorthand here 😉

0xOlias:
Closed in favor of #774

PR #723, opened 2024-03-18
check nullishness of tx.v before decoding

kyscott18:


0xOlias (approved):

0xOlias (commented):
Could add a changeset: "Relaxed not null constraint for SQLite, which fixes zkSync"

PR #730, opened 2024-03-19
Realtime testing

kyscott18:


packages/core/src/sync-realtime/service.ts
```diff
@@ -748,21 +723,23 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
         detectedAtBlockNumber: latestBlockNumber,
-        minimumDepth: latestBlockNumber - this.blocks[0].number,
       });
 
       this.common.logger.warn({
         service: "realtime",
-        msg: `Unable to reconcile >${
-          latestBlockNumber - this.blocks[0].number
-        }-block reorg (network=${this.network.name})`,
+        msg: `Detected >${this.network.finalityBlockCount} block reorg (network=${this.network.name})`,
```
0xOlias, line 742: I think we should move this to `logger.error`, and message

`
Detected unrecoverable reorg at block ${this.finalizedBlock.number} with local hash ${this.finalizedBlock.hash} and remote hash ${remoteFinalizedBlock.hash} (network=${this.network.name})
`

packages/core/src/sync-realtime/service.ts
```diff
@@ -748,21 +723,23 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
    * Check if deep re-org occured by comparing remote "finalized" block to local.
    */
   private reconcileDeepReorg = async (latestBlockNumber: number) => {
+    this.common.logger.trace({
+      service: "realtime",
+      msg: `Checking if reorg has occurred before or at block ${this.finalizedBlock.number} (network=${this.network.name})`,
+    });
```
0xOlias, line 729: I think we should remove this log, and add a new once in the else block below:

```ts
    this.common.logger.trace({
      service: "realtime",
      msg: `Confirmed local hash matches remote hash at finalized block number ${this.finalizedBlock.number} (network=${this.network.name})`,
    });
```

packages/core/src/sync-realtime/service.ts
```diff
@@ -644,68 +647,45 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
+        chainId: this.network.chainId,
+        fromBlock: BigInt(commonAncestor.number),
+      });
 
-        this.emit("shallowReorg", {
-          blockTimestamp: commonAncestor.timestamp,
-          chainId: this.network.chainId,
-          blockNumber: commonAncestor.number,
-        });
+      this.emit("shallowReorg", {
```
0xOlias, line 679: I think we can rename this event to `"reorg"`

packages/core/src/sync-realtime/service.ts
```diff
@@ -748,21 +723,23 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
    * Check if deep re-org occured by comparing remote "finalized" block to local.
    */
   private reconcileDeepReorg = async (latestBlockNumber: number) => {
```
0xOlias, line 725: Maybe rename this function "validateFinalizedBlock"? And can prob remove the argument entirely if we get rid of the deepReorg event.

packages/core/src/sync-realtime/service.ts
```diff
@@ -619,6 +619,9 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
    * @returns True if a re-org has occurred.
    */
   reconcileReorg = async (latestBlockNumber: number) => {
+    const hasDeepReorg = await this.reconcileDeepReorg(latestBlockNumber);
+    if (hasDeepReorg) return true;
+
     this.common.logger.debug({
       service: "realtime",
       msg: `Checking logs for inconsistencies for blocks from ${
```
0xOlias, line 627: Maybe `Validating local chain from block ${this.finalizedBlock.number + 1} to ${latestBlockNumber} (network=${this.network.name})`

0xOlias (commented):
LGTM, kill the flakes. Mostly naming suggestions.

PR #733, opened 2024-03-20
Simplify realtime sync algorithm

kyscott18:
Realtime sync traverses chain block by block

### DIFF
--- a/packages/core/src/sync-realtime/service.ts
+++ b/packages/core/src/sync-realtime/service.ts
@@ -724,7 +724,10 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
    * Check if deep re-org occured by comparing remote "finalized" block to local.
    */
   private reconcileDeepReorg = async (latestBlockNumber: number) => {
-    // this.common.logger.trace, TODO(kyle)
+    this.common.logger.trace({
+      service: "realtime",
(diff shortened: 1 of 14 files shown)


packages/core/src/sync-realtime/service.ts
```diff
@@ -548,106 +516,119 @@ export class RealtimeSyncService extends Emittery<RealtimeSyncEvents> {
...
 
-    const matchedLogs = await this.getMatchedLogs(
-      logs,
-      BigInt(newBlockNumber),
-      true,
+        return;
+      }
+    }
+
+    throw new Error(
```
0xOlias, line 563: Why is this an invariant violation?

0xOlias (commented):
Looking good

packages/core/src/sync-realtime/service.ts
kyscott18: The code should never get to this point because:

- we have a complete local chain
- we first check for deep reorg
- if not deep reorg we know the local chain is at least partly in tact
- we find the common ancestor by stepping through the chain backwards, re-requesting every block and seeing if the remote block fits into the local chain for that block number
- getting to this line means the local chain is in tact but we weren't able to find a common ancestor in the local chain, which is an invariant violation

0xOlias (approved):

PR #734, opened 2024-03-20
feat: checkpoint codec and comparison includes eventType and transactionIndex

jaylmiller:


packages/core/src/utils/checkpoint.ts
```diff
@@ -102,19 +142,12 @@ export const isCheckpointGreaterThan = (a: Checkpoint, b: Checkpoint) => {
...
-  }
-  // If both are undefined, the checkpoints are equal, so a is not greater than b.
-  if (a.logIndex === undefined && b.logIndex === undefined) {
-    return false;
-  }
-  // If only a is undefined, it's considered greater.
-  if (a.logIndex === undefined) return true;
-  // Otherwise b is undefined and a is defined, so b is greater.
-  else return false;
+  if (a.transactionIndex !== b.transactionIndex)
```
0xOlias, line 145: I wonder if it's easier to encode both checkpoints and then compare the strings lexicograpically, OR encode, convert to bigint, and compare.

packages/core/src/utils/checkpoint.test.ts
```diff
@@ -3,16 +3,76 @@ import { expect, test } from "vitest";
...
+    blockNumber: 1,
+    eventType: 1,
+    transactionIndex: 1,
+    eventIndex: 1,
+  };
+
+  expect(decodedCheckpoint).toMatchObject(expectedCheckpoint);
+});
+
+test("decodeCheckpoint decodes an encoded maxCheckpoint", () => {
```
0xOlias, line 61: Nice

packages/core/src/utils/checkpoint.ts
```diff
@@ -2,9 +2,9 @@ export type Checkpoint = {
   blockTimestamp: number;
   chainId: number;
   blockNumber: number;
-  // Execution index of the log within the block.
-  // If undefined, the checkpoint includes all events in the block.
-  logIndex?: number | undefined;
+  transactionIndex: number;
+  eventType: number;
+  eventIndex: number;
```
0xOlias, line 7: We should standardize naming for `eventIndex` - looks like there are some leftover `EXECUTION_INDEX` names in this file.

packages/core/src/utils/checkpoint.test.ts
```diff
@@ -23,112 +83,155 @@ test("isCheckpointEqual returns false if checkpoints are different", () => {
...
-test("isCheckpointGreaterThan compares correctly on chainId", () => {
+test("isCheckpointGreaterThan compares correctly on transactionIndex", () => {
   const checkpointOne = {
     blockTimestamp: 1,
     chainId: 1,
     blockNumber: 1,
-    logIndex: 1,
+    eventType: 1,
+    transactionIndex: 5,
+    eventIndex: 1,
```
0xOlias, line 146: Super nit, but I'd probably order the properties of this object the same as the encoding order

0xOlias (approved):
LGTM! All nits

PR #736, opened 2024-03-20
feat: implement database migrations in the sync-store

jaylmiller:


packages/core/src/sync-store/postgres/migrations.ts
```diff
@@ -455,6 +456,73 @@ const migrations: Record<string, Migration> = {
...
+              lpad(checkpoint_vals."transactionIndex"::text, 16, '0') ||
+              '5' ||
+              lpad(checkpoint_vals."logIndex"::text, 16, '0'))::numeric(75, 0)
+          FROM checkpoint_vals
+          WHERE ponder_sync.logs.id = checkpoint_vals.id;
+        `.compile(db),
+      );
+
+      // sanity check our checkpoint encoding on the first 10 rows of the table
+      const checkRes = await db.executeQuery(
```
0xOlias, line 486: I might add the type here, eg
```ts
const checkRows = db.executeQuery<{
  timestamp: number;
  chainId: number;
  number: number;
  transactionIndex: number;
  logIndex: number;
  checkpoint: bigint;
}>( ...
```

And then below, you should be able to pass these inline:
```ts
        const expected = encodeCheckpoint({
          blockTimestamp: row.timestamp,
          chainId: row.chainId,
          blockNumber: row.number,
          // ... etc
        });
```

packages/core/src/sync-store/postgres/migrations.ts
```diff
@@ -455,6 +456,73 @@ const migrations: Record<string, Migration> = {
         .execute();
     },
   },
+  "2024_03_20_0_checkpoint_in_logs_table": {
+    async up(db: Kysely<any>) {
+      await db.schema
+        .alterTable("logs")
+        .addColumn("checkpoint", "numeric(75, 0)")
+        .execute();
+      await db.executeQuery(
```
0xOlias, line 465: Our business logic assumes that for every row in `logs`, there will be a corresponding row in `blocks` (by hash). However, this is not enforced at the database level. So, what would happen here if there were indeed logs without corresponding blocks? This ties into our chat re factory logs.

0xOlias (approved):
LGTM, one nit. We might have to tweak this later per the comment about factory logs.

packages/core/src/sync-store/postgres/migrations.ts
jaylmiller: The checkpoint would be null for those. Since we're updating from a join against blocks, any logs without blocks would not get updated in this migration.

0xOlias (approved):
LGTM!

PR #739, opened 2024-03-21
Add effective from index on instance tables

kyscott18:


0xOlias (commented):
Looks good in theory - might be worth calling out the specific query that we're trying to speed up here in the PR comment. Also, could write a tiny benchmark for "flush" with 10k+ records to measure a difference.

PR #741, opened 2024-03-21
feat: update sync-store to use new checkpoint encoding

jaylmiller:


packages/core/src/utils/checkpoint.ts
```diff
@@ -15,31 +15,65 @@ const CHAIN_ID_DIGITS = 16;
...
 const BLOCK_NUMBER_DIGITS = 16;
 // Same logic as chain ID.
-const EXECUTION_INDEX_DIGITS = 16;
+const TRANSACTION_INDEX_DIGITS = 16;
+// Currently only event type
+const EVENT_TYPE_DIGITS = 1;
+// Same logic as chain ID.
+const EVENT_INDEX_DIGITS = 16;
+
+export const eventTypes = {
```
0xOlias, line 24: Would consider using screaming snake case here (https://google.github.io/styleguide/tsguide.html#identifiers-constants)

packages/core/src/utils/checkpoint.ts
```diff
@@ -15,31 +15,65 @@ const CHAIN_ID_DIGITS = 16;
...
+    eventType.toString() +
+    eventIndex.toString().padStart(EVENT_INDEX_DIGITS, "0");
 
   if (
     result.length !==
     BLOCK_TIMESTAMP_DIGITS +
       CHAIN_ID_DIGITS +
       BLOCK_NUMBER_DIGITS +
-      EXECUTION_INDEX_DIGITS
+      TRANSACTION_INDEX_DIGITS +
```
0xOlias, line 63: Maybe create a constant above for `CHECKPOINT_LENGTH`?

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -948,13 +983,11 @@ export class PostgresSyncStore implements SyncStore {
...
           ])
-          .where((eb) => this.buildCheckpointCmprs(eb, ">", fromCheckpoint))
-          .where((eb) => this.buildCheckpointCmprs(eb, "<=", toCheckpoint))
-          .orderBy("blocks.timestamp", "desc")
-          .orderBy("logs.chainId", "desc")
-          .orderBy("blocks.number", "desc")
-          .orderBy("logs.logIndex", "desc")
+          .where("logs.checkpoint", ">", encodeCheckpoint(fromCheckpoint))
+          .where("logs.checkpoint", "<=", encodeCheckpoint(toCheckpoint))
+          .orderBy("logs.checkpoint", "asc")
```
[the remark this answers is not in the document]
jaylmiller: Oooh. Good catch there

packages/core/src/utils/checkpoint.ts
```diff
@@ -15,31 +15,65 @@ const CHAIN_ID_DIGITS = 16;
...
+  EVENT_TYPE_DIGITS +
+  EVENT_INDEX_DIGITS;
+
+export const EVENT_TYPES = {
+  logs: 5,
+} as const;
 
 export const encodeCheckpoint = (checkpoint: Checkpoint) => {
-  const { blockTimestamp, chainId, blockNumber, logIndex } = checkpoint;
+  runtimeCheckKey(checkpoint, "blockTimestamp");
```
kyscott18, line 37: Why do we need this?

packages/core/src/utils/checkpoint.ts
```diff
@@ -55,77 +89,68 @@ export const decodeCheckpoint = (checkpoint: string): Checkpoint => {
...
-  if (a.logIndex === undefined && b.logIndex === undefined) {
-    return false;
-  }
-  // If only a is undefined, it's considered greater.
-  if (a.logIndex === undefined) return true;
-  // Otherwise b is undefined and a is defined, so b is greater.
-  else return false;
-};
+export const isCheckpointGreaterThan = (a: Checkpoint, b: Checkpoint) =>
+  encodeCheckpoint(a) > encodeCheckpoint(b);
```
kyscott18, line 145: Any idea the difference in performance of this change? This function could be called millions of times.

packages/core/src/utils/checkpoint.test.ts
```diff
@@ -3,16 +3,76 @@ import { expect, test } from "vitest";
...
+    blockNumber: 1,
+    transactionIndex: 1,
+    eventType: 1,
+    eventIndex: 1,
+  };
+
+  expect(decodedCheckpoint).toMatchObject(expectedCheckpoint);
+});
+
+test("decodeCheckpoint decodes an encoded maxCheckpoint", () => {
```
kyscott18, line 61: Great test

packages/core/src/sync-store/postgres/encoding.ts
```diff
@@ -134,6 +134,7 @@ type LogsTable = {
   topic3: Hex | null;
 
   chainId: number;
+  checkpoint?: string;
```
kyscott18, line 137: Why optional?

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -499,6 +509,26 @@ export class PostgresSyncStore implements SyncStore {
...
+      throw new Error("Number is missing from RPC block");
+    }
+    if (log.transactionIndex === null) {
+      throw new Error("Transaction index is missing from RPC log");
+    }
+    if (log.logIndex === null) {
+      throw new Error("Log index is missing from RPC log");
+    }
+    return encodeCheckpoint({
+      blockTimestamp: Number(BigInt(block.timestamp)),
```
kyscott18, line 523: `hexToNumber(block.timestamp)`

packages/core/src/sync-store/postgres/encoding.ts
jaylmiller: the checkpoint column is nullable b/c we're overloading this table by also using it to track the factory logs which don't have checkpoints yet. the idea is to move those to a separate table (or possibly something else) and then we can make checkpoint non-null

packages/core/src/utils/checkpoint.ts
jaylmiller: Prob a bit slower since its doing some allocation on the heap for the encoding bit. Makes the code much more readable and less error prone but fine to change it back to comparing every field manually if that's better. I don't think it'll make any noticeable diff in runtime perf but fine to change it back and if we'd prefer the manual one

PR #749, opened 2024-03-22
feat: update indexing store with new checkpoint encoding

jaylmiller:


packages/core/src/database/postgres/service.test.ts
```diff
@@ -63,15 +63,19 @@ describe.skipIf(shouldSkip)("postgres database", () => {
...
-      "kysely_migration",
-      "kysely_migration_lock",
-      "function_metadata",
-      "table_metadata",
-      "instance_metadata",
-      "0xPet",
-      "0xPerson",
-    ]);
+    expect(
+      new Set(await getTableNames(database.db, "ponder_cache")),
```
0xOlias, line 67: Could also do `.sort()` on both arrays here. Unfortunately, don't think vitest has a matcher for arrays ignoring order.

packages/core/src/database/postgres/service.ts
```diff
@@ -244,9 +247,9 @@ export class PostgresDatabaseService implements BaseDatabaseService {
         functionId: m.function_id,
         functionName: m.function_name,
         fromCheckpoint: m.from_checkpoint
-          ? decodeCheckpoint(m.from_checkpoint)
+          ? decodeCheckpoint(m.from_checkpoint.toString())
           : null,
-        toCheckpoint: decodeCheckpoint(m.to_checkpoint),
+        toCheckpoint: decodeCheckpoint(m.to_checkpoint.toString()),
```
0xOlias, line 252: Are these still necessary now that we're using VARCHAR?

packages/core/src/database/postgres/service.ts
```diff
@@ -377,34 +380,35 @@ export class PostgresDatabaseService implements BaseDatabaseService {
...
+                  tx,
+                ),
               );
 
               // Truncate cache tables to match metadata.
               await tx
                 .withSchema(CACHE_SCHEMA_NAME)
                 .updateTable(tableId)
-                .set({ effective_to: "latest" })
+                .set({ effective_to: encodeCheckpoint(maxCheckpoint) })
```
0xOlias, line 392: Could use the `LATEST` const from above here

packages/core/src/indexing-store/postgres/store.ts
```diff
@@ -23,6 +27,7 @@ import {
 const MAX_BATCH_SIZE = 1_000 as const;
 const DEFAULT_LIMIT = 50 as const;
 const MAX_LIMIT = 1_000 as const;
+const LATEST = encodeCheckpoint(maxCheckpoint);
```
0xOlias, line 30: Might just make sense to export `LATEST` from `utils/checkpoint.js` seeing how much it's used

packages/core/src/database/postgres/service.ts
jaylmiller: good call forgot to remove these!

packages/core/src/sync-store/store.ts
```diff
@@ -181,6 +181,7 @@ export interface SyncStore {
         log: Log;
         block: Block;
         transaction: Transaction;
+        checkpoint?: Checkpoint;
```
kyscott18, line 184: When would checkpoint be undefined?

packages/core/src/database/postgres/migrations.ts
```diff
@@ -35,6 +35,24 @@ const migrations: Record<string, Migration> = {
         .execute();
     },
   },
+  "2024_03_22_new_checkpoint_encoding": {
+    async up(db: Kysely<any>) {
+      await db.schema
+        .alterTable("function_metadata")
+        .alterColumn("from_checkpoint", (cb) => cb.setDataType("varchar(75)"))
```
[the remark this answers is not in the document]
jaylmiller: if it's numeric and the checkpoint has leading zeros they will be stripped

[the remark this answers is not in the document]
jaylmiller: theres no way to recover the actual checkpoint after leading zeros are stripped

packages/core/src/bin/utils/run.ts
```diff
@@ -130,7 +129,7 @@ export async function run({
     functionIds,
   });
 
-  indexingService.processEvents();
+  indexingService?.processEvents();
```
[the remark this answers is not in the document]
jaylmiller: no i forgot to remove that while debugging

packages/core/src/sync-store/store.ts
jaylmiller: the log's tables checkpoint column is currently nullable. but i think once it gets to this part of the code its impossible to be undefined so i can change it

jaylmiller: fixed :)

0xOlias:
Merged into #770

PR #764, opened 2024-03-29
🐛 Fix broken telemetry url

dawsbot:


0xOlias:
Nice catch - thanks for contributing!

PR #771, opened 2024-04-01
Hono

kyscott18:


packages/core/src/server/graphql/schema.test.ts
```diff
@@ -1,9 +1,7 @@
+import { createSchema } from "@/schema/schema.js";
```
0xOlias, line 1: I would probably add this to other test file (`buildGraphqlSchema.test.ts`). Also, would consider renaming `buildGraphqlSchema.ts` to `schema.ts`, but leave it up to you.

packages/core/src/server/service.ts
```diff
@@ -1,197 +1,181 @@
...
-
-  private handleHealthGet(): Handler {
-    return (_, res) => {
-      if (this.isHealthy) {
-        return res.status(200).send();
+    })
+    .get("/health", async (c) => {
+      if (isHealthy) {
+        c.status(200);
+        return c.text("");
```
0xOlias, line 52: Not sure, but you might be able to chain these: `return c.status(200).text("")`

packages/core/src/server/service.test.ts
```diff
@@ -1,1678 +1,231 @@
...
-    const body = (await response.json()) as any;
-    expect(body.errors).toBe(undefined);
+  expect(response.status).toBe(404);
+});
 
-    const { entityWithBigIntIds } = body.data;
-    expect(entityWithBigIntIds.items).toHaveLength(50);
+// Note that this test doesn't work because the `hono.request` method doesn't actually
+// create a socket connection, it just calls the request handler function directly.
+test.skip("kill", async (context) => {
```
0xOlias, line 221: Could delete this test, but also fine to leave it in as documentation.

packages/core/src/server/service.ts
```diff
@@ -1,197 +1,181 @@
...
+
+    const httpServer = serve(
+      {
+        fetch: hono.fetch,
+        createServer: createServerWithNextAvailablePort,
+        port,
+        // Note that common.options.hostname can be undefined if the user did not specify one.
+        // In this case, Node.js uses `::` if IPv6 is available and `0.0.0.0` otherwise.
+        // https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback
+        hostname: common.options.hostname,
```
0xOlias, line 159: It's a bit suspicious to me that we're passing in hostname here as well as above, but it should work fine.

0xOlias (approved):
LGTM!

packages/core/src/server/graphql/schema.test.ts
kyscott18: It's confusing to me that we have a lot of things already called schema

PR #774, opened 2024-04-03
Indexing service refactor

kyscott18:


packages/core/src/bin/utils/run.ts
```diff
@@ -93,53 +100,140 @@ export async function run({
...
+  );
+
+  syncService.on("fatal", onFatalError);
+
+  const finalizedCheckpoint = await syncService.start();
 
   return async () => {
+    runQueue.pause();
+    runQueue.clear();
+    kill(indexingService);
```
0xOlias, line 234: Overall on this file -  I'm not fully convinced that the indexing "guts" (eg, the queue, the setup event processing, etc) should be exposed here rather than within an indexing service. It seems quite important to test the logic here (eg, what happens when different sequences of checkpoint & reorg events get added to the queue)?

packages/core/src/build/service.ts
```diff
@@ -418,36 +413,64 @@ export class BuildService extends Emittery<BuildServiceEvents> {
...
+            `Validation failed: Event '${eventKey}' is excluded by the event filter defined on the contract '${sourceName}'. Got '${eventName}', expected one of [${filteredEventSelectors
+              .map((s) => source.abiEvents.bySelector[s]!.safeName)
               .map((eventName) => `'${eventName}'`)
               .join(", ")}].`,
           );
           return { success: false, error } as const;
         }
       }
+
+      // 1) Update source topic0 to include only registered event selectors
```
0xOlias, line 469: Can remove the `1)` here.

packages/core/src/common/metrics.ts
```diff
@@ -27,9 +27,9 @@ export class MetricsService {
   ponder_realtime_latest_block_timestamp: prometheus.Gauge<"network">;
   ponder_realtime_reorg_total: prometheus.Counter<"network">;
 
-  ponder_indexing_total_seconds: prometheus.Gauge<"event">;
-  ponder_indexing_completed_seconds: prometheus.Gauge<"event">;
-  ponder_indexing_completed_events: prometheus.Gauge<"network" | "event">;
+  ponder_indexing_total_seconds: prometheus.Gauge;
+  ponder_indexing_completed_seconds: prometheus.Gauge;
+  ponder_indexing_completed_events: prometheus.Gauge;
```
0xOlias, line 32: I think we should figure out a way to get indexing function-specific event counts back into the terminal UI.

packages/core/src/indexing/service.test.ts
```diff
@@ -148,120 +97,163 @@ test("processEvents() calls indexing functions with correct arguments", async (c
...
+      },
+      client: expect.any(Object),
+      db: expect.any(Object),
+    },
+  });
+
+  await cleanup();
+});
+
+test("processEvents killed", async (context) => {
```
0xOlias, line 218: `"processEvents does not process events and returns killed status while killed"`

packages/core/src/indexing/service.test.ts
```diff
@@ -419,51 +410,33 @@ test("processEvents() updates metrics", async (context) => {
...
+  const metrics = await common.metrics.ponder_indexing_function_duration.get();
+  expect(metrics.values).toBeDefined();
 
-  await service.kill();
-  await service.onIdle();
   await cleanup();
 });
 
-test("processEvents() reads data from a contract", async (context) => {
+test("executeSetup() retry", async (context) => {
```
0xOlias, line 439: `"with a user error reverts the indexing store and retries"`

packages/core/src/indexing/service.test.ts
```diff
@@ -472,87 +445,136 @@ test("processEvents() reads data from a contract", async (context) => {
...
-  expect(supplyEvents.length).toBe(2);
 
-  await service.kill();
-  await service.onIdle();
   await cleanup();
 });
 
-test("processEvents() recovers from errors while reading data from a contract", async (context) => {
-  const { common, sources, networks, requestQueues } = context;
+test("executeSetup() error", async (context) => {
```
0xOlias, line 489: `"with a consistent user error retries 3 times then fails"

packages/core/src/utils/neva.ts
```diff
@@ -0,0 +1,3 @@
+export const neva = (_x: never) => {
```
0xOlias, line 1: I think we should name this "never" 😄

packages/core/src/sync-store/store.test.ts
```diff
@@ -1443,39 +1409,19 @@ test("getLogEvents filters on multiple filters", async (context) => {
     ...rpcData.block2,
   });
 
-  const { events } = await syncStore.getLogEvents({
+  console.log(rpcData.block2.block.number);
```
0xOlias, line 1412: rm

packages/core/src/indexing/service.ts
```diff
@@ -1,1187 +1,724 @@
...
+          encodedCheckpoint: contextState.encodedCheckpoint,
+          id,
+        });
+      },
+    };
+    return acc;
+  }, {});
+
+  // build ponderActions
+  const ponderActions = <
```
0xOlias, line 264: I think we could rename this to `cachedActions`.

packages/core/src/indexing/service.ts
```diff
@@ -1,1187 +1,724 @@
...
+      endBlock: source.endBlock,
+      maxBlockRange: source.maxBlockRange,
+    };
+  }
 
-        this.common.logger.debug({
-          service: "indexing",
-          msg: `Reverted indexing store to safe timestamp ${safeCheckpoint.blockTimestamp}`,
+  // build db
+  const db = Object.keys(schema.tables).reduce<
```
0xOlias, line 160: Up to you, but this file might be more readable if we moved this and the actions code back into their own files. You could just pass the `contextState` object as an argument.

packages/core/src/indexing/service.ts
```diff
@@ -1,1187 +1,724 @@
...
-    const fromCheckpoint = state.tasksLoadedToCheckpoint;
-    const toCheckpoint = this.syncService.checkpoint;
-
-    if (
-      isCheckpointGreaterThanOrEqualTo(fromCheckpoint, toCheckpoint) &&
-      state.lastEventCheckpoint !== undefined
-    ) {
-      return;
+      // default:
+      //   neva(event);
```
0xOlias, line 476: ?

packages/core/src/indexing/service.ts
```diff
@@ -1,1187 +1,724 @@
...
-    ) {
-      // Subsequent loads
-      state.lastEventCheckpoint = checkpointMax(
-        lastCheckpoint,
-        state.lastEventCheckpoint,
+  for (let i = 0; i < 4; i++) {
+    try {
+      // set currentEvent
+      currentEvent.context.network.chainId = event.chainId;
+      currentEvent.context.network.name = networkByChainId[event.chainId].name;
```
0xOlias, line 572: Could `networkByChainId` include the entire object `{ chainId, name }`?

packages/core/src/indexing/service.ts
```diff
@@ -1,1187 +1,724 @@
...
+      currentEvent.context.client = clientByChainId[event.chainId];
+      currentEvent.context.contracts = contractsByChainId[event.chainId];
```
0xOlias, line 574: Based on this, it looks like you could construct a `contextByChainId` mapping to encapsulate all of this, and then only update the context state here.

0xOlias (commented):
Looking good!

packages/core/src/indexing/service.ts
kyscott18: This code is needed if you have more than one type of event to index

0xOlias: Got it, makes sense!

packages/core/src/indexing/service.ts
kyscott18: I don't think that's possible because `networkByChainId` is also used for the `metricLabel`

packages/core/src/bin/utils/run.ts
```diff
@@ -93,53 +100,140 @@ export async function run({
...
-  syncService.on("reorg", async (checkpoint) => {
-    await indexingService.handleReorg(checkpoint);
-    await indexingService.processEvents();
+  // process setup events
+  const result = await processSetupEvents(indexingService, {
+    sources,
+    networks,
   });
+  if (result.status === "error") {
+    onReloadableError(result.error);
```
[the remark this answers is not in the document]
kyscott18: Nice catch

0xOlias (approved):
LGTM

packages/core/.gitignore
```diff
@@ -1,5 +1,6 @@
 dist/
 .ponder/
+ponder-env.d.ts
```
0xOlias, line 3: I wonder if we should mock the `runCodegen` function in `run.test.ts` and just assert that it gets called.

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -823,277 +831,275 @@ export class PostgresSyncStore implements SyncStore {
...
+    let cursor = encodeCheckpoint(fromCheckpoint);
 
-                return eb.and(exprs);
-              }) ?? [];
+    while (true) {
+      const events = await this.db.wrap(
+        { method: "getLogEvents" },
+        async () => {
+          // Get full log objects, including the eventSelector clause.
+          const requestedLogs = await this.db
```
0xOlias, line 855: @jaylmiller WDYT about this query? I'm a bit worried about an explosion in row count due to the inner join as the # of sources increases.

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -823,277 +831,275 @@ export class PostgresSyncStore implements SyncStore {
...
+              transaction: Transaction;
+              encodedCheckpoint: string;
+            };
+          });
+        },
+      );
 
-      if (events.length === limit + 1) {
-        events.pop();
+      const hasNextPage = events.length > limit;
```
0xOlias, line 1050: It might be easier to follow if you do `events.length === limit + 1`.

packages/core/src/sync-store/sqlite/store.ts
```diff
@@ -823,290 +831,286 @@ export class SqliteSyncStore implements SyncStore {
...
+              "transactions.r as tx_r",
+              "transactions.s as tx_s",
+              "transactions.to as tx_to",
+              "transactions.transactionIndex as tx_transactionIndex",
+              "transactions.type as tx_type",
+              "transactions.value as tx_value",
+              "transactions.v as tx_v",
+            ])
+            .where("logs.checkpoint", ">", cursor)
+            .where("logs.checkpoint", "<=", encodeCheckpoint(toCheckpoint))
```
0xOlias, line 948: Could encode this once, above

packages/core/src/sync-store/postgres/store.ts
jaylmiller: i think if we have their filters (i.e. whatever the equivalent to `logFilterCmprs` and `factoryCmprs`) be based on the primary row the row count should stay okay. if the row count started getting too big for a given set of filters, we could reduce the checkpoint interval which would reduce the number of rows being joined? So i think this query seems good. I'm not sure of a better way to handle fetching this data than what this query is doing, but good to keep it mind, specifically re the checkpoint interval size

PR #790, opened 2024-04-11
fix(sync-store migrations): move column DDL to separate transaction, bump statement timeout

jaylmiller:
- Moving the DDL to a separate transaction prevents locking the entire table during the longer transaction
- Had to bump the statement_timeout for the potentially large update


packages/core/src/sync-store/postgres/migrations.ts
```diff
@@ -457,28 +457,65 @@ const migrations: Record<string, Migration> = {
...
+        .on("logs")
+        .column("blockNumber")
+        .execute();
+    },
+  },
+  "2024_04_11_0_add_checkpoint_column_to_logs_table": {
+    async up(db: Kysely<any>) {
+      if (await hasDoneCheckpointMigration(db)) {
+        return;
+      }
```
0xOlias, line 481: I wonder if it would be better to use `ADD COLUMN IF NOT EXISTS` rather than the manual check (https://stackoverflow.com/a/38721951). Also, I think now we could remove this migration, and do the `ADD COLUMN IF NOT EXISTS` to the "2024_03_20_0_checkpoint_in_logs_table" migration above.

packages/core/src/sync-store/postgres/migrations.ts
```diff
@@ -492,10 +529,10 @@ const migrations: Record<string, Migration> = {
         checkpoint: bigint;
       }>(
         sql`
-          SELECT blocks.timestamp, blocks."chainId", blocks.number, logs."transactionIndex", logs."logIndex", logs.checkpoint
-          FROM ponder_sync.logs logs
-          JOIN ponder_sync.blocks blocks ON logs."blockHash" = blocks.hash
-          LIMIT 10
+        SELECT blocks.timestamp, blocks."chainId", blocks.number, logs."transactionIndex", logs."logIndex", logs.checkpoint
```
0xOlias, line 532: FWIW, I think we'd be fine to remove the check here now that we've confirmed this several times.

0xOlias (commented):
LGTM, just nits. I verified locally using `reference-erc20` by running `pnpm dev` on `0.3.11`, then `0.4.2`, then this branch. Migrations ran very quickly with only ~5000 log rows in the database.

Do you think we need to make any changes for SQLite?

packages/core/src/sync-store/postgres/migrations.ts
jaylmiller: oh! i didnt know this was a thing! i actually tried to do this, but i didnt see it in the kyseley API, so i thought it wasn't a thing at all. i'll change it to use raw SQL. good catch

jaylmiller: Screen Shot 2024-04-14 at 5 33 48 PM
no `ifNotExists` for `addColumn` in kysely. but i think doing the raw SQL is worth it for simplifying the sql

packages/core/src/sync-store/postgres/migrations.ts
jaylmiller: ok yeah makes sense to remove it at this point

jaylmiller:
I can make similar changes for SQLite. It's unlikely the changes will be noticeable if the table isn't very large, but i guess it's possible there are some big SQLite dbs out there

0xOlias (approved):

PR #793, opened 2024-04-12
Refactor build service

kyscott18:


packages/core/src/build/service.ts
```diff
@@ -42,105 +50,137 @@ export type Build = {
...
-    this.viteNodeServer = new ViteNodeServer(this.viteDevServer);
-    installSourcemapsSupport({
-      getSourceMap: (source) => this.viteNodeServer.getSourceMap(source),
-    });
+/**
+ * Execute, validate, and build the files the make up a Ponder app.
+ * If `watch` is true (dev server), then use vite to re-execute changed files,
+ * and validate and build again. This function only re-executes changes files,
+ * but doesn't attempt to skip any validation or build steps.
+ */
```
0xOlias, line 136: Great description

packages/core/src/build/service.ts
```diff
@@ -153,339 +193,223 @@ export class BuildService extends Emittery<BuildServiceEvents> {
...
+        if (executeSchemaResult.status === "error") {
+          onBuild({ status: "error", error: executeSchemaResult.error });
           return;
         }
+        rawBuild.schema = executeSchemaResult.schema;
       }
 
-      const indexingFunctionFiles = invalidated.filter((file) =>
-        this.indexingFunctionRegex.test(file),
+      const hasIndexingFunctionUpdate = invalidated.some((file) =>
```
0xOlias, line 237: Nice, this makes sense.

packages/core/src/config/sources.ts
```diff
@@ -1,16 +1,23 @@
 import type { Abi, Address, Hex, LogTopic } from "viem";
 import type { AbiEvents } from "./abi.js";
 
+type Topic0 = Hex | Hex[];
+export type Topics =
```
0xOlias, line 5: This seems fine. Might be more accurate to name this `TopicFilter`, actually. IMO `Topics` more accurately describes the `RpcLog["topics"]` type.

packages/core/src/indexing/service.ts
```diff
@@ -1,4 +1,4 @@
-import type { IndexingFunctions } from "@/build/functions/functions.js";
+import type { IndexingFunctions } from "@/build/configAndIndexingFunctions/configAndIndexingFunctions.js";
```
0xOlias, line 1: This is an unfortunately long file name - I wonder if perhaps we don't need the additional `schema` and `configAndIndexingFunctions` folders within the `build` dir now.

packages/core/src/sync-historical/service.ts
```diff
@@ -930,11 +930,12 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
    */
   private _eth_getLogs = async (params: {
     address?: Address | Address[];
-    topics?: LogTopic[];
+    topics: Topics;
```
0xOlias, line 933: Repeating a bit, but I'd feel better about this type-level `Topics` change if we rename it to `TopicFilter` or `TopicsWithSelector` or something.

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -842,7 +843,6 @@ export class PostgresSyncStore implements SyncStore {
             .where((eb) => {
               const logFilterCmprs = sources
                 .filter(sourceIsLogFilter)
-                .filter((source) => source.criteria.topics !== undefined)
```
0xOlias, line 845: Great to get rid of these!

0xOlias (commented):
Looks great - build service changes are a nice improvement.

I'll admit, I'm skeptical of the stricter `Topics` type. Looking at the diff, it seems like we're bending over backward for little gain. I think of the required `topic0` as a fun quirk of our current design, rather than a key invariant. Lmk what you think.

kyscott18:
Agree

packages/core/src/build/configAndIndexingFunctions/configAndIndexingFunctions.ts
```diff
@@ -11,16 +11,33 @@ import {
...
+};
+
+export async function buildConfigAndIndexingFunctions({
   config,
+  rawIndexingFunctions,
   options,
-}: { config: Config; options: Pick<Options, "ponderDir" | "rootDir"> }) {
+}: {
+  config: Config;
+  rawIndexingFunctions: RawIndexingFunctions;
```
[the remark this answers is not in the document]
kyscott18: I guess the distinction here is that there is no "raw" version of a config the way that there is with indexing functions because the config is built into "sources" and "networks"

packages/core/src/build/configAndIndexingFunctions/configAndIndexingFunctions.ts
```diff
@@ -361,7 +443,37 @@ export async function buildConfig({
...
-        topics = buildTopics(rawContract.abi, rawContract.filter);
+        const [topic0FromFilter, ...topicsFromFilter] = buildTopics(
+          rawContract.abi,
+          rawContract.filter,
+        );
+
+        // Validate that the topic0 value defined by the `eventFilter` is a superset of the
+        // registered indexing functions. Simply put, no indexing function is defined for a
+        // log event that is filtered out.
+        if (topic0FromFilter !== null) {
```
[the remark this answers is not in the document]
kyscott18: The first element of the array return from `buildTopics` being defined is an invariant of the current filter design. Added a comment to clarify.

packages/core/src/bin/commands/codegen.ts
```diff
@@ -28,28 +32,28 @@ export async function codegen({ cliOptions }: { cliOptions: CliOptions }) {
   const telemetry = new TelemetryService({ options });
   const common = { options, logger, metrics, telemetry };
 
-  const buildService = new BuildService({ common });
-  await buildService.setup({ watch: false });
+  const buildService = await createBuildService({ common });
 
   const cleanup = async () => {
-    await buildService.kill();
+    await killBuildService(buildService);
```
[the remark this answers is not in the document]
kyscott18: What is your opinion about the current api of the indexing service, do you think all functions should be available off the object or just the primitive "start" and "kill"?

packages/core/src/build/configAndIndexingFunctions/configAndIndexingFunctions.ts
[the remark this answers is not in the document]
0xOlias: Got it, makes sense!

packages/core/src/bin/commands/codegen.ts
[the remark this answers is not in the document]
0xOlias: I suppose my preference would be for all of them to be available off the object. And then, I'd probably define them in the `createThing` scope UNLESS they were being tested independently, in which case you could define them at the top-level accepting a `thing` argument and export them.

Not a super strong preference though, happy to go either way.

packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -360,8 +442,38 @@ export async function buildConfig({
...
 
         // TODO: Explicit validation of indexed argument value format (array or object).
+        // The first element of the array return from `buildTopics` being defined
+        // is an invariant of the current filter design.
         // Note: This can throw.
-        topics = buildTopics(rawContract.abi, rawContract.filter);
+        const [topic0FromFilter, ...topicsFromFilter] = buildTopics(
+          rawContract.abi,
+          rawContract.filter,
+        ) as [Exclude<LogTopic, null>, ...LogTopic[]];
```
0xOlias, line 451: I'm still not sure I follow this. Isn't this line converting the user-provided `filter` object into a `topics` value? I feel like instead of asserting the type, we should have a runtime validation here to confirm that the `topic0` value returned from `buildTopics` is of shape `0xabc...` or [`0xabc...`, `0xdef...`].

Look a few lines down, I think this line could get moved out of the loop:
```ts
const filteredEventSelectors = Array.isArray(topic0FromFilter)
  ? topic0FromFilter
  : [topic0FromFilter];
```

And then, we could validate what I described above at runtime. This would validate against a filter that resolves to an empty `topic0`, eg:
```ts
contract: {
  filter: {
    event: []
  }
}
```

0xOlias (approved):
One last question - other than than ready to merge 👍

packages/core/src/build/configAndIndexingFunctions.ts
kyscott18: The filter that the user provides can only result in topic0 that has the shape `Hex | Hex[]`. This is because we force the user to provide an `event` to the filter, which gets directly translated into topic0.

packages/core/src/bin/commands/dev.ts
```diff
@@ -1,6 +1,7 @@
 import { existsSync } from "node:fs";
 import path from "node:path";
-import { type BuildResult, BuildService } from "@/build/service.js";
+import { createBuildService } from "@/build/index.js";
+import { type BuildResult } from "@/build/service.js";
```
0xOlias, line 4: Might make sense to import this from `build/index.js`

packages/core/src/sync-store/store.test.ts
```diff
@@ -57,7 +57,7 @@ test("insertLogFilterInterval inserts block, transactions, and logs", async (con
 
   await syncStore.insertLogFilterInterval({
     chainId: 1,
-    logFilter: { address: erc20.address },
+    logFilter: { address: erc20.address, topics: [null, null, null, null] },
```
0xOlias, line 60: Do we need the `null`s here? Could this just be an empty array?

0xOlias (approved):

packages/core/src/bin/commands/dev.ts
kyscott18: yes

packages/core/src/sync-store/store.test.ts
kyscott18: yes

PR #798, opened 2024-04-13
Improve telemetry

0xOlias:


packages/core/src/common/options.ts
```diff
@@ -78,9 +78,7 @@ export const buildOptions = ({ cliOptions }: { cliOptions: CliOptions }) => {
 
     telemetryUrl: "https://ponder.sh/api/telemetry",
     telemetryDisabled: Boolean(process.env.PONDER_TELEMETRY_DISABLED),
-    telemetryIsExampleProject: Boolean(
-      process.env.PONDER_TELEMETRY_IS_EXAMPLE_PROJECT,
```
kyscott18, line 82: lol never knew we had this

packages/core/src/common/telemetry.test.ts
```diff
@@ -10,30 +13,89 @@ beforeEach(() => {
...
+    telemetry.record({
+      name: "lifecycle:heartbeat_send",
+      properties: { duration_seconds: process.uptime() },
+    });
+  }
+
+  // Mock fetch to take 91ms to complete.
+  fetchSpy.mockImplementation(
+    () => new Promise((resolve) => setTimeout(resolve, 95)),
+  );
```
kyscott18, line 93: Why 91ms? Also doesn't match the number provider to setTimeout

packages/core/src/common/telemetry.ts
```diff
@@ -1,220 +1,272 @@
...
+      properties: { duration_seconds: process.uptime() },
+    });
+  }, HEARTBEAT_INTERVAL_MS);
+
+  const kill = async () => {
+    clearInterval(heartbeatInterval);
+    await new Promise<void>((resolve) => {
+      if (queue.pending === 0) resolve();
+
+      const timeout = setTimeout(() => {
```
kyscott18, line 192: What does this do?

packages/core/src/common/telemetry.ts
```diff
@@ -1,220 +1,272 @@
...
+      const body = JSON.stringify({
+        distinctId: anonymousId,
+        event: task.name,
+        properties,
+      });
 
-  private get anonymousId() {
-    const storedAnonymousId = this.conf.get("anonymousId");
-    if (storedAnonymousId) return storedAnonymousId;
+      await fetch(options.telemetryUrl, {
```
kyscott18, line 167: What is the retry behavior if this fails?

packages/core/src/common/telemetry.test.ts
0xOlias: Oh, these can both be 100 actually now that we're doing >8 and <12 below. I previously had it expecting 10 which appeared flaky.

kyscott18 (approved):

PR #802, opened 2024-04-16
sync api and realtime sync

kyscott18:


packages/core/src/sync-realtime/service.ts
```diff
@@ -8,878 +8,633 @@ import {
...
+ * 3) Determine if a new range of events has become finalized,
+ *    if so validate the local chain before removing the
+ *    finalized data.
+ *
+ * @param pendingLatestBlock Block to be injested. Must be exactly
+ * 1 block ahead of the local chain.
+ * @returns true if a reorg occurred
+ */
+export const handleBlock = async (
+  realtimeSyncService: Service,
```
0xOlias, line 274: You could probably get away with naming this `service` to make this function less verbose.

packages/core/src/sync-realtime/service.ts
```diff
@@ -8,878 +8,633 @@ import {
...
+  { pendingLatestBlock }: { pendingLatestBlock: SyncBlock },
```
0xOlias, line 275: Fine with this naming, but I often think of this as `newHeadBlock`.

packages/core/src/sync-realtime/service.ts
```diff
@@ -8,878 +8,633 @@ import {
...
-      msg: `Finished syncing new head block ${hexToNumber(
-        newBlock.number,
-      )} (network=${this.network.name})`,
+  // "eth_getLogs" calls can be skipped if a negative match is given from "logsBloom".
+  const mustRequestGetLogs =
+    realtimeSyncService.hasFactorySource ||
+    pendingLatestBlock.logsBloom === zeroLogsBloom ||
+    isMatchedLogInBloomFilter({
+      bloom: pendingLatestBlock.logsBloom,
+      logFilters: realtimeSyncService.sources.map((s) => s.criteria),
```
0xOlias, line 299: Does this work if you have both log filters and factories? The type is `criteria: LogFilterCriteria | FactoryCriteria` here.

packages/core/src/sync-realtime/service.ts
```diff
@@ -8,878 +8,633 @@ import {
...
+  // Determine if a new range has become finalized by evaluating if the
+  // latest block number is 2 * finalityBlockCount >= finalized block number.
+  // Essentially, there is a range the width of finalityBlockCount that is entirely
+  // finalized.
+  const blockMovesFinality =
+    pendingLatestBlockNumber >=
+    realtimeSyncService.finalizedBlock.number +
+      2 * realtimeSyncService.network.finalityBlockCount;
+  if (blockMovesFinality) {
+    const pendingFinalizedBlock = realtimeSyncService.localChain.find(
```
0xOlias, line 384: Should be the same result, but I would think of this value as (current finalized block number + finality block count).

packages/core/src/sync-realtime/service.ts
```diff
@@ -8,878 +8,633 @@ import {
...
-      return true;
-    }
-
-    this.common.logger.trace({
-      service: "realtime",
-      msg: `Confirmed local hash matches remote hash at finalized block number ${this.finalizedBlock.number} (network=${this.network.name})`,
+    // TODO: Update this to insert:
+    // 1) Log filter intervals
+    // 2) Factory contract intervals
+    // 3) Child filter intervals
```
0xOlias, line 417: Can remove this comment, it already does all of this.

packages/core/src/sync-realtime/service.ts
```diff
@@ -8,878 +8,633 @@ import {
...
+    await realtimeSyncService.syncStore.insertRealtimeInterval({
```
0xOlias, line 418: Worth noting that you mutate the `realtimeSyncService.localChain` ABOVE this database query, which seems dangerous.

packages/core/src/sync-realtime/service.ts
```diff
@@ -8,878 +8,633 @@ import {
...
+  realtimeSyncService: Service,
+  pendingLatestBlock: SyncBlock,
+) => {
+  // Prune the local chain of blocks that have been reorged out
+  realtimeSyncService.localChain = realtimeSyncService.localChain.filter(
+    ({ block }) => block.number < hexToNumber(pendingLatestBlock.number),
+  );
+
+  // Block we are attempting to fit into the local chain.
+  let reorgedBlock = pendingLatestBlock;
```
0xOlias, line 463: I might call this `remoteBlock`

packages/core/src/sync-realtime/service.ts
```diff
@@ -8,878 +8,633 @@ import {
...
+ *
+ * @param pendingLatestBlock Block that caused reorg to be detected.
+ * Must be at most 1 block ahead of the local chain.
+ */
+export const handleReorg = async (
+  realtimeSyncService: Service,
+  pendingLatestBlock: SyncBlock,
+) => {
+  // Prune the local chain of blocks that have been reorged out
+  realtimeSyncService.localChain = realtimeSyncService.localChain.filter(
```
0xOlias, line 458: Suspicious to me that you'd mutate this at the very top of this file. What if the `deleteRealtimeData` query fails? It would make more sense to me if you use a copy, or just read the local chain at the desired height rather than mutate here.

It seems to me that the only time we mutate in the function is immediately after the `deleteRealtimeData` call.

packages/core/src/sync-realtime/service.ts
```diff
@@ -8,878 +8,633 @@ import {
...
-    if (localLogs.length !== matchedLogs.length) {
-      await handleLogsInconsistency(
-        localLogs.length === 0
-          ? undefined
-          : localLogs[localLogs.length - 1].blockNumber - 1,
-      );
-      return true;
+    if (realtimeSyncService.localChain.length === 0) break;
+    else {
+      reorgedBlock = await _eth_getBlockByHash(realtimeSyncService, {
```
0xOlias, line 500: Previously, we had a little optimization where we'd hold onto these remote blocks as we fetch them, and then add them to the queue once we find a common ancestor. I'm fine to leave that out if you think it might be bug-prone, just thought I'd mention it.

packages/core/src/sync-realtime/service.ts
```diff
@@ -8,878 +8,633 @@ import {
...
-  };
 
-  /** Returns true if "blocks" has a valid chain of block.parentHash to block.hash. */
-  private isChainConsistent = (
-    blocks: (LightBlock | RealtimeBlock)[],
-  ): boolean => {
-    for (let i = blocks.length - 1; i > 1; i--) {
-      if (blocks[i].parentHash !== blocks[i - 1].hash) return false;
+    // Find any logs matching log filters or child contract filters.
+    // NOTE: It might make sense to just insert all logs rather than introduce
```
0xOlias, line 552: I'm pretty confident that we could move this filtering operation into SQL. Something like: SELECT log.id FROM VALUES (... pending logs) LEFT JOIN (factory child address clause) ...

Could you create an issue for it? Could be a good one for a beginner.

packages/core/src/sync-realtime/service.ts
```diff
@@ -8,878 +8,633 @@ import {
...
-}
+  return false;
+};
+
+const getLatestLocalBlock = ({
+  localChain,
+  finalizedBlock,
+}: Pick<Service, "localChain" | "finalizedBlock">) => {
+  if (localChain.length === 0) {
+    return finalizedBlock;
```
0xOlias, line 638: Is there a scenario where this would come into play?

packages/core/src/sync-realtime/service.ts
```diff
@@ -8,878 +8,633 @@ import {
...
+      ({ block }) =>
+        block.number ===
+        pendingLatestBlockNumber -
+          realtimeSyncService.network.finalityBlockCount,
+    )!.block;
+
+    // Validate the local chain by re-requesting logs and checking for inconsistencies
+    // between the local and recently fetched data. This is neccessary because
+    // degraded rpc providers have trouble indexing logs near the tip of the chain.
+    const hasInvalidLogs = await validateLocalBlockchainState(
```
0xOlias, line 394: Seeing as we're not deeming it critical, I think we might want to wrap this call in a try/catch. I could see that eth_getLogs request failing.

packages/core/src/sync-realtime/service.ts
kyscott18: Immediately upon startup, when the local chain is empty or if the entire local chain was re-orged but the finalized block stayed the same.

packages/core/src/sync-realtime/service.ts
kyscott18: Good idea

packages/core/src/sync-realtime/service.ts
0xOlias: Makes sense, thanks

packages/core/src/sync-realtime/service.ts
kyscott18: I don't think the savings are worth the complexity addition

packages/core/src/sync-realtime/service.ts
```diff
@@ -8,878 +8,633 @@ import {
...
+    queue.pause();
+    queue.clear();
+
+    await queue.onIdle();
+  };
 
-    const newBlockNumber = hexToNumber(newBlock.number);
-    const newBlockTimestamp = hexToNumber(newBlock.timestamp);
+  // Note: this is done just for testing.
+  return enqueue().then(() => queue);
```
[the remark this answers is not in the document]
kyscott18: yes it does, needs some retry logic in the enqueue function

packages/core/src/sync-realtime/service.ts
```diff
@@ -8,878 +8,633 @@ import {
...
+            await handleReorg(realtimeSyncService, newBlock);
 
-    if (this.isProcessBlockQueued) {
-      this.isProcessBlockQueued = false;
-      await this.process();
-    } else {
-      this.emit("idle");
-    }
-  };
+            if (realtimeSyncService.isKilled) return;
```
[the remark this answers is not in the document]
kyscott18: We need this in order to quickly exit on kill. We want to avoid enqueing anything after the queue has been paused, which would cause the queue.onIdle to never resolve.

Right now the kill function sets the isKilled state variable, pauses and clears the queue, and then waits for it to become idle (no pending or unstarted tasks)

PR #805, opened 2024-04-16
Fix factory logs bug

0xOlias:
This PR fixes a bug in the internal `getFactoryChildAddresses` method. Previously, the method attempted to paginate on `blockNumber` without setting an `ORDER BY` on that field, leading to duplicate results. This bug went undetected for quite some time, and only caused an actual error in `0.4` when we began inserting logs in bulk during the historical sync, because the bulk upsert query we are now using throws an error if you attempt to insert two logs with the same ID in the same batch.

### DIFF
--- a/packages/core/src/sync-store/postgres/store.ts
+++ b/packages/core/src/sync-store/postgres/store.ts
@@ -270,32 +270,28 @@ export class PostgresSyncStore implements SyncStore {
       .where("address", "=", address)
       .where("topic0", "=", eventSelector)
       .where("blockNumber", "<=", upToBlockNumber)
+      .orderBy("blockNumber", "asc")
       .limit(pageSize);
 
     let cursor: bigint | undefined = undefined;
 
     while (true) {
       let query = baseQuery;
-
-      if (cursor) {
-        query = query.where("blockNumber", ">", cursor);
-      }
+      if (cursor !== undefined) query = query.where("blockNumber", ">", cursor);
 
       const batch = await this.db.wrap(
         { method: "getFactoryChildAddresses" },
         () => query.execute(),
       );
 
-      const lastRow = batch[batch.length - 1];
-      if (lastRow) {
-        cursor = lastRow.blockNumber;
-      }
-
       if (batch.length > 0) {
         yield batch.map((a) => a.childAddress);
       }
 
+      // If the batch is less than the page size, there are no more pages.
       if (batch.length < pageSize) break;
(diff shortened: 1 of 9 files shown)


packages/core/src/sync-store/postgres/store.ts
```diff
@@ -270,32 +270,28 @@ export class PostgresSyncStore implements SyncStore {
...
-      }
-
       if (batch.length > 0) {
         yield batch.map((a) => a.childAddress);
       }
 
+      // If the batch is less than the page size, there are no more pages.
       if (batch.length < pageSize) break;
+      // Otherwise, set the cursor to the last block number in the batch.
+      cursor = batch[batch.length - 1].blockNumber;
```
jaylmiller, line 294: I'm probably just confused about the context but for my own understanding why does paginating like this work in this case? Like if u set the cursor to the last block num in the batch, how do you know that the next iteration is landing on the next row (like if theres multiple rows with the same blockNumber, the next time you run the query the first row you hit probably won't be the next row in the "page")

0xOlias: Great call out, that was also a bug. Per DMs, should be fixed now by using `id` as the pagination cursor.

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -265,37 +267,34 @@ export class PostgresSyncStore implements SyncStore {
...
         () => query.execute(),
       );
 
-      const lastRow = batch[batch.length - 1];
-      if (lastRow) {
-        cursor = lastRow.blockNumber;
-      }
-
       if (batch.length > 0) {
         yield batch.map((a) => a.childAddress);
```
kyscott18, line 291: Just for my understanding, if the page size is 1,000, this will return 1,001 addresses?

kyscott18 (approved):
Nice fix

packages/core/src/sync-store/postgres/store.ts
0xOlias: I don't think so, I think it would still return 1000. The `+ 1` trick we use in `findMany` is required to support the bidirectional pagination, but I don't think we need it here. Maybe you're pointing out something else though - lmk.

kyscott18: Makes sense - the pagination logic doesn't attempt to check if the next page is going to have any events or not

PR #806, opened 2024-04-17
Windows

### LINKED ISSUE #421: Support Windows without WSL
It sounds like we still have some file path issues on Windows without WSL. I don't think this is a large set of users, but it might be a low lift to get it working. Would be an awesome external contribution. Also might be nice to have some way to verify it on CI so we avoid regressions.

> When I run the command "npm run dev," it auto-generates a .ponder folder. However, I encounter an error stating, "Error during build: Cannot find module 'D:\\GitHub\\friendtech-ponder\\friendtech\\.ponder\\out\\generated\\index.js'". In the .ponder directory, there is an 'out' directory, but no 'generated' subdirectory, which suggests a possible error on Ponder's side.

0xOlias:
This PR adds support for Windows without WSL, and adds Windows to the CI matrix. It also addresses an issue where our logger was not being cleaned up properly (this only caused a bug on Windows filesystems).

I also renamed our `setupContext` test hook to `setupCommon`, and stopped calling it from the `_test/setup.ts` file in favor of including it explicitly in each test file where it is required.

Fixes #421


packages/core/src/_test/e2e/erc20/erc20.test.ts
```diff
@@ -1,36 +1,36 @@
...
 import { range } from "@/utils/range.js";
 import { wait } from "@/utils/wait.js";
+import { rimrafSync } from "rimraf";
 import { zeroAddress } from "viem";
-import { afterEach, beforeEach, describe, expect, test } from "vitest";
+import { beforeEach, describe, expect, test } from "vitest";
 
+const rootDir = path.join(".", "src", "_test", "e2e", "erc20");
+beforeEach(() => {
+  rimrafSync(path.join(rootDir, ".ponder"));
```
kyscott18, line 20: Seems crazy to me that nodejs filesystem apis don't handle windows

kyscott18 (approved):
pondows

packages/create-ponder/vite.config.ts
```diff
@@ -10,5 +10,6 @@ export default defineConfig({
   },
   test: {
     globalSetup: ["src/_test/globalSetup.ts"],
+    testTimeout: 60_000,
```
jaylmiller, line 13: this seems very high for tests. why do we need a whole minute here?

jaylmiller: i usually prefer having the tests fail fast if theres something hanging or something that would cause it to take this long

jaylmiller (approved):
looks good! not a blocker but we should consider not having this super long timeout for our tests

packages/create-ponder/vite.config.ts
0xOlias: Good point. I increased this mostly because they were timing out consistently on my local Windows VM, and I see them timeout occasionally in CI.

These tests are slow for two reasons.
1. They make HTTP requests to Etherscan to fetch ABIs and start blocks. If the user has not provided an etherscan API key, we wait 5 seconds between each request to avoid getting rate limited. (But in CI, we do provide the API key so it's pretty fast).
2. They run `pnpm install` in the generated ponder project, which can be slow, esp if building `better-sqlite3` from source.

These tests rarely fail for any reason other than a timeout (we don't often make changes to `create-ponder`. So on balance I think it's better to set a high timeout simply to save us the annoyance of having the retry them to get a green CI run.

jaylmiller: makes sense! in the future we may want to consider having some tests have a smaller time out to increase local development feedback loop time, but thats pretty minor

kyscott18 (approved):

PR #808, opened 2024-04-17
fix: indexing with overloaded events

kyscott18:


0xOlias (approved):
LGTM, nice fix and test.

PR #820, opened 2024-04-22
Optimize query speed by adding indexes for reference columns

ualtinok:
I didn't add for sqlite as I assume it will be only used in development


0xOlias:
Closed in favor of https://github.com/ponder-sh/ponder/pull/852 (I DM'd you with details)

PR #823, opened 2024-04-22
Receipts

0xOlias:


packages/core/src/sync-store/sqlite/store.ts
```diff
@@ -434,30 +485,37 @@ export class SqliteSyncStore implements SyncStore {
...
-                cmpr("fragmentEventSelector", "=", sql.ref("eventSelector")),
-                cmpr(
+            let baseJoin = join.on((eb) =>
+              eb.and([
+                eb("fragmentAddress", "=", sql.ref("address")),
+                eb("fragmentEventSelector", "=", sql.ref("eventSelector")),
+                eb(
                   "fragmentChildAddressLocation",
-                  "=",
+                  "<=",
```
kyscott18, line 501: This seems like a merge conflict or something that I did, should be "<=" for the transactionReceipts boolean

packages/core/src/sync-store/sqlite/store.ts
```diff
@@ -56,13 +62,15 @@ export class SqliteSyncStore implements SyncStore {
...
     transactions: rpcTransactions,
+    transactionReceipts: rpcTransactionReceipts,
     logs: rpcLogs,
     interval,
   }: {
     chainId: number;
     logFilter: LogFilterCriteria;
     block: RpcBlock;
     transactions: RpcTransaction[];
+    transactionReceipts?: RpcTransactionReceipt[];
```
0xOlias, line 73: I think I'd prefer this to only have one falsey state (empty array) just like the other options.

packages/core/src/utils/fragments.ts
```diff
@@ -42,11 +44,12 @@ export function buildFactoryFragments({
   const fragments = buildFragments({
     address,
     topics,
+    includeTransactionReceipts,
     chainId,
     childAddressLocation,
     eventSelector,
     idCallback: (address_, topic0_, topic1_, topic2_, topic3_) =>
-      `${chainId}_${address_}_${eventSelector}_${childAddressLocation}_${topic0_}_${topic1_}_${topic2_}_${topic3_}`,
+      `${chainId}_${address_}_${eventSelector}_${childAddressLocation}_${topic0_}_${topic1_}_${topic2_}_${topic3_}_${includeTransactionReceipts}`,
```
0xOlias, line 52: I think there's a bug in this file somewhere that is causing the fragment IDs to sometimes use `_false` instead of `_0`.

packages/core/src/sync-store/sqlite/store.ts
kyscott18: Good call

PR #834, opened 2024-04-25
fix: update empty object bug

0xOlias:


packages/core/src/indexing-store/historicalStore.ts
```diff
@@ -112,7 +112,7 @@ export const getHistoricalIndexingStore = ({
       } else {
         updateObject = data;
       }
-      const updateRow = encodeRow(updateObject, table, kind);
+      const updateRow = encodeRow({ id, ...updateObject }, table, kind);
```
0xOlias, line 115: For posterity - we have to include the `id` here to handle the case where `updateObject = {}`. Otherwise, it's SQL syntax error because the SET clause has no values in it.

packages/core/src/indexing-store/historicalStore.ts
```diff
@@ -280,18 +284,13 @@ export const getHistoricalIndexingStore = ({
...
           .withSchema(namespaceInfo.userNamespace)
           .insertInto(tableName)
           .values(createRow)
-          .onConflict((oc) =>
-            // If the updateRow is empty, DO UPDATE SET throws a syntax error.
-            Object.keys(updateRow).length === 0
-              ? oc.column("id").doNothing()
-              : oc.column("id").doUpdateSet(updateRow),
-          )
+          .onConflict((oc) => oc.column("id").doUpdateSet(updateRow))
```
0xOlias, line 293: Including the `id` is better, because if the `DO NOTHING` block hits, the query doesn't return any data - `RETURNING` only works if a row was modified.

PR #844, opened 2024-04-29
`PoolConfig` option

0xOlias:


packages/core/src/database/postgres/service.ts
```diff
@@ -40,56 +41,77 @@ export class PostgresDatabaseService implements BaseDatabaseService {
...
-      // 10 minutes to accommodate slow sync store migrations.
-      statement_timeout: 10 * 60 * 1000,
+      max: internalMax,
+      statement_timeout: 10 * 60 * 1000, // 10 minutes to accommodate slow sync store migrations.
     });
-    this.indexingPool = createPool(this.poolConfig);
-    this.syncPool = createPool(this.poolConfig);
+    this.syncPool = createPool({ ...poolConfig, max: readonlyMax });
+    this.indexingPool = createPool({ ...poolConfig, max: indexingMax });
+    this.readonlyPool = createPool({ ...poolConfig, max: syncMax });
```
0xOlias, line 92: @jaylmiller this is the section to focus on.

jaylmiller: let's update this `createPool` helper to set the `application_name` for the connection. not sure if it needs to be configurable, but for example, it would really assist in debugging if it was set to RAILWAY_DEPLOY_ID (or w/e they call the env var that identifies a deploy)

jaylmiller (commented):
looks good to me otherwise

packages/core/src/database/postgres/service.ts
jaylmiller: that way you could see all the connections for a given deploy by doing
```sql
select * from pg_stat_activity where application_name=<deploy id>
```

jaylmiller: or maybe set application name to like `<deploy_id>_indexing` to distinguish between the different pools we spin up?

PR #846, opened 2024-04-30
Metrics tweaks

0xOlias:
- Adds HTTP server metrics
- Fixes two broken metrics
- Removes dead code from database service
- Reduces the namespace lock timeout from 60 seconds -> 25 seconds

### DIFF
--- a/packages/core/src/common/metrics.ts
+++ b/packages/core/src/common/metrics.ts
@@ -18,8 +18,15 @@ const httpRequestSizeBytes = [
 export class MetricsService {
   registry: prometheus.Registry;
 
-  ponder_rpc_request_duration: prometheus.Histogram<"network" | "method">;
-  ponder_rpc_request_lag: prometheus.Histogram<"network" | "method">;
+  ponder_indexing_total_seconds: prometheus.Gauge;
+  ponder_indexing_completed_seconds: prometheus.Gauge;
+  ponder_indexing_completed_events: prometheus.Gauge<"network" | "event">;
+
+  ponder_indexing_completed_timestamp: prometheus.Gauge;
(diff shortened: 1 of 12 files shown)


packages/core/src/common/metrics.ts
```diff
@@ -126,86 +155,69 @@ export class MetricsService {
...
     });
-    this.ponder_database_method_error_total = new prometheus.Counter({
-      name: "ponder_database_method_error_total",
-      help: "Total number of errors encountered during database operations",
-      labelNames: ["service", "method"] as const,
+    this.ponder_rpc_request_lag = new prometheus.Histogram({
+      name: "ponder_rpc_request_lag",
+      help: "Time RPC requests spend waiting in the request queue",
+      labelNames: ["network", "method"] as const,
+      buckets: httpRequestDurationMs,
```
0xOlias, line 216: This should probably use the database method buckets.

kristofgazso:
ponder_indexing_completed_timestamp and ponder_indexing_has_error should probably have 'network' labels, no?

0xOlias:
> `ponder_indexing_completed_timestamp` and `ponder_indexing_has_error` should probably have 'network' labels, no?

Not a bad idea - unfortunately to avoid reporting misleading metrics, we'd have to update the metric after every event (in the hot loop) which we suspect is bad for perf. So will punt until we have a chance to benchmark. That said, we're aware that observability around indexing progress + liveness is bad. Will give it proper attention soon.

And for `ponder_indexing_has_error `, it would accurately communicate which network the errored event was on, but the logs already offer much more actionable detail about the error (network, block number, event args, stack trace + code frame).

PR #852, opened 2024-05-02
Update internal schema representation

kyscott18:


packages/core/src/schema/utils.ts
```diff
@@ -1,59 +1,73 @@
...
+  return column[" list"];
+};
+
+export const isTable = (tableOrEnum: Schema[string]): tableOrEnum is Table =>
+  Array.isArray(tableOrEnum) === false;
+
+export const isEnum = (tableOrEnum: Schema[string]): tableOrEnum is Enum =>
+  Array.isArray(tableOrEnum) === true;
+
+export const schemaToTables = (
```
0xOlias, line 45: Maybe `getTables`, `getEnums`? In context, it will always be called with a `schema` object as the arg so it's kind of self-documenting.

packages/core/src/server/graphql/entity.ts
```diff
@@ -97,7 +98,8 @@ export const buildEntityTypes = ({
               const whereObject = where ? buildWhereObject(where) : {};
               // Add the parent record ID to the where object.
               // Note that this overrides any existing equals condition.
-              (whereObject[column.referenceColumn] ??= {}).equals = parent.id;
+              (whereObject[column[" referenceColumn"]] ??= {}).equals =
```
0xOlias, line 101: Doesn't look like we need the `??=` anymore with the ternary above?

packages/core/src/schema/columns.ts
```diff
@@ -1,366 +1,298 @@
...
-  TReferences extends `${string}.id` | undefined,
-  TList extends boolean,
-> = () => TReferences extends undefined
-  ? TList extends true
-    ? InternalColumn<TScalar, TReferences, true, TList>
-    : InternalColumn<TScalar, TReferences, true, TList> & {
-        /**
-         * Mark the column as optional.
-         *
-         * - Docs: https://ponder.sh/docs/guides/design-your-schema#optional
```
0xOlias, line 21: Did the JSDoc in this file get moved somewhere else, or deleted?

packages/core/src/indexing/ponderActions.ts
```diff
@@ -184,106 +184,107 @@ export const buildDb = ({
...
-        return indexingStore.delete({
-          tableName,
-          encodedCheckpoint: contextState.encodedCheckpoint,
-          id,
-        });
-      },
-    };
-    return acc;
-  }, {});
+  return Object.keys(schema).reduce<Service["currentEvent"]["context"]["db"]>(
```
0xOlias, line 187: Is this correct? Looks like we might be missing a `schemaToTables`?

packages/core/src/indexing-store/utils/sort.ts
```diff
@@ -32,12 +38,17 @@ export function buildOrderByConditions({
       `Invalid sort. Column does not exist. Got '${columnName}', expected one of [${Object.keys(
         table,
       )
-        .filter((key) => isBaseColumn(table[key]) || isEnumColumn(table[key]))
+        .filter(
+          (columnName) =>
+            isScalarColumn(table[columnName]) ||
+            isReferenceColumn(table[columnName]) ||
```
0xOlias, line 44: So now, a reference column is not a scalar column?

packages/core/src/indexing-store/utils/sort.ts
```diff
@@ -13,7 +19,7 @@ export function buildOrderByConditions({
   table,
 }: {
   orderBy: OrderByInput<any> | undefined;
-  table: Schema["tables"][keyof Schema["tables"]];
+  table: Table;
```
0xOlias, line 22: Nice diff lol

packages/core/src/indexing-store/utils/cursor.ts
```diff
@@ -1,11 +1,11 @@
 import { StoreError } from "@/common/errors.js";
+import type { UserRow } from "@/types/schema.js";
 import { deserialize, serialize } from "@/utils/serialize.js";
 import type { ExpressionBuilder } from "kysely";
-import type { Row } from "../store.js";
 import type { OrderByConditions } from "./sort.js";
 
 export function encodeCursor(
-  record: Row,
+  record: UserRow,
```
0xOlias, line 8: Seems a hint that the type should be called `UserRecord`. Not sure though. In the docs we do use "record" rather than "row".

packages/core/src/indexing-store/store.ts
```diff
@@ -1,11 +1,17 @@
...
+} from "@/types/schema.js";
 import type { Prettify } from "@/types/utils.js";
 import type { Hex } from "viem";
 
 export type ReadonlyStore = {
   findUnique(options: {
     tableName: string;
-    id: string | number | bigint;
-  }): Promise<Row | null>;
+    id: UserIdColumn;
```
0xOlias, line 13: This type could just be called `UserId`. Up to you.

packages/core/src/database/postgres/service.ts
```diff
@@ -683,28 +697,28 @@ export class PostgresDatabaseService implements BaseDatabaseService {
...
       if (isManyColumn(column)) return;
       if (isEnumColumn(column)) {
         // Handle enum types
         // Omit the CHECK constraint because its included in the user table
         builder = builder.addColumn(columnName, "text");
-      } else if (column.list) {
-        // Handle scalar list columns
-        builder = builder.addColumn(columnName, "text");
-      } else {
+      } else if (isReferenceColumn(column) || isListColumn(column) === false) {
```
0xOlias, line 709: Don't quite follow this change.

packages/core/src/indexing-store/utils/sort.ts
kyscott18: correct

packages/core/src/schema/utils.ts
```diff
@@ -71,3 +69,10 @@ export const schemaToEnums = (schema: Schema): { [enumName: string]: Enum } => {
 export const extractReferenceTable = (ref: ReferenceColumn): string => {
   return ref[" reference"].split(".")[0];
 };
+
+export const encodeSchema = (schema: Schema) => {
```
0xOlias, line 73: Do we need a corresponding `decodeSchema`?

kyscott18: At first I didn't have it because the postgres and sqlite return types are different

PR #854, opened 2024-05-03
Allow custom indexes to be defined in "ponder.schema.ts"

kyscott18:
```js
import { createSchema } from "@ponder/core";

export default createSchema((p) => ({
  Account: p.createTable(
    {
      id: p.hex(),
      balance: p.bigint(),
      isOwner: p.boolean(),
    },
    {
      balanceIndex: p.index(["balance", "isOwner"]),
    },
  ),
}));
```

### DIFF
--- a/packages/core/src/bin/utils/run.ts
+++ b/packages/core/src/bin/utils/run.ts
@@ -270,7 +270,7 @@ export async function run({
     }
     await handleFinalize(syncService.finalizedCheckpoint);
 
-    await database.addIndexes({ schema });
+    await database.createIndexes({ schema });
 
     server.setHealthy();
     common.logger.info({
--- a/packages/core/src/bin/utils/run.ts
+++ b/packages/core/src/bin/utils/run.ts
@@ -270,6 +270,8 @@ export async function run({
     }
     await handleFinalize(syncService.finalizedCheckpoint);
 
+    await database.addIndexes({ schema });
+
     server.setHealthy();
     common.logger.info({
       service: "server",
--- a/packages/core/src/build/schema.ts
+++ b/packages/core/src/build/schema.ts
@@ -10,6 +10,7 @@ import {
   schemaToEnums,
   schemaToTables,
 } from "@/schema/utils.js";
+import { dedupe } from "@ponder/common";
 
 export const buildSchema = ({ schema }: { schema: Schema }) => {
   // Validate enums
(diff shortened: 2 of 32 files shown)


packages/core/src/build/schema.ts
```diff
@@ -28,202 +31,277 @@ export const buildSchema = ({ schema }: { schema: Schema }) => {
...
-    // Validate all other columns
-    Object.entries(table).forEach(([columnName, column]) => {
-      if (columnName === "id") return;
-
-      validateTableOrColumnName(columnName, "Column");
-
-      if (isOneColumn(column)) {
-        const usedColumn = Object.entries(table).find(
-          ([c]) => c === column[" reference"],
+  Object.entries(getTables(schema)).forEach(
```
kyscott18, line 34: Github did a bad job of rendering the diff here. I went back through and manually validated the changes

packages/core/src/bin/utils/run.ts
```diff
@@ -270,6 +270,8 @@ export async function run({
     }
     await handleFinalize(syncService.finalizedCheckpoint);
 
+    await database.addIndexes({ schema });
```
0xOlias, line 273: I think this should be called `createIndexes` to better match SQL terminology

packages/core/src/schema/schema.ts
```diff
@@ -79,8 +83,21 @@ type GetTable<
...
+> = {} extends constraints
+  ? {}
+  : {
+      [name in keyof constraints]: Index<columnName | readonly columnName[]>;
+    };
+
+export const createTable = <const table, const constraints>(
+  t: GetTable<table>,
+  c?: GetConstraints<constraints, table>,
+): readonly [table, constraints] => [t as table, c as constraints];
```
0xOlias, line 100: I think I'd find this easier to follow as `{ columns, constraints }`, but open if there's a reason not to do that.

packages/core/src/database/sqlite/service.ts
```diff
@@ -498,6 +498,42 @@ export class SqliteDatabaseService implements BaseDatabaseService {
...
+            const columns = Array.isArray(indexColumn)
+              ? indexColumn.map((ic) => `"${ic}"`).join(", ")
+              : `"${indexColumn}"`;
+
+            await this.db.executeQuery(
+              sql`CREATE INDEX "${sql.raw(this.userNamespace)}"."${sql.raw(
+                indexName,
+              )}" ON "${sql.raw(tableName)}" (${sql.raw(columns)})`.compile(
+                this.db,
+              ),
```
0xOlias, line 520: Let's be careful to use the correct `sql` functions here to ensure that we're escaping user input.

https://kysely-org.github.io/kysely-apidoc/interfaces/Sql.html

kyscott18:
I think we need a line in the "query the database" portion of the docs that says something like `In some cases, query performance can be dramatically improved by (creating a database index)[...]. `

PR #860, opened 2024-05-04
Improve realtime sync resilience

kyscott18:


packages/core/src/sync-realtime/service.ts
```diff
@@ -183,89 +187,95 @@ export const start = (service: Service) => {
...
+        if (service.isKilled) return;
+
+        const error = _error as Error;
+
+        service.common.logger.warn({
+          service: "realtime",
+          msg: `Failed to process '${service.network.name}' block ${newHeadBlockNumber} with error: ${error}`,
+        });
+
+        await wait(250 * 2 ** service.consecutiveErrors);
```
0xOlias, line 264: We should be thoughtful about the backoff & limit.

packages/core/src/sync-realtime/service.ts
```diff
@@ -183,89 +187,95 @@ export const start = (service: Service) => {
...
+
+        // Remove all blocks from the queue. This protects against an
+        // erroneous block causing a fatal error.
+        queue.clear();
```
0xOlias, line 268: Are there any scenarios where we'd be wasting a lot of valid "work" by doing this here?

0xOlias (commented):
Looking good, I expect this will be a very good change.

packages/core/src/sync-realtime/service.ts
kyscott18: Yes, one example would be that the `wait` block is executing and meanwhile real blocks are getting added to the queue. Then once the `wait` finished, the real blocks are instantly cleared. We could add a `paused` state but only if you think it is worth the complexity

packages/core/src/sync-realtime/service.test.ts
```diff
@@ -267,15 +267,14 @@ test("start() retries on error", async (context) => {
...
-  expect(insertSpy).toHaveBeenCalledTimes(6);
+  expect(realtimeSyncService.localChain).toHaveLength(0);
 
   await kill(realtimeSyncService);
 
   await cleanup();
 });
 
-test("start() emits fatal error", async (context) => {
+test.skip("start() emits fatal error", async (context) => {
```
kyscott18, line 277: Need to figure out how to better mock vitest timers, right now the timers are too long to run in a test.

0xOlias (approved):

PR #865, opened 2024-05-06
Max operation tokens validation

0xOlias:


packages/core/src/server/service.test.ts
```diff
@@ -27,7 +27,7 @@ test("port", async (context) => {
     readonlyStore: {} as ReadonlyStore,
   });
 
-  expect(server2.port).toBe(server1.port + 1);
+  expect(server2.port).toBeGreaterThanOrEqual(server1.port + 1);
```
0xOlias, line 30: This fixes a flake I've seen in CI where a server starts up in a different thread and takes the port directly in between `server1` and `server2` here.

packages/core/src/server/graphql/buildGraphqlSchema.ts
```diff
@@ -15,15 +14,7 @@ import { buildSingularField } from "./singular.js";
...
-export type Context = HonoContext<{
-  Variables: { store: IndexingStore; getLoader: GetLoader };
-}> & {
-  get: {
-    <key extends "store" | "getLoader">(_: key): key extends "store"
-      ? IndexingStore
-      : GetLoader;
-  };
-};
+export type Context = { store: IndexingStore; getLoader: GetLoader };
```
kyscott18, line 17: Do you know why hono wants the context to be accessed with get/set? Might be possible that they are doing some splitting or static analysis under the hood.

0xOlias: With this change, we're drawing a distinction between the HTTP request context (managed by e.g. Hono, express) and the GraphQL context. I'm not sure why Hono uses get/set.

kyscott18 (approved):

PR #867, opened 2024-05-08
Add call traces as an event source

### LINKED ISSUE #113: [Feature] Transaction call indexing
Ponder could support call handlers:

```ts
ponder.on("Contract:mint()", async ({ event, context }) => {
  console.log(event.input.tokenId)
})
```

This feature would only be available when using a node/RPC provider that implements the non-standard `trace_filter` method. Currently, this method is supported by:
- Alchemy
- Quicknode (with Trace Mode add-on)
- OpenEthereum and Erigon nodes

It is NOT supported by:
- Anvil
- Hardhat Network
- ~~Infura~~ (Supported as of 6/8/23)
- Moralis

kyscott18:
Use the `trace_` namespace JSON RPC methods to allow for indexing call traces.

```ts
ponder.on("Erc20.transfer()", async ({ event, context }) => {
  console.log(event.args); // ["0x1234...", 100000...]
  console.log(event.result); // true
})
```

Closes #113


kristofgazso:
hey Kyle!

packages/core/src/config/config.ts
```diff
@@ -97,6 +97,10 @@ type TransactionReceiptConfig = {
   includeTransactionReceipts?: boolean;
 };
 
+type FunctionCallConfig = {
+  includeFunctionCalls?: boolean;
```
kyscott18, line 101: I think we should change this to be `includeCallTraces`

packages/core/src/_test/setup.ts
```diff
@@ -263,7 +264,12 @@ export async function setupAnvil(context: TestContext) {
   );
   context.networks = networks;
   context.requestQueues = requestQueues;
-  context.sources = sources as [LogSource, FactorySource, BlockSource];
+  context.sources = sources as [
+    LogSource,
+    FactorySource,
+    CallTraceSource,
```
kyscott18, line 270: The ordering of the sources changed which is gonna show up in the test files diff

pnpm-lock.yaml
```diff
@@ -4543,49 +4565,6 @@ packages:
       pretty-format: 29.7.0
     dev: true
 
-  /@wagmi/cli@1.5.2:
```
kyscott18, line 4546: Seems odd, not really sure how this happened

packages/core/src/sync-store/postgres/migrations.ts
```diff
@@ -686,14 +686,125 @@ const migrations: Record<string, Migration> = {
...
+        .addColumn("subtraces", "integer", (col) => col.notNull())
+        .addColumn("traceAddress", "text", (col) => col.notNull())
+        .addColumn("transactionHash", "varchar(66)", (col) => col.notNull())
+        .addColumn("transactionPosition", "integer", (col) => col.notNull())
+        .addColumn("functionSelector", "varchar(10)", (col) => col.notNull())
+        .addColumn("chainId", "integer", (col) => col.notNull())
+        .addColumn("checkpoint", "varchar(75)", (col) => col.notNull())
+        .execute();
+
+      // The callTraces.blockNumber index supports getLogEvents and deleteRealtimeData
```
0xOlias, line 744: `getEvents`?

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -669,6 +677,225 @@ export class PostgresSyncStore implements SyncStore {
...
+            .insertInto("transactionReceipts")
+            .values(transactionReceipts)
+            .onConflict((oc) => oc.column("transactionHash").doNothing())
+            .execute();
+        }
+
+        // Delete existing traces with the same `transactionHash`. Then, calculate "callTraces.checkpoint"
+        // based on the ordering of "callTraces.traceAddress" and add all traces to "callTraces" table.
+        const traceByTransactionHash: {
+          [transacionHash: Hex]: SyncCallTrace[];
```
0xOlias, line 738: typo

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -669,6 +677,225 @@ export class PostgresSyncStore implements SyncStore {
...
+              eventIndex: BigInt(i),
+            });
+
+            trace.checkpoint = checkpoint;
+          }
+
+          await tx
+            .insertInto("callTraces")
+            .values(traces)
+            .onConflict((oc) => oc.column("id").doNothing())
```
0xOlias, line 784: I don't think it's really possible to have a conflict here considering we just deleted above, but it's double checking.

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -669,6 +677,225 @@ export class PostgresSyncStore implements SyncStore {
...
+        } = {};
+        for (const trace of rpcTraces) {
+          if (traceByTransactionHash[trace.transactionHash] === undefined) {
+            traceByTransactionHash[trace.transactionHash] = [];
+          }
+          traceByTransactionHash[trace.transactionHash].push(trace);
+        }
+
+        for (const transactionHash of Object.keys(traceByTransactionHash)) {
+          const traces = await tx
```
0xOlias, line 748: I'm not sure, but it might be worth batching this using `.where("transactionHash", "in", transactionHashes)`, doing the reordering in-memory, and then inserting all of the merged traces in one query at the end. It's plausible that N could be >100 here.

packages/core/src/sync-store/sqlite/encoding.ts
```diff
@@ -209,6 +209,54 @@ export function rpcToSqliteLog(log: RpcLog): Omit<InsertableLog, "chainId"> {
...
+    blockHash: trace.blockHash,
+    blockNumber: encodeAsText(trace.blockNumber),
+    error: trace.error ?? null,
+    gasUsed: encodeAsText(trace.result.gasUsed),
+    output: trace.result.output,
+    subtraces: trace.subtraces,
+    traceAddress: JSON.stringify(trace.traceAddress),
+    transactionHash: trace.transactionHash,
+    transactionPosition: trace.transactionPosition,
+    functionSelector: trace.action.input.slice(0, 10) as Hex,
```
0xOlias, line 256: Do we need to consider casing here?

packages/core/src/sync/events.ts
```diff
@@ -78,8 +148,15 @@ export const decodeEvents = (
             topics: event.log!.topics,
           });
 
+          if (
+            event.log!.topics[0] === undefined ||
+            source.abiEvents.bySelector[event.log!.topics[0]] === undefined
+          ) {
+            throw new Error();
+          }
```
0xOlias, line 156: What does this guard against?

0xOlias (commented):
Looking great!

packages/core/src/sync-historical/service.ts
```diff
@@ -1013,15 +1228,165 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
...
+    );
+
+    const revertedTransactions = new Set<Hash>();
+    for (const receipt of transactionReceipts) {
+      if (receipt.status === "0x0") {
+        revertedTransactions.add(receipt.transactionHash);
+      }
+    }
+
+    const persistentTraces = traces.filter(
```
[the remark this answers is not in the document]
kyscott18: I previously had that name but it was kinda confusing to me because some of the traces can have errors so they are technically "unsuccessful"

packages/core/src/build/configAndIndexingFunctions.test.ts
```diff
@@ -495,6 +504,46 @@ test("buildConfigAndIndexingFunctions() includeTransactionReceipts", async () =>
...
+    networks: {
+      mainnet: { chainId: 1, transport: http("http://127.0.0.1:8545") },
+      optimism: { chainId: 10, transport: http("http://127.0.0.1:8545") },
+    },
+    contracts: {
+      a: {
+        includeFunctionCalls: true,
+        network: {
+          mainnet: {},
+          optimism: { includeFunctionCalls: false },
```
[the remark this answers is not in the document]
kyscott18: Yes I wanted to test the override

packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -307,287 +315,360 @@ export async function buildConfigAndIndexingFunctions({
...
-          `Validation failed: Contract '${baseContract.contractName}' cannot specify both 'factory' and 'address' options.`,
-        );
-      }
+          const baseContract = {
+            // id: `log_${rawContract.contractName}_${rawContract.networkName}`,
+            contractName: rawContract.contractName,
+            networkName: rawContract.networkName,
+            chainId: network.chainId,
+            abi: rawContract.abi,
+            // abiEvents: abiEvents,
```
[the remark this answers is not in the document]
kyscott18: Whoops

packages/core/src/sync-historical/service.ts
```diff
@@ -567,10 +716,6 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
     clearInterval(this.progressLogInterval);
     this.queue.pause();
     this.queue.clear();
-    this.common.logger.debug({
-      service: "historical",
-      msg: `Killed historical sync service (network=${this.network.name})`,
-    });
```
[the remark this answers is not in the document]
kyscott18: No good reason, just noticed that we didn't have it on any of the other services

packages/core/src/sync-historical/service.ts
```diff
@@ -1013,15 +1228,165 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
...
+
+    for (const trace of persistentTraces) {
+      const blockNumber = hexToNumber(trace.blockNumber);
+
+      if (tracesByBlockNumber[blockNumber] === undefined) {
+        tracesByBlockNumber[blockNumber] = [];
+      }
+      if (txHashesByBlockNumber[blockNumber] === undefined) {
+        txHashesByBlockNumber[blockNumber] = new Set<Hash>();
+      }
```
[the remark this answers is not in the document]
kyscott18: image

packages/core/src/sync-realtime/service.ts
```diff
@@ -331,77 +336,147 @@ export const handleBlock = async (
...
+      throw new Error(
+        `Received call trace with block hash '${callTrace.blockHash}' that does not match current head block '${newHeadBlock.hash}'`,
+      );
+    }
+  }
+
+  // Protect against RPCs returning empty traces. Known to happen near chain tip.
+  // Use the fact that any stateRoot change produces a trace.
+  if (
+    newHeadBlock.stateRoot !== getLatestLocalBlock(service).stateRoot &&
```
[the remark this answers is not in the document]
kyscott18: `blockTraces` actually contains all trace, not just call traces. I think there might be a situation where the state root changes but there is no transactions (block rewards given out) but I'm not confident. This logic is definitely less fool proof than the "logs" version.

packages/core/src/sync/events.ts
kyscott18: This guards against us getting a random value for topic0 back from the rpc. This doesn't actually change the behavior of this function, it just more explicitly throws an error here rather than accessing a property on undefined

packages/core/src/sync-historical/service.ts
```diff
@@ -1013,15 +1228,165 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
...
+            startBlock: BigInt(traceInterval.startBlock),
+            endBlock: BigInt(traceInterval.endBlock),
+          },
+          traceFilter: traceFilter.criteria,
+          chainId: traceFilter.chainId,
+          block,
+          transactions,
+          transactionReceipts: transactionReceipts.filter((txr) =>
+            transactionHashes.has(txr.transactionHash),
+          ),
```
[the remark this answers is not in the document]
kyscott18: No we don't because all `traceIntervals` include transaction receipts regardless of the config passed by the user.

packages/core/src/sync-realtime/service.ts
```diff
@@ -331,77 +336,147 @@ export const handleBlock = async (
...
+  const blockCallTraces = blockTraces.filter(
+    (trace) => trace.type === "call",
+  ) as SyncCallTrace[];
+  const newCallTraces = filterCallTraces({
+    callTraces: blockCallTraces,
+    callTraceFilters: service.callTraceSources.map((s) => s.criteria),
+  });
+
+  // Check that traces refer to the correct block
+  for (const callTrace of newCallTraces) {
```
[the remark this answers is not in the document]
kyscott18: Been somewhat scarred by the RPC responses in the past so I would like to air on the side of safety for this one until we can see that it has a tangible negative performance effect

PR #882, opened 2024-05-13
Factory trace filters

kyscott18:


packages/core/src/sync-realtime/service.ts
```diff
@@ -710,7 +716,74 @@ const getMatchedLogs = async (
       logs,
       logFilters: [
         ...service.logFilterSources.map((l) => l.criteria),
-        ...factoryLogFilters,
+        ...factoryLogFilters.filter((f) => f.address.length !== 0),
```
kyscott18, line 719: I think I actually caught a subtle bug in the `getMatchedLogs` function here that would cause us to match on every "address" in the case when there were no child addresses in the database.

packages/core/src/sync-store/sqlite/store.ts
```diff
@@ -899,6 +903,244 @@ export class SqliteSyncStore implements SyncStore {
...
+    transactionReceipts: rpcTransactionReceipts,
+    traces: rpcTraces,
+    interval,
+  }: {
+    chainId: number;
+    factory: FactoryCallTraceFilterCriteria;
+    block: RpcBlock;
+    transactions: RpcTransaction[];
+    transactionReceipts: RpcTransactionReceipt[];
+    traces: SyncCallTrace[];
```
0xOlias, line 920: Any reason this isn't called `RpcCallTrace`?

packages/core/src/sync-store/sqlite/store.ts
```diff
@@ -899,6 +903,244 @@ export class SqliteSyncStore implements SyncStore {
...
+              .insertInto("transactionReceipts")
+              .values(transactionReceipts)
+              .onConflict((oc) => oc.column("transactionHash").doNothing())
+              .execute();
+          }
+
+          // Delete existing traces with the same `transactionHash`. Then, calculate "callTraces.checkpoint"
+          // based on the ordering of "callTraces.traceAddress" and add all traces to "callTraces" table.
+          const traceByTransactionHash: {
+            [transacionHash: Hex]: SyncCallTrace[];
```
0xOlias, line 966: Typo

packages/core/src/sync-store/sqlite/store.ts
```diff
@@ -1373,21 +1665,22 @@ export class SqliteSyncStore implements SyncStore {
...
-    const logOrFactorySources = sources.filter(
-      (s): s is LogSource | FactorySource =>
-        sourceIsLog(s) || sourceIsFactory(s),
+    const logSources = sources.filter(
+      (s): s is LogSource | FactoryLogSource =>
+        sourceIsLog(s) || sourceIsFactoryLog(s),
+    );
+    const callTraceSources = sources.filter(
+      (s): s is CallTraceSource | FactoryCallTraceSource =>
+        sourceIsCallTrace(s) || sourceIsFactoryCallTrace(s),
```
0xOlias, line 1674: Nice, this makes a lot of sense now

packages/core/src/sync-store/sqlite/store.ts
```diff
@@ -1373,21 +1665,22 @@ export class SqliteSyncStore implements SyncStore {
...
+    const shouldJoinLogs = logSources.length !== 0;
     const shouldJoinTransactions =
-      logOrFactorySources.length !== 0 || callTraceSources.length !== 0;
+      logSources.length !== 0 || callTraceSources.length !== 0;
     const shouldJoinTraces = callTraceSources.length !== 0;
     const shouldJoinReceipts =
-      logOrFactorySources.some(
-        (source) => source.criteria.includeTransactionReceipts,
-      ) ||
+      logSources.some((source) => source.criteria.includeTransactionReceipts) ||
```
0xOlias, line 1683: `shouldJoinReceipts` will always be true if there is a call trace source, right? Because `callTraceSource.includeTransactionReceipts` is hard-coded to `true`?

packages/core/src/sync-realtime/service.ts
```diff
@@ -338,16 +341,18 @@ export const handleBlock = async (
 
   const shouldRequestLogs =
     positiveBloomFilter &&
-    (service.logFilterSources.length > 0 || service.factorySources.length > 0);
-  const shouldRequestTraces = service.callTraceSources.length > 0;
+    (service.logFilterSources.length > 0 ||
+      service.factoryLogSources.length > 0);
```
0xOlias, line 345: Correct me if I'm wrong but I think this is redundant, and we can simply rename positiveBloomFilter to `shouldRequestLogs`?

packages/core/src/sync-realtime/service.ts
```diff
@@ -108,9 +110,10 @@ export const create = ({
   onFatalError: (error: Error) => void;
 }): Service => {
   const logFilterSources = sources.filter(sourceIsLog);
```
0xOlias, line 112: Should this be renamed to `logSources` for consistency?

0xOlias (approved):

packages/core/src/sync-store/sqlite/store.ts
kyscott18: The `Rpc` prefixed types are all from viem

PR #887, opened 2024-05-15
`p.json()`

kyscott18:


packages/core/src/indexing-store/utils/sort.ts
```diff
@@ -42,7 +43,8 @@ export function buildOrderByConditions({
           (columnName) =>
             isScalarColumn(table[columnName]) ||
             isReferenceColumn(table[columnName]) ||
-            isEnumColumn(table[columnName]),
+            isEnumColumn(table[columnName]) ||
+            isJSONColumn(table[column]),
```
0xOlias, line 47: I think we probably want to disallow sorting by JSON columns (just like filtering).

packages/core/src/database/postgres/service.ts
```diff
@@ -730,6 +731,12 @@ export class PostgresDatabaseService implements BaseDatabaseService {
           if (isOptionalColumn(column) === false) col = col.notNull();
           return col;
         });
+      } else if (isJSONColumn(column)) {
+        // Handle json columns
+        builder = builder.addColumn(columnName, "text", (col) => {
```
0xOlias, line 736: I think we should use `JSONB` as the column type for both Postgres and SQLite.

To ease the bigint pain, we could consider automatically converting all bigints in the object to strings on insertion (in the `encodeValue` function). However, that opens at least one can of worms - to ensure that `create` and `findUnique` behave as expected (you get the same object you put in), we would need to know which values were bigints and got stringified. There is prior art for this (usually convert the bigint into an object like `{ __bigint: true, value: "10000000000000000" }` such that you can decode it confidently), but it's not pretty, and makes the database content ugly.

So, my instinct is to disallow bigints (implicitly) by letting the SQL error bubble up to the user on insertion. We could of course catch that error and offer a more useful string like we do for all the constraints.

0xOlias (commented):
Comments for discussion

packages/core/src/database/postgres/service.ts
kristofgazso: `jsonb` is goated, gotta make way for the king

PR #892, opened 2024-05-16
Improve RPC and database retry behavior

kyscott18:


packages/core/src/indexing-store/utils/errors.ts
```diff
@@ -1,4 +1,5 @@
 import {
+  ConstraintCheckError,
```
0xOlias, line 2: Should be `CheckConstraintError`

https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS

packages/core/src/indexing-store/utils/errors.ts
```diff
@@ -26,6 +27,11 @@ export function parseStoreError(err: unknown, args: Record<string, unknown>) {
     error.message?.includes("violates not-null constraint")
   ) {
     error = new NotNullConstraintError(error.message);
+  } else if (
+    error.message?.includes("CHECK constraint failed") ||
+    error.message?.includes("violates check constraint")
+  ) {
```
0xOlias, line 33: Is it possible to add a test for this?

packages/core/src/utils/requestQueue.ts
```diff
@@ -105,9 +109,30 @@ export const createRequestQueue = ({
...
+          });
+          throw error;
+        }
+
+        if (i === RETRY_COUNT) {
+          common.logger.warn({
+            service: "sync",
+            msg: `Failed '${request.method}' RPC request after ${
+              i + 1
+            } attempts with error: ${error}`,
```
0xOlias, line 125: Let's be thoughtful about how to include errors within logs. 

I believe this will call `toString` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/toString#description) on the error object, but elsewhere we do `Error: ${error.message}`, and other times we omit the `with error` or `error:` entirely.

Not a solution here, just drawing attention.

packages/core/src/utils/requestQueue.ts
```diff
@@ -154,6 +179,7 @@ export const createRequestQueue = ({
 function shouldRetry(error: Error) {
   if ("code" in error && typeof error.code === "number") {
     if (error.code === -1) return true; // Unknown error
+    if (error.code === InvalidInputRpcError.code) return true;
```
0xOlias, line 182: Might be worth adding to the changeset.

Fixed a bug where certain RPC errors were incorrectly marked as non-retryable.

packages/core/src/database/sqlite/service.ts
```diff
@@ -220,7 +220,7 @@ export class SqliteDatabaseService implements BaseDatabaseService {
               } catch (err) {
                 const error = err as Error;
                 if (!error.message.includes("already exists")) throw error;
-                throw new Error(
+                throw new NonRetryableError(
```
0xOlias, line 223: Nice catch

packages/core/src/indexing-store/utils/errors.ts
kyscott18: We have a test called "create() throws on invalid enum value"

0xOlias: Oh I see, I suppose what I'm suggesting is to add `expect(error).instanceOf(CheckConstraintError);` to that test, to confirm that the error strings we're checking here are correct. Also this way, if we upgraded our SQLite/PG driver and those strings changed, we'd catch it.

packages/core/src/utils/requestQueue.ts
kyscott18: We should use `.message` in most places (wherever we don't want the stack trace)

packages/core/src/database/sqlite/service.test.ts
```diff
@@ -336,7 +336,7 @@ describe.skipIf(shouldSkip)("sqlite database", () => {
     await databaseTwo.kill();
   });
 
-  test("setup throws if there is a table name collision", async (context) => {
+  test.only("setup throws if there is a table name collision", async (context) => {
```
0xOlias, line 339: rm

packages/core/src/indexing-store/historical.test.ts
```diff
@@ -138,14 +141,16 @@ test("create() throws on invalid enum value", async (context) => {
...
       tableName: "Pet",
       encodedCheckpoint: encodeCheckpoint(createCheckpoint(10)),
       id: "id1",
       data: { name: "Skip", kind: "NOTACAT" },
-    }),
-  ).rejects.toThrow();
+    })
+    .catch((error) => error);
+
+  expect(error).toBeInstanceOf(CheckConstraintError);
```
0xOlias, line 153: Nice 🫡

0xOlias (approved):
One `.only` to remove but otherwise LGTM!

0xOlias (approved):

PR #893, opened 2024-05-16
`p.json`

kyscott18:


docs/pages/docs/schema.mdx
```diff
@@ -91,12 +94,18 @@ await Account.create({
...
+      middleName: "james"
+    }
   },
 });
 ```
 
+
 </div>
 
+`p.json(){:ts}` also accepts an optional generic type, which can be used to acheived end to end type safety.
```
kyscott18, line 107: Needs work

docs/pages/docs/schema.mdx
```diff
@@ -91,12 +94,18 @@ await Account.create({
     lastActiveAt: 1679337733,
     isAdmin: true,
     graffiti: "LGTM",
+    json: {
+      middleName: "james"
```
kristofgazso, line 98: is your middle name actually james?

kyscott18: yes

packages/core/src/indexing-store/utils/encoding.ts
```diff
@@ -208,6 +220,8 @@ function decodeValue(
       return JSON.parse(value as string);
     }
     return value as UserColumn;
+  } else if (isJSONColumn(column)) {
+    return encoding === "postgres" ? value : JSON.parse(value as string);
```
0xOlias, line 224: Initially thought we might want a try/catch here as well, but I suppose there's a very low risk of an error with this `JSON.parse`, because we already stringified it and managed to insert + select it out of the database.

packages/core/src/indexing-store/utils/errors.ts
```diff
@@ -26,6 +27,8 @@ export function parseStoreError(err: unknown, args: Record<string, unknown>) {
     error.message?.includes("violates not-null constraint")
   ) {
     error = new NotNullConstraintError(error.message);
+  } else if (error.message?.includes("Do not know how to serialize a BigInt")) {
+    error = new JSONSerializeError(error.message);
```
0xOlias, line 31: New contender for most common bug report?

0xOlias (approved):
Awesome work!

PR #898, opened 2024-05-17
Fixes duplicate index creation error

kyscott18:


packages/core/src/database/postgres/service.ts
```diff
@@ -307,6 +307,29 @@ export class PostgresDatabaseService implements BaseDatabaseService {
               )} ago`,
             });
 
+            // Remove any indexes, will be recreated once the app
+            // becomes healthy.
+            await Promise.all(
```
0xOlias, line 312: A nested for loop might be more readable here (in a transaction so no concurrency benefit).

packages/core/src/database/sqlite/service.ts
```diff
@@ -280,6 +280,29 @@ export class SqliteDatabaseService implements BaseDatabaseService {
...
+            // becomes healthy.
+            await Promise.all(
+              Object.entries(getTables(schema)).flatMap(
+                ([tableName, table]) => {
+                  if (table.constraints === undefined) return [];
+
+                  return Object.keys(table.constraints).map(async (name) => {
+                    await tx
+                      .withSchema(this.userNamespace)
+                      .schema.dropIndex(`${tableName}_${name}`)
```
0xOlias, line 293: Not sure it matters, but the order of `.schema` and `.withSchema` is different here than in the postgres service.

0xOlias (approved):
LGTM, nits

PR #903, opened 2024-05-20
update many pagination

kyscott18:


0xOlias (approved):
Do we have a test for `updateMany` with `> MAX_BATCH_SIZE` rows? Other than that, LGTM

packages/core/src/indexing-store/historical.ts
```diff
@@ -142,51 +147,73 @@ export const getHistoricalStore = ({
...
+        if (_rows.length === 0) {
+          break;
+        } else {
+          cursor = encodeValue(_rows[_rows.length - 1].id, table.id, kind);
+        }
+      }
+
+      return rows;
+    } else {
+      return db.wrap({ method: `${tableName}.updateMany` }, async () => {
```
0xOlias, line 216: Is it OK that the path above is async/await and this one returns the promise?

packages/core/src/indexing-store/realtime.ts
```diff
@@ -188,64 +193,84 @@ export const getRealtimeStore = ({
...
-              throw parseStoreError(err, updateObject);
-            });
+    const rows: UserRecord[] = [];
+    let cursor: DatabaseColumn = null;
+
+    while (true) {
+      const _rows = await db.wrap(
+        { method: `${tableName}.updateMany` },
+        async () => {
+          return await db.transaction().execute(async (tx) => {
```
0xOlias, line 203: Might be able to remove a layer here

0xOlias (approved):

packages/core/src/indexing-store/historical.ts
kyscott18: yes

PR #917, opened 2024-05-28
Support future end block

### LINKED ISSUE #813: [Feature] Support future `endBlock`
Currently, ponder doesn't allow `endBlock` to be greater than the latest block.
It would be nice if in such case, it would allow it and stop indexing the concerned contract/events when that block is reached.

kyscott18:
Fixes several bugs pertaining to unfinalized start and end blocks

closes #813

### DIFF
--- a/packages/core/src/sync-historical/service.ts
+++ b/packages/core/src/sync-historical/service.ts
@@ -211,10 +211,6 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
         switch (source.type) {
           case "log": {
             if (!isHistoricalSyncRequired) {
-              this.logFilterProgressTrackers[source.id] = new ProgressTracker({
-                target: [startBlock, finalizedBlockNumber],
-                completed: [[startBlock, finalizedBlockNumber]],
-              });
               this.common.metrics.ponder_historical_total_blocks.set(
                 {
                   network: this.network.name,
@@ -304,16 +300,6 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
 
           case "factoryLog": {
             if (!isHistoricalSyncRequired) {
-              this.factoryChildAddressProgressTrackers[source.id] =
-                new ProgressTracker({
-                  target: [startBlock, finalizedBlockNumber],
-                  completed: [[startBlock, finalizedBlockNumber]],
-                });
-              this.factoryLogFilterProgressTrackers[source.id] =
-                new ProgressTracker({
-                  target: [startBlock, finalizedBlockNumber],
-                  completed: [[startBlock, finalizedBlockNumber]],
-                });
               this.common.metrics.ponder_historical_total_blocks.set(
                 {
                   network: this.network.name,
@@ -487,12 +473,6 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
 
           case "callTrace": {
             if (!isHistoricalSyncRequired) {
-              this.traceFilterProgressTrackers[source.id] = new ProgressTracker(
-                {
-                  target: [startBlock, finalizedBlockNumber],
-                  completed: [[startBlock, finalizedBlockNumber]],
-                },
-              );
               this.common.metrics.ponder_historical_total_blocks.set(
                 {
                   network: this.network.name,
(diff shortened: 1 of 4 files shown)


packages/core/src/sync-historical/service.ts
```diff
@@ -201,20 +199,15 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
...
             finalizedBlockNumber,
-            latestBlockNumber,
           });
 
         switch (source.type) {
           case "log": {
             if (!isHistoricalSyncRequired) {
-              this.logFilterProgressTrackers[source.id] = new ProgressTracker({
-                target: [startBlock, finalizedBlockNumber],
-                completed: [[startBlock, finalizedBlockNumber]],
```
kyscott18, line 216: This is what was causing the original issue with unfinalized start blocks

packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -342,6 +342,10 @@ export async function buildConfigAndIndexingFunctions({
         ? undefined
         : endBlockMaybeNan;
 
+      if (endBlock !== undefined && endBlock < startBlock) {
```
0xOlias, line 345: Nice

packages/core/src/sync-historical/service.ts
```diff
@@ -201,20 +199,15 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
     await Promise.all(
       this.sources.map(async (source) => {
         const { isHistoricalSyncRequired, startBlock, endBlock } =
-          validateHistoricalBlockRange({
+          getHistoricalBlockRange({
```
0xOlias, line 202: Inline this?

packages/core/src/sync/service.ts
```diff
@@ -89,9 +90,15 @@ export const create = async ({
...
       case "checkpoint": {
-        syncService.networkServices.find(
+        const networkService = syncService.networkServices.find(
           (ns) => ns.network.chainId === realtimeSyncEvent.chainId,
-        )!.realtime!.checkpoint = realtimeSyncEvent.checkpoint;
+        )!;
+
+        // "realtime" property may be undefined when `kill()` has been
+        // invoked but hasn't completed.
+        if (networkService.realtime === undefined) return;
```
0xOlias, line 99: Would it be simpler to just use `?.` instead of `!.` above?

packages/core/src/sync/service.ts
```diff
@@ -142,9 +155,28 @@ export const create = async ({
...
+
+        // Check if the finalized blockNumber is greater than the end block of all
+        // sources for the network. Potentially kill the realtime sync and remove the
+        // network from checkpoint calculations.
+        if (
+          networkService.realtime.endBlock !== undefined &&
+          realtimeSyncEvent.checkpoint.blockNumber >
+            networkService.realtime.endBlock
+        ) {
+          networkService.realtime.realtimeSync.kill();
```
0xOlias, line 177: I think we should add an info log here: `Synced final end block for '${network.name}' (${networkService.realtime.endBlock}), killing realtime sync service`

packages/core/src/sync/service.ts
kyscott18: There's actually two different things that can be undefined. `networkService` can be undefined depending on the result of `.find()`. However, it's an invariant that it will always be defined, which is why the `!` is used. 

The `realtime` property could be undefined if the case described in the comment occurs.

packages/core/src/sync-historical/service.ts
```diff
@@ -198,12 +197,10 @@ export class HistoricalSyncService extends Emittery<HistoricalSyncEvents> {
...
       this.sources.map(async (source) => {
-        const { isHistoricalSyncRequired, startBlock, endBlock } =
-          getHistoricalBlockRange({
-            startBlock: source.startBlock,
-            endBlock: source.endBlock,
-            finalizedBlockNumber,
-          });
+        const startBlock = source.startBlock;
+        const endBlock = source.endBlock ?? finalizedBlockNumber;
+        const isHistoricalSyncRequired =
```
0xOlias, line 202: We might be able to go one step further and consolidate the `if (!isHistoricalSyncRequired) { ...` logic from each branch below

0xOlias (approved):
One more nit but LGTM!

0xOlias (approved):

PR #921, opened 2024-05-31
Skip reindexing on redeploy

kyscott18:


packages/core/src/database/postgres/service.ts
```diff
@@ -332,14 +505,15 @@ export class PostgresDatabaseService implements BaseDatabaseService {
               .withSchema(this.internalNamespace)
               .updateTable("namespace_lock")
               .set({ is_locked: 1, heartbeat_at: Date.now() })
+              .where("namespace", "=", this.userNamespace)
```
kyscott18, line 508: @0xOlias Was this an existing bug?

PR #929, opened 2024-06-04
Historical indexing store with in-memory cache

kyscott18:


packages/core/src/indexing-store/utils/encoding.ts
```diff
@@ -38,14 +47,130 @@ const scalarToTsType = {
...
+  table,
+  schema,
+  encoding,
+  skipValidate,
+}: {
+  record: Partial<UserRecord>;
+  table: Table;
+  schema: Schema;
+  encoding: "sqlite" | "postgres";
+  skipValidate: boolean;
```
0xOlias, line 61: `skipValidation`

packages/core/src/indexing-store/utils/encoding.ts
```diff
@@ -63,140 +188,179 @@ export function encodeRow(
...
+function validateField({
+  value,
+  column,
+  schema,
+  skipValidate,
+}: {
+  value: UserColumn;
+  column: Column;
+  schema: Schema;
+  skipValidate: boolean;
```
0xOlias, line 204: After reading through this function - should this option be called `skipConstraintValidation` or `skipDatabaseValidation`?

It seems like this function _always_ validates encoding issues - it's not clear to me which category of validations get skipped when `skipValidate` is true.

packages/core/src/indexing-store/readonly.ts
```diff
@@ -116,7 +120,9 @@ export const getReadonlyStore = ({
       if (after === null && before === null) {
         query = query.limit(limit + 1);
         const rows = await query.execute();
-        const records = rows.map((row) => decodeRow(row, table, kind));
+        const records = rows.map((row) =>
+          decodeRecord({ record: row, table, encoding }),
+        );
```
0xOlias, line 125: I wonder if there's an opportunity to always use "row" for database-compatible objects (either just before inserting or just after selecting) and "record" for user-land objects. If yes, I think it would make sense for this to look like `decodeRow({ row, table, encoding })`.

packages/core/src/indexing-store/historical.ts
```diff
@@ -1,363 +1,893 @@
...
+    db,
+  });
+
+  common.logger.debug({
+    service: "indexing",
+    msg: `Using a ${Math.round(maxSizeBytes / (1024 * 1024))} mB cache.`,
+  });
+
+  /** True if the cache contains the complete state of the store. */
+  let isCacheFull = _isCacheFull;
```
0xOlias, line 82: `isCacheComplete` may be better - "full" makes it seem like it's full and needs to be flushed.

packages/core/src/indexing-store/historical.ts
```diff
@@ -1,363 +1,893 @@
...
+      find: {},
+    };
+  }
+
+  const getCacheId = (id: UserId, tableName: string): string | number => {
+    if (tables[tableName].table.id[" scalar"] === "hex")
+      return padHex(id as Hex, {
+        size: Math.ceil(((id as Hex).length - 2) / 2),
+        dir: "left",
+      }).toLowerCase();
```
0xOlias, line 104: What's going on here?

packages/core/src/indexing-store/historical.ts
```diff
@@ -1,363 +1,893 @@
...
-          .returningAll()
-          .execute()
-          .catch((err) => {
-            throw parseStoreError(err, data.length > 0 ? data[0] : {});
-          });
+  };
+
+  const shouldFlush = () => cacheSizeBytes > maxSizeBytes;
+
+  const flush = createQueue<void, boolean | void>({
```
0xOlias, line 144: What's the purpose of the queue?

packages/core/src/indexing-store/historical.ts
```diff
@@ -1,363 +1,893 @@
...
+      id: UserId;
+      data?: Omit<UserRecord, "id">;
+    }) => {
+      const encodedId = getCacheId(id, tableName);
+
+      if (
+        storeCache[tableName].insert[encodedId] !== undefined ||
+        storeCache[tableName].update[encodedId] !== undefined ||
+        (storeCache[tableName].find[encodedId] !== undefined &&
+          storeCache[tableName].find[encodedId].record !== undefined)
```
0xOlias, line 409: Could be this?
```
storeCache[tableName].find[encodedId]?.record !== undefined
```

packages/core/src/indexing-store/historical.ts
```diff
@@ -1,363 +1,893 @@
...
+        const bytes = insertEntry.bytes;
+        delete storeCache[tableName].insert[encodedId];
+
+        cacheSize--;
+        cacheSizeBytes -= bytes;
+
+        return true;
+      } else if (isCacheFull) {
+        return false;
+      } else if (findEntry !== undefined) {
```
0xOlias, line 822: Are you sure this belongs above the `if (findEntry !== undefined)`?

packages/core/src/indexing-store/historical.ts
```diff
@@ -1,363 +1,893 @@
...
+    } else if (typeof col === "boolean") {
+      // p.boolean
+      size += 4;
+    } else if (typeof col === "bigint") {
+      // p.bigint
+      size += 48;
+    } else if (Array.isArray(col)) {
+      // p.list
+      for (const e of col) {
+        size += getRecordSize(e);
```
0xOlias, line 881: Doesn't this assume that `col` is an array of objects, because it will hit line 865 above? What happens if it's an array of primitives? Same story for the `p.json()` branch.

0xOlias (commented):
Looking good!

packages/core/src/indexing-store/historical.ts
kyscott18: All of the cases are equivalent, because of the bytes encoding we are doing in the db: "0xa", "0xA", "0x0a", "0x0A"

packages/core/src/indexing-store/historical.ts
kyscott18: Nice

packages/core/src/indexing-store/historical.ts
kyscott18: To prevent race conditions from user-land concurrency

packages/core/src/indexing-store/utils/encoding.ts
kyscott18: `skipValidate` is meant to mean "skip any extra validation that would be caught by the database". Because of the very small overhead of validating in js runtime compared to a db query, I decided to never skip validation and remove the parameter.

0xOlias (approved):

packages/core/src/indexing-store/historical.ts
```diff
@@ -334,36 +366,38 @@ export const getHistoricalStore = ({
       tableName: string;
       id: UserId;
     }) => {
-      const id = structuredClone(_id);
-      const cacheKey = getCacheKey(id, tableName);
+      if (shouldFlush()) await flush({ isFullFlush: false });
```
0xOlias, line 369: Is there a reason you moved this to the top of the function in this last commit?

0xOlias (approved):

packages/core/src/indexing-store/historical.ts
kyscott18: I think it should be outside of the `wrap`

packages/core/src/indexing-store/historical.ts
```diff
@@ -1,363 +1,880 @@
...
+  common: Common;
+  isCacheExhaustive: boolean;
+}): HistoricalStore => {
+  const maxSizeBytes = common.options.indexingCacheBytes;
+  const storeCache: StoreCache = {};
+  const tables = getTables(schema);
+
+  common.logger.debug({
+    service: "indexing",
+    msg: `Using a ${Math.round(maxSizeBytes / (1024 * 1024))} mB cache.`,
```
0xOlias, line 104: `Using a ${Math.round(maxSizeBytes / (1024 * 1024))} MB indexing cache`

0xOlias (approved):

PR #931, opened 2024-06-10
Batch mergeIntervals subquery to avoid timeout

kyscott18:


packages/core/src/sync-store/postgres/store.ts
```diff
@@ -161,24 +164,32 @@ export class PostgresSyncStore implements SyncStore {
...
+          while (true) {
             const { id: logFilterId } = await tx
               .insertInto("logFilters")
               .values(fragment)
               .onConflict((oc) => oc.column("id").doUpdateSet(fragment))
               .returningAll()
               .executeTakeFirstOrThrow();
 
-            const existingIntervalRows = await tx
+            const existingIntervals = await tx
```
0xOlias, line 177: Might be worth adding a comment here, "This is a trick to add a LIMIT to a DELETE statement"

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -198,9 +209,18 @@ export class PostgresSyncStore implements SyncStore {
...
-          });
-        }),
-      );
+
+            if (existingIntervals.length !== MAX_MERGE_INTERVALS) break;
+
+            if (mergedIntervalRows.length === 0) {
+              // This occurs when there are too many non-mergeable ranges with the same logFilterId. Should be almost impossible.
+              throw new NonRetryableError(
+                `'logFilterIntervals' table for chain '${chainId}' has reached an unrecoverable level of fragmentation. Wiping your database is recommended.`,
```
0xOlias, line 218: Would omit the 2nd sentence. If this happens, I think ideally the user would report it so we can investigate.

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -198,9 +209,18 @@ export class PostgresSyncStore implements SyncStore {
                 .values(mergedIntervalRows)
                 .execute();
             }
-          });
-        }),
-      );
+
+            if (existingIntervals.length !== MAX_MERGE_INTERVALS) break;
+
+            if (mergedIntervalRows.length === 0) {
```
0xOlias, line 215: Double check these exit conditions

0xOlias (approved):
LGTM

PR #932, opened 2024-06-10
fix timer rounding

kyscott18:


packages/core/src/utils/timer.ts
```diff
@@ -12,5 +12,6 @@ export function startClock() {
  * @returns The timestamp in milliseconds (ms).
  */
 export function hrTimeToMs(diff: [number, number]) {
-  return Math.round(diff[0] * 1000 + diff[1] / 1_000) / 1_000;
+  const ns = diff[0] * 10 ** 9 + diff[1];
+  return ns / 10 ** 6;
```
0xOlias, line 16: So, this returns milliseconds including micro/nanoseconds after the decimal place, eg `12.928472` or `0.001234` ?

kyscott18: yes, up to six decimals places, 1 nanosecond granularity

0xOlias (approved):

PR #936, opened 2024-06-12
getEvents performance improvements

kyscott18:


packages/core/src/common/options.ts
```diff
@@ -35,6 +35,7 @@ export type Options = {
   indexingCacheBytes: number;
 
   syncMaxIntervals: number;
+  syncEventsQuerySize: number;
```
0xOlias, line 38: Naming ideas:
```
indexingCacheMaxBytes
indexingCacheFlushRatio

syncStoreMaxDistinctIntervals
syncStoreMaxBatchSize
```

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -2202,114 +2215,35 @@ export class PostgresSyncStore implements SyncStore {
...
       });
 
-      const hasNextPage = events.length === limit + 1;
-
-      if (!hasNextPage) {
-        yield events;
-        break;
+      // set fromCursor + seconds
+      if (events.length === 0) {
+        this.seconds = Math.round(this.seconds * 2);
```
0xOlias, line 2220: Could consider making these constants explicit options, but fine to skip for now I think.

packages/core/src/sync-store/sqlite/store.ts
```diff
@@ -2059,10 +2072,10 @@ export class SqliteSyncStore implements SyncStore {
                 "transactionReceipts.type as txr_type",
               ]),
           )
-          .where("events.checkpoint", ">", cursor)
-          .where("events.checkpoint", "<=", encodedToCheckpoint)
+          .where("events.checkpoint", ">", fromCursor)
+          .where("events.checkpoint", "<=", toCursor)
```
0xOlias, line 2076: Worth double-checking the exclusivity rules here again (I think we're good though).

packages/core/src/sync-store/store.test.ts
```diff
@@ -2661,7 +2662,7 @@ test("getEvents filters on toCheckpoint (inclusive)", async (context) => {
 
   const ag = syncStore.getEvents({
     sources: [sources[0]],
-    fromCheckpoint: zeroCheckpoint,
+    fromCheckpoint: createBlockCheckpoint(rpcData.block2.block, false),
```
0xOlias, line 2665: For my understanding: Now, it's no longer OK/safe to pass zeroCheckpoint as fromCheckpoint?

packages/core/src/sync/service.ts
```diff
@@ -30,6 +32,13 @@ export type Service = {
 
   // state
   checkpoint: Checkpoint;
+  /** Checkpoint of the earliest start block. */
```
0xOlias, line 35: Great comments here

packages/core/src/sync/service.ts
```diff
@@ -279,6 +316,20 @@ export const create = async ({
           sources: networkSources,
           requestQueue,
           cachedTransport: cachedTransport({ requestQueue, syncStore }),
+          startCheckpoint: {
```
0xOlias, line 319: Could consider evaluating these two above, like initialFinalizedCheckpoint

0xOlias (approved):
All nits - looks great

packages/core/src/sync-store/postgres/store.ts
```diff
@@ -67,12 +67,16 @@ export class PostgresSyncStore implements SyncStore {
...
+  private seconds: number;
+
   constructor({
     db,
     common,
   }: { db: HeadlessKysely<SyncStoreTables>; common: Common }) {
     this.db = db;
     this.common = common;
+
+    this.seconds = common.options.syncEventsQuerySize * 2;
```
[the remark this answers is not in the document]
kyscott18: It's helpful to be an instance variable for the "cold" cases, when sync is happening at the same time as indexing

packages/core/src/sync-store/store.test.ts
kyscott18: it is fine, just does a ton of extra queries

PR #939, opened 2024-06-15
Fixes bug in historical indexing update cache

kyscott18:
The "update" branch of the flush function uses `onConflict` to overwrite existing database rows, but when there are no rows besides "id", nothing needs to be updated. Previously this code path was causing an SQL syntax error, now it just exits early.


0xOlias (approved):
LGTM - I assume that test was failing before the change?

kyscott18:
> LGTM - I assume that test was failing before the change?

yes

PR #943, opened 2024-06-17
Custom API Functions

### LINKED ISSUE #514: Custom Server Alternative to GraphQL
With the completion of #370, it becomes possible for Ponder to support custom load-time logic. This would work in parallel or completely replace GraphQL to serve incoming requests. Ponder could completely internalize a web server, and define its own api for serving requests:

```ts
import { app } from "@/generated";

app.get("/players", async ({request: Request, context}): Response => {
  const players = await context.db.Player.findMany({
    where: {
      id: {
        startsWith: "J",
      },
    },
  });
  
  // Do an rpc request, perform computation on players, etc

  return new Response(players);
});
```

or support any arbitrary framework, which Ponder would be in charge of running. For example, `server/index.ts` could be the default entrypoint, which would contain a file like:

```ts
import { serve } from "@hono/node-server";
import { Hono } from "hono";
import { context } from "@/generated";

const app = new Hono();

app.get("/players", async (c) => {
  const players = await context.db.Player.findMany({
    where: {
      id: {
        startsWith: "J",
      },
    },
  });

  // Do an rpc request, perform computation on players, etc
  
  return c.json(players);
});

serve(app);
```

The benefits to this solution is supporting maximum flexibility. The environment would not be as sandboxed which would potentially created harder to recover from errors.

Both of these solutions would also allow for queries directly against the db. We need to determine a methodology for supporting a strongly typed ORM.

### LINKED ISSUE #561: Turn off GraphQL server
Sort of piggybacking on [Issue 514](https://github.com/ponder-sh/ponder/issues/514) I think it might be helpful to turn off the GraphQL server completely for certain people.

Currently I'm using Ponder and Reservoir Light Sync Nodes to dump everything into a Supabase database, then using its built-in GraphQL server so I can query the union of data.

In this setup, the GraphQL server for Ponder doesn't add much benefit for me, and mostly only overhead.

kyscott18:
closes #514, closes #561


docs/pages/docs/query/api-functions.mdx
```diff
@@ -0,0 +1,136 @@
+---
+title: "API functions"
+description: "Use API functions to customize the API layer of your app."
+---
+
+import { Callout, Steps } from "nextra/components";
+
+# API functions (experimental)
+
+**API functions** are user-defined functions that respond to incoming Web requests. You can use them to customize the API layer of your app with direct SQL queries, authentication, data from external sources, and more.
```
kyscott18, line 10: Is "Web" supposed to be capitalized

gskril:
quick note that mentions of `/health` and `/metrics` in the docs should probs be updated to `_ponder/health` and `_ponder/metrics`. tripped me up for a sec 😅

packages/core/src/build/service.ts
```diff
@@ -278,6 +320,80 @@ export const start = async (
...
+      buildService.viteNodeRunner.moduleCache.invalidateDepTree(files);
+
+      common.logger.info({
+        service: "build",
+        msg: `Hot reload ${files
+          .map((f) => `'${path.relative(common.options.rootDir, f)}'`)
+          .join(", ")}`,
+      });
+
+      buildService.viteNodeRunner.moduleCache.deleteByModuleId("@/generated");
```
0xOlias, line 379: Could these invalidations happen within `executeServer`? I think I'd find that easier to follow.

packages/core/src/build/service.ts
```diff
@@ -31,6 +34,9 @@ export type Service = {
   // static
   common: Common;
   srcRegex: RegExp;
+  serverRegex: RegExp;
```
0xOlias, line 37: I think it might be better to consistently use "api" instead of "server" when referring to API functions.

Also after reading this file, I wonder if it would be easier to have an `indexingRegex` and an `apiRegex`. For now, the `indexingRegex` would be everything within `src/` excluding `src/api/`.

packages/core/src/build/service.ts
```diff
@@ -48,15 +54,24 @@ export type Build = {
...
+  routes: PonderRoutes;
+};
+
 export type BuildResult =
   | { status: "success"; build: Build }
   | { status: "error"; error: Error };
 
+export type BuildResultServer =
+  | { status: "success"; build?: BuildServer }
+  | { status: "error"; error: Error };
```
0xOlias, line 73: Similar to comment above - I think this would make more sense as `IndexingBuildResult` and `ApiBuildResult`.

packages/core/src/database/sqlite/service.ts
```diff
@@ -51,6 +51,8 @@ export class SqliteDatabaseService implements BaseDatabaseService {
   private userNamespace: string;
   private internalNamespace: string;
 
+  userDatabaseFile: string;
```
0xOlias, line 54: I think for SQLite we should open a new readonly database connection.

https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#new-databasepath-options

packages/core/src/server/service.ts
```diff
@@ -1,47 +1,59 @@
...
 }): Promise<Server> {
-  const hono = new Hono<{
-    Variables: { store: ReadonlyStore; getLoader: GetLoader };
-  }>();
+  // Create hono app
 
-  let port = common.options.port;
   const startTime = Date.now();
 
+  const readonlyStore = getReadonlyStore({
```
0xOlias, line 43: Why'd you choose to instantiate the stores here rather than in `bin`?

packages/core/src/hono/index.ts
```diff
@@ -0,0 +1,61 @@
+import type { Hono } from "hono";
+import type { Handler, MiddlewareHandler } from "./handler.js";
+
+export type PonderRoutes = {
+  method: "GET" | "POST" | "USE";
```
0xOlias, line 5: IS `"USE"` supposed to be here?

packages/core/src/graphql/index.ts
```diff
@@ -0,0 +1,72 @@
...
+import { buildGraphQLSchema } from "./buildGraphqlSchema.js";
+import { buildLoaderCache } from "./buildLoaderCache.js";
+
+/**
+ * Middleware for GraphQL with an interactive web view.
+ *
+ * - Docs: [TODO(kyle)]
+ *
+ * @example
+ * import { ponder } from "@/generated";
```
0xOlias, line 16: I think should be from "@ponder/core" ?

packages/core/src/graphql/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  } = {
+    // Default limits are from Apollo:
+    // https://www.apollographql.com/blog/prevent-graph-misuse-with-operation-size-and-complexity-limit
+    maxOperationTokens: 1000,
+    maxOperationDepth: 100,
+    maxOperationAliases: 30,
+  },
+) =>
+  createMiddleware(async (c) => {
+    const readonlyStore = c.get("readonlyStore");
```
0xOlias, line 38: Could be interesting to attempt to implement the GraphQL resolvers using the drizzle select API.

But in the meantime, we might want to obfuscate this key name if we're not going to expose/document it. Same for metadataStore.

packages/core/src/graphql/index.ts
```diff
@@ -0,0 +1,72 @@
...
+    const metadataStore = c.get("metadataStore");
+    const schema = c.get("schema");
+    const graphqlSchema = buildGraphQLSchema(schema);
+
+    if (c.req.method === "GET") {
+      return c.html(graphiQLHtml(c.req.path));
+    }
+
+    const yoga = createYoga({
+      schema: graphqlSchema,
```
0xOlias, line 48: Still thinking on the best way to make it possible to "extend". the default GraphQL schema here.

0xOlias (commented):
Looking great!

packages/core/src/_test/utils.ts
```diff
@@ -694,10 +694,12 @@ export async function waitForIndexedBlock(
       reject(new Error("Timed out while waiting for the indexed block."));
     }, 5_000);
     const interval = setInterval(async () => {
-      const response = await fetch(`http://localhost:${port}/status`);
+      const response = await fetch(`http://localhost:${port}/_ponder/status`);
```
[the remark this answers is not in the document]
kyscott18: Nice catch

package.json
```diff
@@ -32,7 +36,6 @@
   "packageManager": "pnpm@8.6.10",
   "pnpm": {
     "patchedDependencies": {
-      "graphql@16.8.1": "patches/graphql@16.8.1.patch",
```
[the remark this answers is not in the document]
kyscott18: The patched dependency wasn't working with the new middleware setup, my guess is some vite thing because vite is executing that code.

packages/core/src/bin/commands/dev.ts
```diff
@@ -53,41 +55,80 @@ export async function dev({ cliOptions }: { cliOptions: CliOptions }) {
...
+    ) => {
+      if (result.type === "indexing") {
+        await cleanupReloadable();
+        await cleanupReloadableServer();
+
+        if (result.status === "success") {
+          uiService.reset();
+          metrics.resetMetrics();
+
+          cachedBuild = result.build;
```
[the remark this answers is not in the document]
kyscott18: Caching the build so that if the server reloads it still has access to build artifacts it needs (database config, schema)

packages/core/src/bin/commands/dev.ts
```diff
@@ -53,41 +55,80 @@ export async function dev({ cliOptions }: { cliOptions: CliOptions }) {
...
     uiService.kill();
   };
 
   const shutdown = setupShutdown({ common, cleanup });
 
+  let cachedBuild: Build;
+  let cachedBuildServer: BuildServer | undefined;
+
+  // Note: an update to the "indexing" build triggers a reload of the server runtime.
+  // This is to ensure that updates to either the config or schema are reflected in
```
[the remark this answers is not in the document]
kyscott18: > In other words, a change to an indexing function will never actually change API function behavior.
Yes, but a change to the schema will change API function behavior, and the schema + config + indexing function build is still coupled

packages/core/src/hono/index.ts
kyscott18: yes

packages/core/src/server/service.ts
kyscott18: Seemed more concise, don't have to have duplicate code in `bin/commands/serve.ts` and `bin/utils/runServer.ts`

packages/core/src/drizzle/runtime.ts
```diff
@@ -0,0 +1,236 @@
...
+              drizzleColumns[columnName]!.primaryKey();
+          } else if (isOptionalColumn(column) === false) {
+            drizzleColumns[columnName] = drizzleColumns[columnName]!.notNull();
+          }
+        }
+      }
+    }
+
+    if (database.kind === "postgres") {
+      if (dbNamespace === "public") {
```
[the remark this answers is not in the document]
kyscott18: drizzle throws and error if you try to set the namespace to public

PR #944, opened 2024-06-18
dynamic batch size based on number of columns

kyscott18:


packages/core/src/common/options.ts
```diff
@@ -105,6 +106,7 @@ export const buildOptions = ({ cliOptions }: { cliOptions: CliOptions }) => {
     databaseHeartbeatInterval: 10 * 1000,
     databaseHeartbeatTimeout: 25 * 1000,
 
+    indexingMaxQueryParams: 5000,
```
0xOlias, line 109: SQLite max is 32766 (https://www.sqlite.org/limits.html#max_variable_number)

Postgres max is 65535 https://www.postgresql.org/docs/current/limits.html

kyscott18: Don't have access to the database type in this file, so set it to 30k

0xOlias (approved):

PR #954, opened 2024-06-29
Add totalCount field to plural connection type

jaydenwindle:
This PR adds an additional field `totalCount` to the connection type for all plural queries. If the `totalCount` field is included in a query, an additional database call is made to fetch the total number of unique records matching the currently applied `where` filters. If the `totalCount` field is not queried, no additional database calls are made.


0xOlias:
Hi, thanks for opening. I'm open to this addition, but have some ideas for how to tweak it:
- I think it's more common to include the `totalCount` field on the Connection type itself (see https://graphql.org/learn/pagination/#end-of-list-counts-and-connections). In our case, that would look like this:
```graphql
query {
  friends(where: { id_lt: 100 }) {
    items {
      id
      name
      age
    }
    pageInfo {
      endCursor
      hasNextPage
    }
    totalCount # On the Page type, not the PageInfo type
  }
}
```
- I'm mildly concerned that users will include `totalCount` more often than needed, which can easily wreck performance. When rendering a paginated view, you likely only need to query the total count once up-front. Don't have an immediate solution - might be as simple as a warning callout in the docs.
- There's a bit more work to do to update the `findMany` function signature type, and I have some naming ideas there. Will follow up with a few commits. We also need to update two docs pages accordingly:
  - https://ponder.sh/docs/query/graphql#pagination
  - https://ponder.sh/docs/indexing/create-update-records#findmany

jaydenwindle:
Thanks for considering this!

We can definitely move `totalCount` into the Connection type, that makes a lot more sense.

I agree that the performance considerations should be mentioned in the docs. It seems like there may be a way to use CTEs to include both queries in a single database call without too much performance overhead. Happy to investigate this if performance is a big concern.

jaydenwindle:
@0xOlias I have refactored the `totalCount` query logic to use CTEs instead of a second database call in order to mitigate the performance concerns. I've also moved the `totalCount` field to the connection type and renamed the `findMany` argument to `withTotalCount`.

Let me know if there are any other changes you would like me to make!

tk-o:
@0xOlias what do you think about this PR? The resulting change would be much welcomed 🙃

0xOlias:
The migration to Drizzle in `0.7` means we have to use a slightly different approach here, which I just implemented in #1259.

Closing in favor of that PR - thanks @jaydenwindle for the effort on this, and thanks @tk-o for the nudge.

PR #963, opened 2024-07-10
Status endpoint with progress per network

kyscott18:


packages/core/src/database/postgres/migrations.ts
```diff
@@ -29,6 +30,9 @@ class StaticMigrationProvider implements MigrationProvider {
 export const migrationProvider = new StaticMigrationProvider();
 
 export type InternalTables = {
+  _metadata: {
+    status: Status;
+  };
```
0xOlias, line 35: Is this correct?

PR #970, opened 2024-07-16
feat: `CASCADE` on `DROP VIEW` operation

robiiinos:
Hello,
We submit this change to modify the logic when dropping existing views prior to creating new ones that will serve data from the isolated Ponder schema, to include `CASCADE` option.

In our workflow, we'd like to implement custom views based on the Ponder instance views to compute data and calculate statistics, which we cannot do through the current ORM limitations (such as `MEDIAN`, etc..); but we are aware that using the actual Ponder would work great for simple things such as `COUNT`, `SUM`, .. (otherwise we would have created an entity and stored this data directly onto it).

To solve our views being dropped by the `CASCADE` option, we'd be using `TRIGGER` based on the `sql_drop` internal Postgres event to catch these changes and re-create our views on-the-fly. However, with the current behavior, the indexer fails to drop its views during updates due to our dependent custom views, blocking its change to 'healthy'.

```
ERROR:  cannot drop view "******" because other objects depend on it
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
```

Regarding tests, they should cover the addition of this option already, and we are not adding new code, but rather using the kysely lib' helpers.

Happy to hear any feedback, and insights about this change!


0xOlias:
Hi @robiiinos, appreciate the PR and especially your thorough comment. The requirement makes sense.

If able, could you share a bit more context about your system? I'm specifically curious if our upcoming API functions feature might solve your use case without needing changes at the database level. With that feature, you'll be able to write custom raw SQL queries directly in your app, offering a much better developer experience than what we have today. WIP docs here: https://ponder-docs-git-kjs-api2-ponder-sh.vercel.app/docs/query/api-functions

More broadly, we're interested in simplifying the database design considerably, ideally going as far as removing the pattern of using views to serve the latest data. Still at the design stage here though, so feedback like this is very helpful.

0xOlias (approved):
LGTM, added a test.

Thanks for the additional context btw, that's very helpful. Will keep this use case / requirement in mind.

PR #972, opened 2024-07-17
fix: p.many bug with self references

### LINKED ISSUE #971: [Bug] Type error for `p.many()` with self reference
There is an unnecessary type error when using the `p.many()` virtual field with a self reference. This works fine at runtime, so we should eliminate the type error.

```ts
export default createSchema((p) => ({
  Domain: p.createTable({ // <--- type error here
    id: p.string(),
    parentId: p.string().references("Domain.id"),

    // One-to-one self-reference
    parent: p.one("parentId"),

    // One-to-many self-reference
    // All Domains where parentId = this Domain's id
    subdomains: p.many("Domain.parentId"),
  }),
})
```

kyscott18:
Close #971

### DIFF
--- a/.changeset/yellow-points-try.md
+++ b/.changeset/yellow-points-try.md
@@ -1,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
+Fixed type error for `p.many()` with self reference.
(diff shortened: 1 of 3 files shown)


0xOlias (approved):
LGTM

PR #983, opened 2024-07-20
fix: Grafana Dashboard definitions

robiiinos:
Wanted to have better overview of my Ponder instances and discovered that there was a Prometheus and Grafana setup available (there seems not to be mentioned anywhere in the README or the documentation, TO-DO?) and successfully managed to setup instances and data ingestion on Railway (after a quite long fail-and-retry process).

Noticed a bunch of graph's were out-of-date and using deprecated and/or renamed fields exposed on the `/metrics` endpoint. Documentation make it well-aware that this is not the final API, but I found it quite helpful to have a bunch of definitions to quickstart everything, so I went ahead and updated all graph's as required.

There is still a bunch of things to address before this can be merged, I've added comment's where input is appreciated.

### DIFF
--- a/.changeset/grumpy-yaks-travel.md
+++ b/.changeset/grumpy-yaks-travel.md
@@ -1,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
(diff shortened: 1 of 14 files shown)


typedarray:
Thanks for your work on this but closing for now - lots of metric names have changed in recent versions. Will revisit once the metric design is more stable.

PR #991, opened 2024-07-22
feat: Implement support for alternative Subgraph providers

mirshko:
Hey there, 

I've added in cursory support for alternate subgraph providers, for example Alchemy Subgraph (formerly Satsuma)

I've used the templates format to allow for more graph-compliant spec subgraph providers in the future.

Would love ya'lls feedback on this.

One thing I noticed is the `chainIdByGraphNetwork` values being hard-coded is this possible something we could have as a param passed into the subgraph template so we could build ponder apps for l2s / l3s that are not yet in that file?

### DIFF
--- a/packages/create-ponder/src/helpers/getGraphProtocolChainId.ts
+++ b/packages/create-ponder/src/helpers/getGraphProtocolChainId.ts
@@ -15,8 +15,6 @@ const chainIdByGraphNetwork: Record<string, number | undefined> = {
   fantom: 250,
   "fantom-testnet": 4002,
   bsc: 56,
-  chapel: -1,
-  clover: 0,
   avalanche: 43114,
   fuji: 43113,
   celo: 42220,
@@ -24,7 +22,6 @@ const chainIdByGraphNetwork: Record<string, number | undefined> = {
   fuse: 122,
   moonbeam: 1284,
   moonriver: 1285,
-  mbase: -1,
   base: 8453,
   "base-sepolia": 84532,
   "arbitrum-one": 42161,
@@ -36,7 +33,7 @@ const chainIdByGraphNetwork: Record<string, number | undefined> = {
 };
 
 export const getGraphProtocolChainId = (networkName: string) => {
-  return chainIdByGraphNetwork[networkName];
+  return chainIdByGraphNetwork[networkName] ?? 0;
 };
 
 export const subgraphYamlFileNames = ["subgraph.yaml"].concat(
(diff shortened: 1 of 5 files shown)


packages/create-ponder/src/helpers/getGraphProtocolChainId.ts
```diff
@@ -36,7 +33,7 @@ const chainIdByGraphNetwork: Record<string, number | undefined> = {
 };
 
 export const getGraphProtocolChainId = (networkName: string) => {
-  return chainIdByGraphNetwork[networkName];
+  return chainIdByGraphNetwork[networkName] ?? 0;
```
0xOlias, line 36: Updated this to default to `0` rather than throw, per your comment.

packages/create-ponder/src/index.ts
```diff
@@ -259,7 +263,22 @@ export async function run({
   }
 
   let subgraph: string | undefined = options.subgraph;
+  let subgraphProvider: SubgraphProviderId | undefined =
+    options.subgraphProvider;
   if (templateMeta.id === "subgraph") {
+    if (subgraphProvider === undefined) {
+      const result = await prompts({
```
0xOlias, line 270: If the user doesn't include the `--subgraph-provider` option, they will get a prompt to select one:

packages/create-ponder/src/index.ts
```diff
@@ -504,6 +527,10 @@ export async function run({
     )
     .option("--etherscan [url]", "Use the Etherscan template")
     .option("--subgraph [id]", "Use the subgraph template")
+    .option(
+      "--subgraph-provider [provider]",
```
0xOlias, line 531: Renamed to avoid ambiguity.

0xOlias (approved):
Thanks for opening - added a couple commits to get this across the line quickly.

Screen Shot 2024-07-22 at 3 42 28 PM

Note that the subgraph template is very barebones right now - lots of room for improvement here (convert schema, add `ponder.on(...)` placeholders for the specific events that the subgraph uses rather than the first 4 in each ABI).

packages/create-ponder/src/index.ts
mirshko: Oh sweet, I initially tried this approach, but the prompt was never triggering for me.

mirshko:
Awesome, looking forward to getting this in!

kyscott18 (approved):

PR #995, opened 2024-07-23
fix: SQLite Directory Selection to Prefer Config Directory

erensanlier:
## Summary

This PR addresses an issue with the SQLite directory selection process in `packages/core/src/build/configAndIndexingFunctions.ts`. The code has been updated to prefer the `config.database.directory` if it is provided. If the directory is not specified in the configuration, it will fall back to the default directory `{ponderDir}/sqlite`.

## Changes

- Updated `packages/core/src/build/configAndIndexingFunctions.ts` to use `config.database.directory` if available when setting the SQLite directory.
- Ensured proper resolution of the directory path.

## Testing

I didn't test this, couldn't figure out how to. Would appreciate your help @kyscott18 @0xOlias.

## Related Work

- https://github.com/ponder-sh/ponder/pull/679

### DIFF
--- a/.changeset/metal-apes-care.md
+++ b/.changeset/metal-apes-care.md
@@ -1,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
+Fixed a bug where the SQLite `directory` option in `ponder.config.ts` was not respected.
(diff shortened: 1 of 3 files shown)


0xOlias (approved):
Nice find, added a test and LGTM!

PR #1003, opened 2024-07-30
fix: ipv6 support by default

sinasab:
Even though there's [a comment in the code](https://github.com/ponder-sh/ponder/blob/main/packages/core/src/server/service.ts#L216-L218) indicating that the default hostname behavior should adaptively use ipv6 when possible, the default CLI arg prevents this from actually happening in practice. An easy way to verify this is running a simple example project and attempting to navigate to `http://[::1]:42069/health`, which should fail.

This PR updates the default param to be an empty string, which seems to exhibit the intended behavior. 

Tested locally by verifying both normal `http://localhost:4069/health` and `http://[::1]:42069/health` work for one of the example projects.

Note that it's possible to work around this issue by specifying `--hostname ::` when running `ponder start`.

### DIFF
--- a/.changeset/wise-tomatoes-appear.md
+++ b/.changeset/wise-tomatoes-appear.md
@@ -1,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
+Fix default ipv6 support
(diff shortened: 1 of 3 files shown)


0xOlias (approved):
Great catch, I'm sure it wasn't easy to spot this. Looks like this was a regression introduced in [this PR](https://github.com/ponder-sh/ponder/pull/770).

LGTM.

PR #1005, opened 2024-07-31
Enforce minimum polling interval

### LINKED ISSUE #1004: Enforce minimum `pollingInterval`
An error should be thrown when `pollingInterval` is set below a minimum threshold (~100ms)

kyscott18:
Close #1004

### DIFF
--- a/packages/core/src/build/configAndIndexingFunctions.ts
+++ b/packages/core/src/build/configAndIndexingFunctions.ts
@@ -264,7 +264,7 @@ export async function buildConfigAndIndexingFunctions({
         network.pollingInterval! < 100
       ) {
         throw new Error(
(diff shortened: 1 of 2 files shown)


packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -259,6 +259,15 @@ export async function buildConfigAndIndexingFunctions({
         }
       });
 
+      if (
+        network.pollingInterval !== undefined &&
+        network.pollingInterval! < 100
+      ) {
+        throw new Error(
+          `Invalid 'pollingInterval' for network '${networkName}'. Minimum is 100 ms.`,
```
0xOlias, line 267: Maybe "Expected 100 milliseconds or greater , got ${network.pollingInterval} milliseconds."

0xOlias (approved):
LGTM

PR #1011, opened 2024-07-31
Add Translate Schema When Starting a New Repo With Subgraph

beepidibop:
`schema.graphql` isn't automatically translated to `ponder.schema.ts` when a new repo is started with the `subgraph` option.

This PR adds this function.

I tested the function with a few `schema.graphql` from [Messari](https://github.com/messari/subgraphs/tree/master/subgraphs) and they seem to work so far. There might still be some edge cases that haven't been accounted for.

image

### DIFF
--- a/packages/create-ponder/src/helpers/graphqlTranslator.ts
+++ b/packages/create-ponder/src/helpers/graphqlTranslator.ts
@@ -1,0 +1,248 @@
+import { gql } from "graphql-tag";
+
+import type {
+  DefinitionNode,
+  NamedTypeNode,
+  NonNullTypeNode,
+} from "graphql/language";
+
+const GraphqlTypesToPonderTypes = {
+  ID: "string",
+  String: "string",
+  Int: "int",
+  BigInt: "bigint",
+  BigDecimal: "float",
+  Float: "float",
+  Boolean: "boolean",
+  Bytes: "hex",
+};
+
+export function translateSchema(schema: string): string {
+  const schemaAST = gql(schema);
+  const result: string[] = [];
+  const enums: string[] = [];
+  const enumsResult: string[] = [];
+  const idTypes: { [table: string]: string } = {};
+  const referenceFields: { [table: string]: { [key: string]: string } } = {};
+  const tables: {
+    [table: string]: {
+      columns: { [key: string]: string };
+      descriptions: { [key: string]: string };
+    };
+  } = {};
+
+  // Parse enums
+  // Need to do all the enums first before processing other types
+  // because there is no other way to determine if a graphql definition is referring to
+  // an enum or another table
+  Object.keys(schemaAST.definitions).forEach((definitionName) => {
+    const definition = schemaAST.definitions[
+      definitionName as any as number
+    ] as DefinitionNode;
+    if (definition.kind === "EnumTypeDefinition") {
+      const enumName = definition.name.value;
+      enums.push(enumName);
+      const values: string[] = [];
+      definition.values?.forEach((value) => {
+        values.push(value.name.value);
+      });
+
+      enumsResult.push(
+        `	${enumName}: p.createEnum(${JSON.stringify(values)}),\n\n`,
+      );
+    }
+  });
+
+  // Record id types for ObjectTypeDefinition
+  // This is needed for references later on
+  Object.keys(schemaAST.definitions).forEach((definitionName) => {
+    const definition = schemaAST.definitions[
+      definitionName as any as number
+    ] as DefinitionNode;
+    if (definition.kind === "ObjectTypeDefinition") {
+      const tableName = definition.name.value;
+      definition.fields?.forEach((field) => {
+        const fieldName = field.name.value;
+        if (fieldName === "id") {
+          const nonNullTypeNode = field.type as NonNullTypeNode;
+          const namedTypeNode = nonNullTypeNode.type as NamedTypeNode;
+          if (!namedTypeNode?.name?.value) {
+            console.log(field);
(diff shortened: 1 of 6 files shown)


packages/create-ponder/src/helpers/graphqlTranslator.ts
```diff
@@ -0,0 +1,265 @@
+import { gql } from "graphql-tag";
```
0xOlias, line 1: Any reason not to use `parse` from `graphql`? Looks like `graphql-tag` uses it under the hood anyway.

packages/create-ponder/src/helpers/graphqlTranslator.ts
```diff
@@ -0,0 +1,265 @@
...
+    if (definition.kind === "ObjectTypeDefinition") {
+      const tableName = definition.name.value;
+      definition.fields?.forEach((field) => {
+        const fieldName = field.name.value;
+        if (fieldName === "id") {
+          const nonNullTypeNode = field.type as NonNullTypeNode;
+          const namedTypeNode = nonNullTypeNode.type as NamedTypeNode;
+          if (!namedTypeNode?.name?.value) {
+            console.log(field);
+            throw new Error(`Unsupported id type for ${tableName}`);
```
0xOlias, line 72: Are you aware of a test case where this is reachable?

packages/create-ponder/src/helpers/graphqlTranslator.ts
```diff
@@ -0,0 +1,265 @@
...
+    const definition = schemaAST.definitions[
+      definitionName as any as number
+    ] as DefinitionNode;
+    if (definition.kind === "ObjectTypeDefinition") {
+      const tableName = definition.name.value;
+      tables[tableName] = { columns: {}, descriptions: {} };
+      definition.fields?.forEach((field) => {
+        const fieldName = field.name.value;
+        tables[tableName]!.descriptions[fieldName] = field.description?.value
+          ? `// ${field.description?.value}`
```
0xOlias, line 94: Epic that we can get the column descriptions as well!

packages/create-ponder/src/helpers/graphqlTranslator.ts
```diff
@@ -0,0 +1,265 @@
...
+		${Object.keys(referenceFields?.[tableName] ?? {})
+      .map(
+        (key) => `
+		${key}: ${referenceFields![tableName]![key]}`,
+      )
+      .join(", \n")}
+	}),\n`,
+        );
+      } else {
+        throw new Error(
```
0xOlias, line 133: Same question as above - any idea if this is reachable?

packages/create-ponder/src/subgraph.ts
```diff
@@ -99,6 +100,21 @@ export const fromSubgraphId = async ({
...
+  );
+
+  const schemaPath = path.join(rootDir, "./ponder.schema.ts");
+
+  // Translate and write the schema file
+  const translatedSchema = translateSchema(schemaRaw);
+
+  writeFileSync(
+    schemaPath,
+    await prettier.format(translatedSchema, { parser: "typescript" }),
```
0xOlias, line 115: FWIW, this will take care of whitespace and newline characters, so we can omit them above if it helps readability.

0xOlias (changes requested):
Awesome work! Quick questions.

I'm a bit worried about throwing at all in this file. I think the best outcome is to collect a list of warning messages and log them at the end - we already do this with the etherscan template if e.g. the ABI is not found. This would make it so `create-ponder` always succeeds, but the schema file might be only partially generated.

packages/create-ponder/src/helpers/graphqlTranslator.ts
```diff
@@ -0,0 +1,265 @@
...
+      const values: string[] = [];
+      definition.values?.forEach((value) => {
+        values.push(value.name.value);
+      });
+
+      enumsResult.push(
+        `	${enumName}: p.createEnum(${JSON.stringify(values)}),\n\n`,
+      );
+    }
+  });
```
0xOlias, line 55: Just a style nit - I'd write this as
```ts
// Build enums first because they are required to determine if a
// column type references an enum or another table.
for (const definition of schemaAST.definitions) {
  if (definition.kind === "EnumTypeDefinition") {
    const name = definition.name.value;
    const values = definition.values?.map((value) => value.name.value) ?? [];
    enums.push(name);
    enumsResult.push(`${name}: p.createEnum(${JSON.stringify(values)})`);
  }
}
```

packages/create-ponder/src/helpers/graphqlTranslator.ts
beepidibop: No real reason, it was just the first thing that I found that works

packages/create-ponder/src/helpers/graphqlTranslator.ts
beepidibop: Probably not. It's there as a check to see if the id type is not a list but yeah it probably won't happen on an already deployed subgraph.

packages/create-ponder/src/helpers/graphqlTranslator.ts
beepidibop: The Graph docs had some other types other than `@entity`, but yeah it should be a skip+warning instead of throwing an error
image

packages/create-ponder/src/subgraph.ts
beepidibop: Yee I agree

typedarray:
Closing for now because we moved a new Drizzle-based schema definition in `0.7`.

PR #1014, opened 2024-08-08
refactor `sync`

kyscott18:
### `sync`

- Moved chain specific logic into `localSync` 
- Limited almost all checkpoint comparison to one function
- `status` table is much more accurate during historical backfill
- `Source`, `Event`, and `Filter` types improved

### `sync-store`
- Combined `postgres` and `sqlite` submodules
- Low level access to each table (`insertLogFilterInterval` => `insertLogs` + `insertBlock` + `insertInterval`(
- `filterChildAddresses` takes a list of potential child addresses and removes addresses that are not
- Combined all `interval` manipulation into two functions (`insertInterval` and `getIntervals`)

### `sync-realtime`

- Events are passed to `sync` in-memory without being written to `sync-store`
- Avoids bringing all child addresses into memory
- More accurate filter matching logic

### `sync-historical`

- Idempotent, no longer a long running process (remove `p-queue` and `emittery`)
- `sync` function takes an interval and returns a promise when the interval is synced
- `blockCache` used to avoid fetching the same block twice

### Other

- Slightly improve `decodeEvents` performance


packages/core/src/sync/index.ts
```diff
@@ -1,318 +1,592 @@
...
-  transactionPosition: number;
-  type: "suicide";
-};
-export type SyncRewardTrace = {
-  action: {
-    author: Address;
-    rewardType: "block" | "uncle";
-    value: Hex;
+
+  /** Updates `status` to record progress for each network. */
```
0xOlias, line 177: Might be useful to have a longer comment here explaining the "fallback" logic, eg
```
1. If historical indexing is complete, use the most recent block present in the realtime sync with a checkpoint less than or equal to the overall indexed checkpoint.
2. If during historical indexing, use the most recent block present in the latest batch of events. If there are no events in the batch for a given network, the status block does not get updated for that network. It will look like that network is behind, but it actually just means there weren't any events.
3. ???
```

packages/core/src/sync/index.ts
```diff
@@ -1,318 +1,592 @@
...
+      );
+      // Sync the next interval of each chain.
+      await Promise.all(_localSyncs.map((l) => l.sync()));
+      /**
+       * `latestBlock` is used to calculate the `to` checkpoint, if any
+       * network hasn't yet ingested a block, run another iteration of this loop.
+       * It is an invariant that `latestBlock` will eventually be defined. See the
+       * implementation of `LocalSync.latestBlock` for more detail.
+       */
+      if (_localSyncs.some((l) => l.latestBlock === undefined)) continue;
```
0xOlias, line 305: Reading this line, the naming of `localSync.latestBlock` is a bit confusing considering this is the historical sync (would expect to see "historical" somewhere).

packages/core/src/sync/index.ts
```diff
@@ -1,318 +1,592 @@
...
+      const to = getChainsCheckpoint("latest") ?? end;
+
+      /*
+       * Extract events with `syncStore.getEvents()`, paginating to
+       * avoid loading too many events into memory.
+       */
+      while (true) {
+        if (isKilled) return;
+        if (from === to) break;
+        const limit = args.common.options.syncEventsQuerySize;
```
0xOlias, line 319: `getEventsMaxBatchSize` ?

packages/core/src/sync/index.ts
```diff
@@ -1,318 +1,592 @@
...
+        // convert `estimateSeconds` to checkpoint
+        const _estimate = encodeCheckpoint({
```
0xOlias, line 321: Maybe `estimatedTo`

packages/core/src/sync/index.ts
```diff
@@ -1,318 +1,592 @@
...
+            // Update local sync, record checkpoint before and after
+            let from = getChainsCheckpoint("latest")!;
+            localSync.latestBlock = event.block;
+            const to = getChainsCheckpoint("latest")!;
+
+            // Add block, logs, transactions, receipts, and traces to the sync-store.
+
+            const chainId = network.chainId;
+
+            await Promise.all([
```
0xOlias, line 406: Awesome that this is all in one place now, clear path to better batching, in-memory tricks, etc

0xOlias (commented):
Epic PR - almost all nits.

packages/core/src/bin/utils/run.ts
```diff
@@ -154,38 +128,27 @@ export async function run({
...
+           * Note: statusBlocks should be assigned before any other
+           * synchronous statements in order to prevent race conditions and
+           * ensure its correctness.
+           */
+          const status = sync.getStatus();
+
+          const result = await handleEvents(
+            decodeEvents(common, sources, event.events),
+            event.checkpoint,
+          );
```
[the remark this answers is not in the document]
kyscott18: Have a separate branch, kjs/decode, that does this and add some optimizations to `decodeEvents`

packages/core/src/bin/utils/run.ts
```diff
@@ -245,42 +206,41 @@ export async function run({
...
     // checkpoint is between the last processed event and the finalized
     // checkpoint.
-    common.metrics.ponder_indexing_completed_seconds.set(
-      syncService.checkpoint.blockTimestamp -
-        syncService.startCheckpoint.blockTimestamp,
-    );
-    common.metrics.ponder_indexing_completed_timestamp.set(
-      syncService.checkpoint.blockTimestamp,
-    );
+    if (end) {
```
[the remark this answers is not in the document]
kyscott18: Never, fixed

packages/core/src/sync-historical/index.ts
```diff
@@ -0,0 +1,530 @@
...
+
+  // const logMetadata = new Map<LogFilter, { range: number }>();
+
+  /**
+   * Intervals that have been completed for all "event" and "address" filters
+   * in `args.sources`.
+   *
+   * Note: `intervalsCache` is not updated after a new interval is synced.
+   */
+  const intervalsCache: Map<Filter | Factory, Interval[]> = new Map();
```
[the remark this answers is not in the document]
kyscott18: Not stateful across calls, just a snapshot at startup (see comment)

packages/core/src/sync-historical/index.ts
```diff
@@ -0,0 +1,530 @@
...
+
+          // sync required intervals, account for chunk sizes
+          await Promise.all(
+            requiredIntervals.map(async (interval) => {
+              if (source.type === "contract") {
+                const filter = source.filter;
+                switch (filter.type) {
+                  case "log": {
+                    const maxChunkSize =
+                      source.maxBlockRange ?? args.network.defaultMaxBlockRange;
```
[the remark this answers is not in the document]
kyscott18: yes, this is more for specific rpc providers that support different ranges for errors

packages/core/src/sync-historical/index.ts
```diff
@@ -0,0 +1,530 @@
...
+   * Extract block, using `blockCache` to avoid fetching
+   * the same block twice. Also, update `latestBlock`.
+   *
+   * @param number Block to be extracted
+   * @param transactionHashes Hashes to be inserted into the sync-store
+   *
+   * Note: This function could more accurately skip network requests by taking
+   * advantage of `syncStore.hasBlock` and `syncStore.hasTransaction`.
+   */
+  const syncBlock = async (
```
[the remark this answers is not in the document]
kyscott18: Not true right now, as "logs" or "callTraces" can have many entries that correspond to the same block

packages/core/src/sync-historical/index.ts
```diff
@@ -0,0 +1,530 @@
...
+
+          if (isKilled) return;
+
+          await blockPromise;
+
+          // Mark `interval` for `filter` as completed in the sync-store
+          await args.syncStore.insertInterval({
+            filter: source.filter,
+            interval,
+          });
```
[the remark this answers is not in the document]
kyscott18: We don't. Instead we assume that this interval will never be called again

packages/core/src/sync/index.ts
```diff
@@ -1,318 +1,592 @@
...
+      )
+    ) {
+      return undefined;
+    }
+
+    const checkpoints = [...localSyncs.entries()]
+      // Remove completed networks from checkpoint calculation for "latest"
+      .filter(
+        ([, localSync]) => tag !== "latest" || localSync.isComplete() === false,
+      )
```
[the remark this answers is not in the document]
kyscott18: Nice, this is way easier to understand

packages/core/src/sync/index.ts
kyscott18: Did you see the inline comments

packages/core/src/sync-store/index.ts
```diff
@@ -211,7 +207,7 @@ export const createSyncStore = ({
               await tx
                 .insertInto("blockFilters")
                 .values(fragment)
-                .onConflict((oc) => oc.column("id").doUpdateSet(fragment))
+                .onConflict((oc) => oc.column("id").doNothing())
                 .executeTakeFirstOrThrow();
```
0xOlias: I don't think you need the `executeTakeFirstOrThrow` here, could probably cause issues

packages/core/src/sync-store/index.ts
```diff
@@ -433,11 +429,10 @@ export const createSyncStore = ({
...
         await db.transaction().execute(async (tx) => {
           while (true) {
-            const { id } = await tx
+            await tx
               .insertInto(`${table}s`)
               .values(fragment)
-              .onConflict((oc) => oc.column("id").doUpdateSet(fragment))
-              .returningAll()
+              .onConflict((oc) => oc.column("id").doNothing())
               .executeTakeFirstOrThrow();
```
0xOlias, line 436: Same here

packages/core/src/sync/index.ts
```diff
@@ -152,26 +150,29 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
-        return blockToCheckpoint(
-          block,
-          network.chainId,
-          tag === "start" ? "down" : "up",
-        );
-      });
-
-    // Return `true` if all networks have been filtered out
+    let checkpoints = [...localSyncs.entries()];
+    if (tag === "latest")
```
0xOlias, line 154: Could add comment "When calculating the latest block, ignore any networks where the historical sync is complete"

packages/core/src/sync/index.ts
```diff
@@ -152,26 +150,29 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
+      checkpoints = checkpoints.filter(
+        ([, localSync]) => localSync.isComplete() === false,
+      );
+    // Return early if all networks are complete
     if (checkpoints.length === 0) return undefined;
```
0xOlias, line 159: To confirm: If I ask for the "finalized" block when all historical syncs are complete, I'll get `undefined` ?

PR #1024, opened 2024-08-14
fix: #1022

### LINKED ISSUE #1022: cli eta column format
The ETA column moves around a lot while syncing.

https://github.com/user-attachments/assets/b007b93b-6ddc-4329-8764-4bd0f6eaf661

Prefixing single digit minutes and seconds helps prevent it from shifting.

https://github.com/user-attachments/assets/b9e80bf6-be0f-49ee-92c7-844ad7a8abd5

Happy to open a PR to prefix these single digits. You are also welcome to close this issue if it's no big deal.

tmm:
Fixes #1022

### DIFF
--- a/packages/core/src/utils/format.ts
+++ b/packages/core/src/utils/format.ts
@@ -9,7 +9,7 @@ export const formatEta = (ms: number) => {
 
   const hstr = h > 0 ? `${h}h ` : "";
   const mstr = m > 0 || h > 0 ? `${m < 10 && h > 0 ? "0" : ""}${m}m ` : "";
-  const sstr = s > 0 || m > 0 ? `${s < 10 ? "0" : ""}${s}s` : "";
+  const sstr = s > 0 || m > 0 ? `${s < 10 && m > 0 ? "0" : ""}${s}s` : "";
 
   return `${hstr}${mstr}${sstr}`;
 };
--- a/packages/core/src/utils/format.ts
+++ b/packages/core/src/utils/format.ts
@@ -8,8 +8,8 @@ export const formatEta = (ms: number) => {
   const s = seconds - h * 3600 - m * 60;
 
   const hstr = h > 0 ? `${h}h ` : "";
-  const mstr = m > 0 || h > 0 ? `${m}m ` : "";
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/format.ts
```diff
@@ -8,8 +8,8 @@ export const formatEta = (ms: number) => {
   const s = seconds - h * 3600 - m * 60;
 
   const hstr = h > 0 ? `${h}h ` : "";
-  const mstr = m > 0 || h > 0 ? `${m}m ` : "";
-  const sstr = s > 0 || m > 0 ? `${s}s` : "";
+  const mstr = m > 0 || h > 0 ? `${m < 10 ? "0" : ""}${m}m ` : "";
```
tmm, line 11: might only be worth prefixing if `h > 0`

```suggestion
  const mstr = m > 0 || h > 0 ? `${m < 10 && h > 0 ? "0" : ""}${m}m ` : "";
```

packages/core/src/utils/format.ts
```diff
@@ -8,8 +8,8 @@ export const formatEta = (ms: number) => {
...
+  const sstr = s > 0 || m > 0 ? `${s < 10 ? "0" : ""}${s}s` : "";
```
tmm, line 12: might only be worth prefixing if `m > 0`

```suggestion
  const sstr = s > 0 || m > 0 ? `${s < 10 && m > 0 ? "0" : ""}${s}s` : "";
```

packages/core/src/utils/format.ts
tmm, line 1: could also add a space ` ` instead of a `0` since it will take up the same width 🤷‍♂️

packages/core/src/utils/format.ts
0xOlias: Yea I think it looks better without the leading zero.

before
```
│     ETA │
┼─────────┤
│ 03m 08s │
```

after
```
│     ETA │
┼─────────┤
│  3m 08s │
```

packages/core/src/utils/format.ts
0xOlias: Tried it - I think `0` looks better!

0xOlias (approved):
LGTM, confirmed locally

PR #1026, opened 2024-08-14
Accurate historical -> realtime handoff

kyscott18:


packages/core/src/sync/local.ts
```diff
@@ -127,7 +128,12 @@ export const createLocalSync = async (
       if (block === undefined) return;
       _latestBlock = block;
     },
-    finalizedBlock,
+    get finalizedBlock() {
```
kyscott18, line 131: Probably a clue that we should just include this module in `sync/index.ts` at some point instead of separating it in `sync/local.ts`. Let's do this when we start playing with multichain impls

0xOlias (approved):

PR #1030, opened 2024-08-15
remove filter transactions and redundant inserts

kyscott18:


packages/core/src/sync-store/index.ts
```diff
@@ -164,98 +164,62 @@ export const createSyncStore = ({
...
-                    })
-                    .execute();
-                }
+            for (const fragment of buildLogFilterFragments(filter)) {
+              if (isAddressFactory(filter.address)) {
+                await db
+                  .insertInto("factoryLogFilterIntervals")
+                  .values({
+                    factoryId: fragment.id,
+                    ...intervalToBlock(interval),
```
0xOlias, line 173: Could move this to the upper scope and re-use

PR #1032, opened 2024-08-15
batch insert transactions

kyscott18:


packages/core/src/sync-historical/index.ts
```diff
@@ -480,11 +488,23 @@ export const createHistoricalSync = async (
...
         }),
       );
-      await args.syncStore.insertBlocks({
-        blocks: await Promise.all(blockCache.values()),
-        chainId: args.network.chainId,
-      });
+
+      await Promise.all([
+        args.syncStore.insertBlocks({
+          blocks: await Promise.all(blockCache.values()),
```
0xOlias, line 494: Maybe do this before and reuse?
```
const blocks = await Promise.all(blockCache.values())
```

packages/core/src/sync-historical/index.ts
```diff
@@ -321,14 +326,17 @@ export const createHistoricalSync = async (
...
-      });
+      // Filter out transactions that aren't relevant
+      const newTransactions = block.transactions.filter(({ hash }) =>
+        transactionHashes.has(hash),
+      );
+
+      // Add `newTransactions` to the cache
+      for (const transaction of newTransactions) {
+        transactionsCache.add(transaction.hash);
+      }
```
0xOlias, line 339: I don't think the `block.transactions.filter( ... )` is necessary here? Could do
```
// Add `newTransactions` to the cache
if (transactionHashes !== undefined) {
  for (const transactionHash of transactionHashes) {
    transactionsCache.add(transaction.hash);
  }
}
```

Someday we could use [`Set.prototype.union()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/union)!

kyscott18: Nice catch - after looking at this more closely it made sense to just include the transaction specific logic in the filter functions. Now `syncLogFilter` and `syncCallTraceFilter` manage adding the required transaction to the `transactionCache`

0xOlias (approved):

PR #1039, opened 2024-08-16
Add `sync_duration` metric

kyscott18:
Record `sync_duration` for each network.

Use this new metric to provide a more accurate sync ETA. Also display sync for each network, instead of for each network + source


0xOlias (approved):
LGTM

PR #1044, opened 2024-08-16
Update estimate parameters

kyscott18:


packages/core/src/sync/index.ts
```diff
@@ -322,28 +326,36 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
-        const receivedDensity = (cursorTime - fromTime) / (events.length || 1);
 
-        // Use range and number of events returned to update estimate
-        // 10 <= estimate(new) <= estimate(prev) * 2
-        estimateSeconds = Math.min(
-          Math.max(10, Math.round(receivedDensity * getEventsMaxBatchSize)),
-          estimateSeconds * 2,
-        );
+        try {
+          const { events, cursor } = await args.syncStore.getEvents({
```
0xOlias, line 331: One potential challenge here is that the timeouts might get caught in the sync store error/retry logic. We might need to move some amount of this logic into that layer.

To see what this looks like, you can set the `statement_timeout` parameter at the pg pool level to something like 500ms (we currently default it to 2 minutes).

0xOlias (approved):
LGTM

PR #1048, opened 2024-08-19
Export new `alias` helper function for select API

kyscott18:


.changeset/eight-bees-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
+Exported "alias" helper function.
```
0xOlias, line 5: Exported Drizzle [`alias`](https://orm.drizzle.team/docs/joins#aliases--selfjoins)  function which can be used in API functions.

packages/core/src/index.ts
```diff
@@ -57,4 +57,7 @@ export {
   sumDistinct,
   max,
   min,
+  TableAliasProxyHandler,
```
0xOlias, line 60: Do we need to export this?

kyscott18: Oh no that was a bug

PR #1052, opened 2024-08-20
Database refactor

kyscott18:
## Database service
- Only considers one schema
- no publishing views
- ignore schema "lock" when using `ponder dev`
- drop finalized rows from reorg tables

## Server service
- `/healthy` returns `200` immediately
- `/ready` returns `200` after historical backfill is finished
- remove `maxHealthcheckDuration`


packages/core/src/database/index.ts
```diff
@@ -0,0 +1,1211 @@
...
+              }
+
+              // Revert unfinalized data
+
+              const { blockTimestamp, chainId, blockNumber } = decodeCheckpoint(
+                previousApp.checkpoint,
+              );
+              args.common.logger.info({
+                service: "database",
+                msg: `Reverting operations prior to finalized checkpoint (timestamp=${blockTimestamp} chainId=${chainId} block=${blockNumber})`,
```
0xOlias, line 955: `Reverting operations after finalized checkpoint`

packages/core/src/database/index.ts
```diff
@@ -0,0 +1,1211 @@
...
+
+              return {
+                status: "success",
+                checkpoint: previousApp.checkpoint,
+              } as const;
+            }
+
+            /**
+             * If table name collision, exit
+             * Else, start
```
0xOlias, line 974: Is this comment correct? Looks like you may have forgotten to write the table name collision logic.

packages/core/src/database/index.ts
```diff
@@ -0,0 +1,1211 @@
...
+        if (result.status === "locked") {
+          throw new NonRetryableError(
+            `Failed to acquire lock on schema '${namespace}'. A different Ponder app is actively using this database.`,
+          );
+        }
+      }
+
+      heartbeatInterval = setInterval(async () => {
+        try {
+          const app = await getApp(orm.internal);
```
0xOlias, line 1042: Seems a bit weird that we need to get the app each time. Could instead use JSON operators (https://www.sqlite.org/json1.html#jset) to only update the heartbeat_at property.

Something like this:
```
.set({
  value: ksql`json_set(value, '$.heartbeat_at', ${encodeApp({ ...app!, heartbeat_at: Date.now() })})`,
})
```

packages/core/src/database/index.ts
```diff
@@ -0,0 +1,1211 @@
...
+            provider: postgresMigrationProvider,
+            migrationTableSchema: "ponder_sync",
+          });
+        }
+
+        const { error } = await migrator.migrateToLatest();
+        if (error) throw error;
+      });
+    },
+    async prepareEnv({ buildId }) {
```
0xOlias, line 531: Any reason not to call this `setup` ?

packages/core/src/database/index.ts
```diff
@@ -0,0 +1,1211 @@
...
+      });
+
+      await orm.internal.destroy();
+      await orm.user.destroy();
+      await orm.readonly.destroy();
+      await orm.sync.destroy();
+
+      if (sql === "sqlite") {
+        // @ts-ignore
+        driver.user.close();
```
0xOlias, line 1189: Don't need to close the internal database here?

packages/core/src/database/index.test.ts
```diff
@@ -0,0 +1,803 @@
...
+    .where("key", "=", "app")
+    .select("value")
+    .executeTakeFirst();
+
+  await databaseTwo.orm.internal
+    .updateTable("_ponder_meta")
+    .where("key", "=", "app")
+    .set({
+      value:
+        database.sql === "sqlite"
```
0xOlias, line 284: Another opportunity to use JSON operators

packages/core/src/database/index.test.ts
```diff
@@ -0,0 +1,803 @@
...
+    database.sql === "sqlite"
+      ? JSON.parse(row!.value!).checkpoint
+      : // @ts-ignore
+        row!.value!.checkpoint,
+  ).toStrictEqual(encodeCheckpoint(maxCheckpoint));
+
+  await database.kill();
+});
+
+test("finalize delete reorg table rows", async (context) => {
```
0xOlias, line 429: Nice test

packages/core/src/database/index.test.ts
```diff
@@ -0,0 +1,803 @@
...
+
+    const indexes = await getUserIndexNames(databaseTwo, "Person");
+
+    expect(indexes).toStrictEqual([
+      database.sql === "sqlite" ? "sqlite_autoindex_Person_1" : "Person_pkey",
+    ]);
+
+    await databaseTwo.kill();
+  },
+  { timeout: 30_000 },
```
0xOlias, line 646: Is this necessary?

0xOlias (commented):
LFG, great work! Mostly nits

packages/core/src/database/index.ts
```diff
@@ -0,0 +1,1211 @@
...
+    );
+  }
+  ////////
+  // Helpers
+  ////////
+
+  const getApp = async (db: Kysely<PonderInternalSchema>) => {
+    const row = await db
+      .selectFrom("_ponder_meta")
+      .where("key", "=", "app")
```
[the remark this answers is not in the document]
kyscott18: "status" is also a valid key. Check out the type `PonderInternalSchema`

packages/core/src/database/index.ts
```diff
@@ -0,0 +1,1211 @@
...
+                            return col;
+                          },
+                        );
+                      }
+                    }
+
+                    builder = builder
+                      .addColumn(
+                        "operation_id",
+                        sql === "sqlite" ? "integer" : "serial",
```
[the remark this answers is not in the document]
kyscott18: yes

packages/core/src/database/index.ts
kyscott18: `internal` and `user` isn't differentiated in sqlite. I thought that because they use the same file it doesn't make a difference, is this incorrect?

packages/core/src/database/index.ts
```diff
@@ -0,0 +1,1211 @@
...
+                  // @ts-ignore
+                  path.join(args.databaseConfig.directory, "ponder.db-wal"),
+                  {
+                    force: true,
+                  },
+                );
+                args.common.logger.debug({
+                  service: "database",
+                  msg: `Removed '.ponder/sqlite/ponder.db' file`,
+                });
```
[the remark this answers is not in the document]
kyscott18: Yes

packages/core/src/database/index.ts
```diff
@@ -0,0 +1,1211 @@
...
+              previousApp.build_id === buildId &&
+              previousApp.checkpoint !== encodeCheckpoint(zeroCheckpoint)
+            ) {
+              await tx
+                .updateTable("_ponder_meta")
+                .set({
+                  value: encodeApp({
+                    ...previousApp,
+                    is_locked: true,
+                    is_dev: false,
```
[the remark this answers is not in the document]
kyscott18: What would the bug be?

packages/core/src/database/index.ts
```diff
@@ -0,0 +1,1211 @@
...
+
+    driver = {
+      user: _createSqliteDatabase(userFile),
+      readonly: createReadonlySqliteDatabase(userFile),
+      sync: _createSqliteDatabase(syncFile),
+    };
+
+    driver.user.exec(`ATTACH DATABASE '${userFile}' AS public`);
+    driver.readonly.exec(`ATTACH DATABASE '${userFile}' AS public`);
+    // driver.readonly.exec(`ATTACH DATABASE '${syncFile}' AS public`);
```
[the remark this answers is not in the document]
kyscott18: I don't think so, now we just use the `WithSchema` kysely plugin for postgres and the query builder should just write to the correct schema everywhere in the codebase

packages/core/src/database/index.ts
```diff
@@ -0,0 +1,1211 @@
...
+
+  return {
+    sql,
+    namespace,
+    driver,
+    orm,
+    async migrateSync() {
+      await orm.sync.wrap({ method: "migrateSyncStore" }, async () => {
+        // TODO: Probably remove this at 1.0 to speed up startup time.
+        // TODO(kevin) is the `WithSchemaPlugin` going to break this?
```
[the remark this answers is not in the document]
kyscott18: Yes, could you look at this and how it interacts with the `moveLegacyTables` function? Want to make sure that I didn't break the migration

packages/core/src/database/index.ts
```diff
@@ -0,0 +1,1211 @@
...
+                .where("namespace", "=", namespace)
+                .execute();
+
+              if (namespaceCount!.count === 1) {
+                fs.rmSync(
+                  // @ts-ignore
+                  path.join(args.databaseConfig.directory, "ponder.db"),
+                  {
+                    force: true,
+                  },
```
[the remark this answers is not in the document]
kyscott18: Great catch

packages/core/src/database/index.ts
```diff
@@ -0,0 +1,1211 @@
...
+            }
+
+            /**
+             * If schema is locked, exit
+             *
+             * Determine if the schema is locked by examining the lock and heartbeat
+             */
+            const expiry =
+              previousApp.heartbeat_at +
+              args.common.options.databaseHeartbeatTimeout;
```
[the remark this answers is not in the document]
kyscott18: All in ms

PR #1055, opened 2024-08-21
Dynamic log ranges in historical sync

kyscott18:
Deprecated `maxBlockRange`


packages/core/src/sync-historical/index.ts
```diff
@@ -66,7 +75,10 @@ export const createHistoricalSync = async (
    */
   const transactionsCache = new Set<Hash>();
 
-  // const logMetadata = new Map<LogFilter, { range: number }>();
+  const logRequestMetadata = new Map<
```
0xOlias, line 78: `getLogsRequestMetadata` ?

packages/core/src/config/networks.ts
```diff
@@ -8,65 +8,10 @@ export type Network = {
   transport: ReturnType<Transport>;
   pollingInterval: number;
   maxRequestsPerSecond: number;
-  defaultMaxBlockRange: number;
   finalityBlockCount: number;
   disableCache: boolean;
 };
 
-export function getDefaultMaxBlockRange({
```
0xOlias, line 16: Stoked to delete this.

packages/utils/src/getLogsRetryHelper.ts
```diff
@@ -24,6 +24,8 @@ export type GetLogsRetryHelperReturnType =
       shouldRetry: true;
       /** Suggested values to use for (fromBlock, toBlock) in follow-up eth_getLogs requests. */
       ranges: { fromBlock: Hex; toBlock: Hex }[];
+      /** `true` if the error message suggested to use this range on retry. */
+      isSuggestedRange: boolean;
```
0xOlias, line 28: I wonder if instead of a boolean, this should be an optional number equal to a "confirmed" static range. Eg for a Cloudflare error, it would return `staticMaxRange: 800`.

packages/core/src/sync-historical/index.ts
```diff
@@ -133,6 +152,91 @@ export const createHistoricalSync = async (
...
+  const logRequestHelper = async ({
+    address,
+    topics,
+    interval,
+    filter,
+  }: {
+    address?: Address | Address[];
+    topics?: LogTopic[];
+    interval: Interval;
+    filter: LogFilter | LogFactory;
```
0xOlias, line 167: Aren't the address and topics available on the `filter` ?

packages/core/src/sync-historical/index.ts
```diff
@@ -133,6 +152,91 @@ export const createHistoricalSync = async (
...
+  }): Promise<SyncLog[]> => {
+    /**
+     * Use the recommended range if available, else don't chunk the interval at all
```
0xOlias, line 170: Could be a one line comment

packages/core/src/sync-historical/index.ts
```diff
@@ -133,6 +152,91 @@ export const createHistoricalSync = async (
...
+
+    let intervals: Interval[];
+    if (getLogsRequestMetadata.has(filter)) {
+      intervals = getChunks({
+        interval,
+        maxChunkSize: getLogsRequestMetadata.get(filter)!.range,
+      });
+    } else {
+      intervals = [interval];
+    }
```
0xOlias, line 181: Could do
```ts
const metadata = getLogsRequestMetadata.get(filter)
const intervals = metadata
  ? getChunks({ interval, maxChunkSize: metadata.range })
  : [interval]
```

packages/core/src/sync-historical/index.ts
```diff
@@ -65,9 +74,19 @@ export const createHistoricalSync = async (
...
+   * Data about the range passed to "eth_getLogs" for all log
+   *  filters and log factories.
+   */
+  const getLogsRequestMetadata = new Map<
+    LogFilter | LogFactory,
+    {
+      /** Estimate optimal range to use for "eth_getLogs" requests */
+      range: number;
+      /** `true` if an exact range has been suggested by the error message */
+      isSuggestedByError: boolean;
```
0xOlias, line 87: Another idea would be `estimatedRange` and `confirmedRange`, where `confirmedRange` only gets set when we are certain of a hard cap. Then always prefer `confirmedRange` and fall back to `estimatedRange` if not present.

packages/core/src/sync-historical/index.ts
```diff
@@ -133,6 +152,91 @@ export const createHistoricalSync = async (
   // Helper functions for specific sync tasks
   ////////
 
+  /**
+   * Split "eth_getLogs" requests into ranges inferred from errors.
+   */
+  const logRequestHelper = async ({
```
0xOlias, line 158: I would call this `getLogsRequestHelper` or `getLogsDynamic`

0xOlias (commented):
Couple naming / structure ideas. This is going to be great.

packages/core/src/sync-historical/index.ts
kyscott18: Yes, going to update this a bit, but it's not as straight-forward as just using `filter` because of factory contracts

0xOlias (approved):

PR #1058, opened 2024-08-21
Extract realtime data concurrently

kyscott18:


packages/core/src/sync-realtime/index.ts
```diff
@@ -52,15 +52,18 @@ type CreateRealtimeSyncParameters = {
   onFatalError: (error: Error) => void;
 };
 
+type BlockData = {
```
kyscott18, line 55: Seems like there exists a better name for this

packages/core/src/sync-realtime/index.ts
```diff
@@ -126,14 +129,64 @@ export const createRealtimeSync = (
...
+
+        if (isAddressFactory(filter.address)) {
+          return isMatched && childAddresses.has(log.address);
+        }
+
+        return isMatched;
+      }),
+    );
+
+    // Remote call traces that don't match a filter, accounting for factory addresses
```
0xOlias, line 174: Typo, "Remove"

packages/core/src/sync-realtime/index.ts
```diff
@@ -276,8 +329,12 @@ export const createRealtimeSync = (
 
   /**
    * Request and filter all relevant data pertaining to `block` and `args.sources`
```
0xOlias, line 331: Fetch all data (logs, traces, receipts) for the specified block required by `args.sources`

packages/core/src/sync-realtime/index.ts
```diff
@@ -356,46 +413,20 @@ export const createRealtimeSync = (
...
-      logFilters.some((filter) => {
-        const isMatched = isLogFilterMatched({ filter, block, log });
-
-        if (isAddressFactory(filter.address)) {
-          return isMatched && childAddresses.has(log.address);
-        }
-
-        return isMatched;
-      }),
+      logFilters.some((filter) => isLogFilterMatched({ filter, block, log })),
```
0xOlias, line 418: Is it worthwhile to filter here if we're going to filter again later on? The factory filtering _must_ happen sequentially (can't happen in `extract` which runs in parallel), so it might make sense to do all filtering in handleBlock? Don't feel strongly either way.

packages/core/src/sync-realtime/index.ts
```diff
@@ -126,14 +129,64 @@ export const createRealtimeSync = (
...
       msg: `Started syncing '${args.network.name}' block ${hexToNumber(block.number)}`,
     });
 
-    const { logs, callTraces, transactions, transactionReceipts } =
-      await extract(block);
+    // Get all addresses from `logs` that are from a factory
+    const childAddresses = new Set<Address>();
+    for (const filter of factories) {
+      if (logs.length > 0) {
+        const _childAddresses = await args.syncStore.filterChildAddresses({
```
0xOlias, line 148: Might be worth creating an issue to load this once on startup and maintain it in-memory as new blocks come in.

packages/core/src/sync-realtime/index.ts
```diff
@@ -516,7 +548,9 @@ export const createRealtimeSync = (
               );
               const pendingBlocks = await Promise.all(
                 missingBlockRange.map((blockNumber) =>
-                  _eth_getBlockByNumber(args.requestQueue, { blockNumber }),
+                  _eth_getBlockByNumber(args.requestQueue, {
+                    blockNumber,
+                  }).then((block) => extract(block)),
```
0xOlias, line 553: We could rename `extract` to something like `fetchRequiredBlockData` or `fetchFullBlockData` if we like that name

0xOlias (approved):
LGTM

packages/core/src/sync-realtime/index.ts
```diff
@@ -134,14 +137,87 @@ export const createRealtimeSync = (
...
+
+      for (let i = 0; i < logFilters.length; i++) {
+        if (isFilterMatched[i]!) matchedFilters.add(logFilters[i]!);
+      }
+
+      return isFilterMatched.some((m) => m);
+    });
+
+    // Remove call traces that don't match a filter, accounting for factory addresses
+    callTraces = callTraces.filter((callTrace) => {
```
0xOlias, line 191: Confirming my understanding here - is this logic equivalent?
```ts
callTraces = callTraces.filter((callTrace) => {
  let isMatched = false;
  
  for (const filter of callTraceFilters) {
    if (
      isCallTraceFilterMatched({ filter, block, callTrace }) &&
      (isAddressFactory(filter.toAddress)
        ? childAddresses.has(callTrace.action.to)
        : true)
    ) {
      matchedFilters.add(filter);
      isMatched = true;
    }
  }
  
  return isMatched;
});
```

kyscott18: yes

PR #1061, opened 2024-08-22
Update `viem` peer dependency

### LINKED ISSUE #988: [BUG] `context.client.getEnsName` does not exist
While inside an indexing function, if we use `context.client` to lookup an ENS Name using `getEnsName`, it will throw because it's not a function.

Function + Error:
```
TypeError: context.client.getEnsName is not a function
    at /Desktop/ponder/my-app/src/index.ts:4:38
  2 |
  3 | ponder.on("ExampleContract:Transfer", async ({ context }) => {
> 4 |     const ens = await context.client.getEnsName({
    |                                      ^
  5 |         address: "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045"
  6 |     });
  7 |
```

Expected result: `getEnsName` should return `vitalik.eth`

kyscott18:
Also closes #988


0xOlias (approved):
LGTM, just needs a changeset

PR #1062, opened 2024-08-23
Cache `checksumAddress` results

kyscott18:
Benchmark for `reference-erc20` with a full sync cache:

| before | after | delta |
|--------|-----|-------|
|    2319 | 1370| 40.9%|


jaylmiller (commented):
seems good but perhaps we want to add some test coverage for these utility functions

PR #1063, opened 2024-08-27
Fix: extra blocks added to database in realtime

kyscott18:


packages/core/src/sync-realtime/index.ts
```diff
@@ -395,9 +411,24 @@ export const createRealtimeSync = (
         }
 
         return isMatched;
-      }),
+      });
+
+      for (let i = 0; i < callTraceFilters.length; i++) {
+        if (isFilterMatched[i]!) matchedFilters.add(callTraceFilters[i]!);
```
0xOlias, line 417: Could move this into the map above

packages/core/src/sync/index.ts
```diff
@@ -430,7 +430,10 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
             const chainId = network.chainId;
 
             await Promise.all([
-              args.syncStore.insertBlocks({ blocks: [event.block], chainId }),
+              args.syncStore.insertBlocks({
```
0xOlias, line 433: Should we also conditionally call these sync store methods based on `event.filters` ? Or are they each smart enough to no-op if there is no data to insert?

kyscott18: They each no-op and skip metrics if the data is empty

packages/core/src/sync-realtime/index.ts
kyscott18: After a second glance, with the way there are two separate branches in the block above for filters that have a factory or not, I think it is more readable how it is right now

PR #1064, opened 2024-09-03
fix(ui): display hostname correctly in ui

jaylmiller:


packages/core/src/ui/app.tsx
```diff
@@ -374,7 +376,9 @@ const App = (ui: UiState) => {
       <Box flexDirection="column">
         <Text bold>GraphQL </Text>
         <Box flexDirection="row">
-          <Text>Server live at http://localhost:{port}</Text>
+          <Text>
+            Server live at http://{hostname}:{port}
```
kyscott18, line 380: should we also update the transport "https"

packages/core/src/ui/app.tsx
```diff
@@ -76,13 +77,14 @@ export const buildUiState = () => {
     },
 
     port: 0,
+    hostname: "localhost",
```
0xOlias, line 80: Made a change so that if the user has not specified a custom hostname, it will display `localhost` instead of `0.0.0.0` (matches Next.js behavior).

The key fix (display the custom hostname if the user provided one) remains. Does that seem OK @jaylmiller ?

jaylmiller: `0.0.0.0` seems more accurate since thats what we default to, no?

PR #1074, opened 2024-09-05
v0.6

kyscott18:
* remove deprecated maxHistoricalTaskConcurrency config

* more

* chore: changeset


oxSaturn:
May I ask why `maxBlockRange` was removed and what's the replacement?

kyscott18:
> May I ask why `maxBlockRange` was removed and what's the alternative?

We dynamically detect allowed ranges based on rpc errors. If this is causing an issue on your app, please report it

PR #1080, opened 2024-09-06
Re-introduce concurrency between `sync` and `indexing`

kyscott18:


packages/core/src/sync/index.ts
```diff
@@ -280,117 +356,155 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
+      if (allHistoricalSyncExhaustive) break;
+
+      /**
+       * At least one network has a `finalized` block that is considered "stale".
+       */
+
+      latestFinalizedFetch = Date.now();
+
+      await Promise.all(
+        Array.from(localSyncData.entries()).map(
```
0xOlias, line 475: Note that you could skip refetching finalized block for any chains that are exhaustive/completed here.

packages/core/src/sync/index.ts
```diff
@@ -78,24 +150,81 @@ type CreateSyncParameters = {
...
+          }),
+        onFatalError: args.onFatalError,
+      });
+
+      historicalSync.initializeMetrics(
+        blockProgress.finalized as SyncBlock,
+        true,
+      );
+
+      localSyncData.set(network, {
```
0xOlias, line 221: Maybe `localSyncContext` ?

packages/core/src/sync/index.ts
```diff
@@ -280,117 +356,155 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
     let from = start;
 
     while (true) {
-      const _localSyncs = args.networks.map(
-        (network) => localSyncs.get(network)!,
+      const syncGenerator = mergeAsyncGenerators(
+        Array.from(localSyncData.values()).map(
+          ({ blockProgress, historicalSync }) =>
+            localHistoricalSyncHelper({
```
0xOlias, line 362: Maybe `localHistoricalSyncGenerator` ?

packages/core/src/sync/index.ts
```diff
@@ -60,13 +64,81 @@ export type Status = {
...
+  return {
+    ...(rounding === "up" ? maxCheckpoint : zeroCheckpoint),
+    blockTimestamp: hexToNumber(block.timestamp),
+    chainId: BigInt(chainId),
+    blockNumber: hexToBigInt(block.number),
+  };
+};
+
+/** Returns true if all possible blocks have been synced. */
+export const isSyncExhaustive = (blockProgress: BlockProgress) => {
```
0xOlias, line 101:
```ts
/** Returns true if all filters have a defined end block and the current
 *  sync progress has reached the final end block.
 */
export const isSyncComplete ...
```

0xOlias (commented):
Looks great, some naming ideas.

0xOlias (approved):

Issue #975, opened 2024-07-17
[feature] default values for fields


Sofianel5:
Databases and ORMs generally allow fields to be defined with a default value. Doing this now in ponder is a little janky. For example, I have a schema where I want to keep track of points users accrue (all users start at 0).

## Current Implementation

In order to implement this right now, I need to either explicitly set `points` to 0 whenever I create a new user or set points to be an optional value. The former would be extremely tedious as I create new accounts in many different watchers, and its much cleaner to just do
```
await Account.upsert({
    id: event.args.from,
});
```
then explicitly set points to 0 across all watchers that create accounts but have nothing to do with points. However the latter requires consumers of the indexer to explicitly cast null to 0 on the points field of accounts. Currently, my has to do this.
```
// schema.ponder.ts
export default createSchema((p) => ({
  Account: p.createTable({
    id: p.hex(),
    tokens: p.many("TokenBalance.ownerId"),
    points: p.bigint().optional(),
    transferFromEvents: p.many("TransferEvent.fromId"),
    transferToEvents: p.many("TransferEvent.toId"),
  }),
}));

// Contract.ts
ponder.on("Contract:PointEarned", async ({ event, context }) => {
  const { Account } = context.db;

  await Account.upsert({
    id: event.args.user,
    update: ({ current }) => ({
      points: (current.points ?? 0n) + 1n,
    }),
  });
});
```

## Usage Example

Here's an example of what this might look like in `ponder.schema.ts` with the requested feature.
```
import { createSchema } from "@ponder/core";

export default createSchema((p) => ({
  Account: p.createTable({
    id: p.hex(),
    tokens: p.many("TokenBalance.ownerId"),
    points: p.bigint(0n), // Set a default value of 0 for points
    transferFromEvents: p.many("TransferEvent.fromId"),
    transferToEvents: p.many("TransferEvent.toId"),
  }),
}));
```
Then I don't need to change anything about my indexer code except increment the points counter.
```
ponder.on("Contract:PointEarned", async ({ event, context }) => {
  const { Account } = context.db;

  await Account.upsert({
    id: event.args.user,
    update: ({ current }) => ({
      points: current.points + 1n,
    }),
  });
});
```

0xOlias:
Great idea, this would be a nice quality of life improvement. I propose a slightly different API:

```ts
import { createSchema } from "@ponder/core";

export default createSchema((p) => ({
  Account: p.createTable({
    id: p.hex(),
[...]

PR #1085, opened 2024-09-09
Add `.default()` as column definition modifier

3commascapital:
- [x] add hook for use in schema
- [x] use default for method name
- [x] write tests to verify that default value is used


packages/core/src/database/postgres/service.test.ts
```diff
@@ -55,6 +56,16 @@ const schemaTwo = createSchema((p) => ({
   }),
 }));
 
+const schemaWithDefaults = createSchema((p) => ({
+  Dog: p.createTable({
+    id: p.string().defaultTo("0"),
+    name: p.string().defaultTo("firstname"),
+    age: p.int().defaultTo(5).optional(),
+    bigAge: p.bigint().defaultTo(5n).optional(),
```
0xOlias, line 64: Couple general notes:
- Per the discussion [here](https://github.com/ponder-sh/ponder/issues/975) I propose we go with `.default()` over `.defaultTo()`
- I'm not sure if this is handled yet, but adding `.default` to a column should automatically/implicitly make it optional. Perhaps we should validate against including both `.optional()` and `.default()` to a column?

3commascapital: ah, thanks for linking. forgot to do that. regarding your points:
* if we do not mind the js `.default` export pattern and wish to distinguish, i am ok with that
* I don't know that it should actually. Optional only denotes nullability, which should still be possible to do even if you started with a non null value. In other words, when I first insert, I don't have a value, the database will use the default value for me, but I should be able to turn that into null later on or use null during insertion.

0xOlias: Re 1: I think it's fine - this is being used in a context rather distant from a module export. Fwiw Drizzle uses [`.default`](https://orm.drizzle.team/docs/column-types/pg#default-value) as well.

Re 2: Ah yes, my mistake - you're right. The interesting case is: if a column has a default and is NOT optional, it should still be optional when inserting or updating records of that table (but not when selecting). I'm not sure how well our existing type system is set up to handle this, cc @kyscott18

3commascapital: 💯

3commascapital: should (probably) be addressable in this pr

packages/core/src/schema/common.ts
```diff
@@ -10,6 +23,7 @@ export type ScalarColumn<
   " scalar": scalar;
   " optional": optional;
   " list": list;
+  " default": DefaultValue<scalar>;
```
kyscott18, line 26: I think `" default"` may need to be implemented similar to the other properties in this type. The type needs to be as strict as possible so that when we go to infer the types for the store methods (`db.create()`, ...), they are also strictly typed.

This is going to be tricky because of the interplay between, scalar, optional, and default, but I think something closer to this would work:

```ts
export type ScalarColumn<
  scalar extends Scalar = Scalar,
  optional extends boolean = boolean,
  list extends boolean = boolean,
  _default extends optional extends false
    ? scalar | undefined
    : scalar | null | undefined = optional extends false
    ? scalar | undefined
    : scalar | null | undefined,
> = {
  " type": "scalar";
  " scalar": scalar;
  " optional": optional;
  " list": list;
  " default": _default;
};
```

This also applies to the other column types in this file.

packages/core/src/schema/columns.ts
```diff
@@ -8,6 +9,43 @@ import type {
   ScalarColumn,
 } from "./common.js";
 
+type DefaultTo<column extends BuilderScalarColumn> = (
+  val: DefaultValue<column[" scalar"]>,
+) => BuilderScalarColumn<
+  column[" scalar"],
+  column[" optional"],
+  column[" list"]
```
kyscott18, line 17: Need to pass some information about the type of `column[" default"]` to `BuildScalarColumn`

packages/core/src/database/postgres/service.test.ts
```diff
@@ -427,6 +438,46 @@ describe.skipIf(shouldSkip)("postgres database", () => {
     await database.kill();
   });
 
+  test("default values are populated during insertion", async (context) => {
```
kyscott18, line 441: Nice test - is there a way to tell if a column has a default value, similar to `getTableNames` function used in this test file

packages/core/src/database/postgres/service.ts
```diff
@@ -785,6 +792,10 @@ export class PostgresDatabaseService implements BaseDatabaseService {
           columnName,
           scalarToSqlType[column[" scalar"]],
           (col) => {
+            const sCol = column as ScalarColumn;
```
kyscott18, line 795: Should `ReferenceColumn`s also support `.default`? I think it makes sense because they already support `.optional`

packages/core/src/schema/columns.ts
```diff
@@ -8,6 +9,43 @@ import type {
...
+
+const defaultTo =
+  <column extends BuilderScalarColumn>(col: column): DefaultTo<column> =>
+  (val: DefaultValue<column[" scalar"]>) => {
+    const scalar = col[" scalar"];
+    let v = val;
+    if (scalar === "hex") {
+      let value = val as string;
+      if (value?.startsWith("0x")) {
+        value = value.slice(2);
```
kyscott18, line 30: What is this for?

kyscott18 (commented):
Nice work - reviewed most of this. I think we need to add extra logic to `database/sqlite/service.ts` and `schema/infer.ts` but looking good.

packages/core/src/database/postgres/service.test.ts
3commascapital: hm. i don't know. are you asking for type generation purposes? i would presume that it would show up in a `\d table_name` command on postgres

packages/core/src/schema/columns.ts
3commascapital: This is to chop off the 0x from an address or other byte data to be used as default. Otherwise the 0x is stored as hex data, which prefixes the bytes. Here is the test, without that slice: ```

  Array [
    Object {
      "age": 5,
      "bigAge": 5n,
      "bowl": Object {
        "data": Array [
          48,
+         120,
+         48,
          48,
          48,
          48,
          48,
```
as you can see, 0x (48 and 120 respectively) prefix the series of 0s

I now realize, what this really means is that my 0's are being inserted as ascii and not as bytes. I will fix that

File: packages/core/src/database/postgres/service.ts (hunk 5 above)
Comment (3commascapital) [reply packages/core/src/database/postgres/service.ts:795]: i think you are right, that relationals could use a default value as well

File: packages/core/src/schema/schema.ts [hunk 7]
```diff
@@ -123,7 +123,7 @@ const P = {
   many,
   enum: _enum,
   index,
-};
+} as P;
```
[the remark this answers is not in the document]
3commascapital: without this coersion, i get the following error during typecheck: 
```
packages/core build: src/schema/schema.ts(112,7): error TS4023: Exported variable 'P' has or is using name 'DefaultTo' from external module "~/ponder/packages/core/src/schema/columns" but cannot be named.
```

packages/core/src/schema/common.ts
```diff
@@ -29,6 +43,7 @@ export type JSONColumn<type = any, optional extends boolean = boolean> = {
   " type": "json";
   " json": type;
   " optional": optional;
+  " default": DefaultValue<"string">;
```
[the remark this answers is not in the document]
3commascapital: i probably did the typing wrong here. could certainly use some more specialized knowledge. boolean could theoretically be an id type, but the current system does not allow it

packages/core/src/schema/schema.ts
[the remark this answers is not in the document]
3commascapital: it looks like this is solved by adding the default param [as suggested](https://github.com/ponder-sh/ponder/pull/1085#discussion_r1757583613)

packages/core/src/database/postgres/service.test.ts
3commascapital: i was able to creae one [here](https://github.com/ponder-sh/ponder/pull/1085/files#diff-ae6cfa84163e684033b864abfb27a9245eb057f932f6e2b8b69135bfb5b05fc6R1026)

0xOlias:
Closed in favor of #1177

PR #1088, opened 2024-09-10
Fix utility types imports

swan-of-bodom:
Fix imports from docs https://ponder.sh/docs/utilities/types


0xOlias (approved):
Confirmed, thanks!

PR #1092, opened 2024-09-10
in-memory child address tracking

### LINKED ISSUE #1091: Hold all factory child addresses in memory
When handling factories, the realtime sync currently queries the database on every block to determine which logs or traces are matched by a filter that contains a factory. 

We could skip this database query if instead, on startup, we queried all child addresses, stored them in memory, and updated when a new child is registered.

kyscott18:
Closes #1091 

- [ ] better tests for `getChildAddress`

Realtime child address reorg handling strategy:

- separate `unfinalizedChildAddresses` and `finalizedChildAddresses`
- track `factoryLogsPerBlock`, so that `unfinalizedChildAddresses` can always be recomputed
- on reorg, evict reorged blocks from `factoryLogsPerBlock` and recompute `unfinalizedChildAddresses`
- on finalization, add child addresses from finalized blocks to `finalizedChildAddresses`,  evict finalized blocks from `factoryLogsPerBlock` and recompute `unfinalizedChildAddresses`


packages/core/src/sync-realtime/index.ts
```diff
@@ -624,7 +628,7 @@ export const createRealtimeSync = (
 
             // New block is exactly one block ahead of the local chain.
             // Attempt to ingest it.
-            await handleBlock({ block, ...rest });
+            handleBlock({ block, ...rest });
```
kyscott18, line 631: Happy path worker for the realtime queue (bottleneck for handling a new block) is now synchronous

packages/core/src/sync/index.ts
```diff
@@ -692,7 +696,24 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
+
+          const initialChildAddresses = new Map<Factory, Set<Address>>();
+
+          for (const { filter } of args.sources) {
+            if (
+              filter.chainId === network.chainId &&
+              "address" in filter &&
+              isAddressFactory(filter.address)
+            ) {
+              const addresses = await args.syncStore.getChildAddresses({
```
0xOlias, line 708: Any reason not to use `Promise.all` here?

packages/core/src/sync/source.ts
```diff
@@ -77,3 +78,29 @@ export const isAddressFactory = (
...
+  } else {
+    const start = 2 + 12 * 2;
+    const length = 20 * 2;
+    const topicIndex =
+      factory.childAddressLocation === "topic1"
+        ? 1
+        : factory.childAddressLocation === "topic2"
+          ? 2
+          : 3;
+    log.topics;
```
0xOlias, line 103: Typo?

Looks good to me, but while we're here I think it's worthwhile to add a couple unit tests for this, e.g. around casing

packages/core/src/sync-realtime/index.ts
```diff
@@ -176,7 +175,8 @@ export const createRealtimeSync = (
         if (
           isLogFilterMatched({ filter, block, log }) &&
           (isAddressFactory(filter.address)
-            ? childAddresses.has(log.address)
+            ? finalizedChildAddresses.get(filter.address)!.has(log.address) ||
+              unfinalizedChildAddresses.get(filter.address)!.has(log.address)
```
0xOlias, line 179: Are we certain that the casing is consistent when doing these comparisons?

packages/core/src/sync-realtime/index.ts
```diff
@@ -282,11 +288,56 @@ export const createRealtimeSync = (
...
+          hexToNumber(lb.number) <= hexToNumber(pendingFinalizedBlock.number),
+      );
+
       localChain = localChain.filter(
         (lb) =>
           hexToNumber(lb.number) > hexToNumber(pendingFinalizedBlock.number),
       );
 
+      // add child address from newly finalized blocks to `finalizedChildAddresses`
+      for (const filter of factories) {
```
0xOlias, line 302: Nit, but I'd find this more readable if the `factories` loop came after the `if (factoryLogs !== undefined)` check

packages/core/src/sync-realtime/index.ts
```diff
@@ -282,11 +288,56 @@ export const createRealtimeSync = (
...
+          }
+        }
+      }
+
+      // recompute `unfinalizedChildAddresses`
+      unfinalizedChildAddresses.clear();
+
+      for (const filter of factories) {
+        unfinalizedChildAddresses.set(filter, new Set());
+        for (const { hash } of localChain) {
```
0xOlias, line 322: I wonder if we should rename `localChain` to `unfinalizedBlocks`

packages/core/src/sync-realtime/index.ts
```diff
@@ -321,6 +372,30 @@ export const createRealtimeSync = (
...
+            }
+          }
+        }
+      }
+    }
+
+    // delete reorged blocks from `factoryLogsPerBlock`
+    for (const { hash } of reorgedBlocks) {
+      factoryLogsPerBlock.delete(hash);
+    }
```
0xOlias, line 397: Should this happen before we recompute `unfinalizedChildAddresses` ? It shouldn't matter, just double checking.

0xOlias (commented):
Business logic looks good to me!

PR #1094, opened 2024-09-11
update pnpm

kyscott18:


0xOlias (commented):
Do we need to update the contributing guide, and will existing contributors need to update their `pnpm` version to continue contributing?

PR #1096, opened 2024-09-11
fix: metrics for sources with unfinalized start blocks

kyscott18:


0xOlias (approved):
LGTM, could add a changeset

PR #1101, opened 2024-09-13
Improve observability

kyscott18:


packages/core/src/common/metrics.ts
```diff
@@ -20,24 +20,24 @@ export class MetricsService {
 
   ponder_indexing_total_seconds: prometheus.Gauge;
   ponder_indexing_completed_seconds: prometheus.Gauge;
-  ponder_indexing_completed_events: prometheus.Gauge<"network" | "event">;
+  ponder_indexing_completed_events: prometheus.Gauge<"event">;
```
0xOlias, line 23: We'll need a changeset for the metric label changes - could update the one we already have on `0.6`

packages/core/src/common/metrics.ts
```diff
@@ -20,24 +20,24 @@ export class MetricsService {
...
   ponder_indexing_has_error: prometheus.Gauge;
 
-  ponder_indexing_function_duration: prometheus.Histogram<"network" | "event">;
-  ponder_indexing_function_error_total: prometheus.Counter<"network" | "event">;
-
+  ponder_indexing_function_duration: prometheus.Histogram<"event">;
   ponder_indexing_abi_decoding_duration: prometheus.Histogram;
 
+  ponder_sync_block: prometheus.Gauge<"network">;
+  // ponder_sync_events: prometheus.Gauge<"network">;
```
0xOlias, line 32: Remove?

packages/core/src/common/metrics.ts
```diff
@@ -255,72 +260,87 @@ export async function getHistoricalSyncProgress(metrics: MetricsService) {
...
+  const extractMetric = (
+    metric: prometheus.MetricObjectWithValues<
+      prometheus.MetricValue<"network">
+    >,
+    network: string,
+  ) => {
+    return metric.values.find((m) => m.labels.network === network)?.value;
+  };
+
+  const totalBlocksMetric = await metrics.ponder_historical_total_blocks.get();
```
0xOlias, line 272: It's a pain that all of these have to be async when we know that there is no async `collect` function registered for any of our metrics. As a nerdsnipe, we could create a `SyncRegistry` subclass that enforces sync for everything.

packages/core/src/common/metrics.ts
```diff
@@ -255,72 +260,87 @@ export async function getHistoricalSyncProgress(metrics: MetricsService) {
...
+      rps[networkName]!.shift();
+    }
+  }
 
-  return {
-    overall: progress,
-    networks,
-  };
+  return totalBlocksMetric.values.map(({ value, labels }) => {
+    const totalBlocks = value;
```
0xOlias, line 305: Could make this more compact by doing `const network = labels.network as string` here at the top

packages/core/src/ui/app.tsx
```diff
@@ -31,13 +22,13 @@ export type UiState = {
...
     events: {
       eventName: string;
-      networkName: string;
       count: number;
       averageDuration: number;
-      errorCount: number;
     }[];
   };
 
+  app: Awaited<ReturnType<typeof getAppProgress>>;
```
0xOlias, line 30: Any reason not to do this trick for the `indexing` type right above?

packages/core/src/ui/app.tsx
```diff
@@ -31,13 +22,13 @@ export type UiState = {
...
+
   realtimeSyncNetworks: {
```
0xOlias, line 32: Should be able to remove this from the type and initial object

0xOlias (approved):
Looks great

packages/core/src/sync/index.ts
```diff
@@ -455,6 +474,22 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
 
             yield { events, checkpoint: to };
             from = cursor;
+
+            getAppProgress(args.common.metrics).then(({ eta, progress }) => {
```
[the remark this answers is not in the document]
kyscott18: I thought it was a bad idea to have metrics be a part of the hot loop

packages/core/src/sync/index.ts
```diff
@@ -597,7 +638,16 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
             for (const network of args.networks) {
               updateRealtimeStatus({ checkpoint: cursor, network });
             }
-            args.onRealtimeEvent({ type: "block", checkpoint: to, events });
+            args
```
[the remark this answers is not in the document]
kyscott18: This makes sure the log gets shown after the event is completed

packages/core/src/sync/index.ts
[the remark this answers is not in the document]
0xOlias: Here's the code that's actually running when we await the metrics: https://github.com/siimon/prom-client/blob/master/lib/histogram.js#L102-L125

In practice, we don't use the `collect` function pattern for any of these metrics (only use it for the postgres pool metrics which aren't used internally, and even those are actually synchronous).

PR #1102, opened 2024-09-13
fix: avoid refetching event data for known blocks

kyscott18:


packages/core/src/sync-realtime/index.ts
```diff
@@ -675,11 +675,25 @@ export const createRealtimeSync = (
...
+          const latestLocalBlock = getLatestLocalBlock();
+
+          // We already saw and handled this block. No-op.
+          if (latestLocalBlock.hash === block.hash) {
+            args.common.logger.trace({
+              service: "realtime",
+              msg: `Skipped processing '${args.network.name}' block ${hexToNumber(block.number)}, already synced`,
+            });
+
+            return;
```
0xOlias, line 689: Is it safe to return here?

kyscott18: yes

PR #1110, opened 2024-09-19
retry eth_getLogs for publicnode

chenxsan:
I was running into this error when using publicnode service:

Screenshot-cHr4WfwW@2x

Thought it would be helpful to include it as well.


kyscott18:
This is great! Thank you

kyscott18 (approved):

PR #1115, opened 2024-09-20
in-memory realtime sync

### LINKED ISSUE #1091: Hold all factory child addresses in memory

kyscott18:
The "sync" portion of ponder handling extracting and ordering events is entirely in-memory, removing the need for database queries. Another property is that only finalized data is added to the `sync-store`. This is a combination of #1092 and #1103. Closes #1091.

This involves:
- storing child address information in memory
- filtering,  formatting, and ordering events (replacing `getEvents()`)

### Child address methodology
- separate `unfinalizedChildAddresses` and `finalizedChildAddresses`
- track `factoryLogsPerBlock`, so that `unfinalizedChildAddresses` can always be recomputed
- on reorg, evict reorged blocks from `factoryLogsPerBlock` and recompute `unfinalizedChildAddresses`
- on finalization, add child addresses from finalized blocks to `finalizedChildAddresses`,  evict finalized blocks from `factoryLogsPerBlock` and recompute `unfinalizedChildAddresses`

### Sync ordering methodology
- store unindexed events in `unindexedEvents`
- store unfinalized event data in `unfinalizedEventData`
- when a new block is received, format with `buildEvents()` and add to `unindexedEvents`
- when the overall checkpoint is moved forward, select events that are now able to be indexed, remove them from `unindexedEvents`, and pass them to the indexing functions
- on reorg, removed events with reorged block hashes from `unindexedEvents` and `unfinalizedEventData`
- on finalization, print a warning if the chain's finalized checkpoint is farther along than the overall indexing checkpoint
- on finalization, removed newly finalized data from `unfinalizedEventData` and insert it into sync-store

### Other
A metric to track how far away from tip the indexing progress is (known block - indexed block) might be helpful. However, it also may be unintuitive for many consumers when using omnichain ordering as the metric will appear far behind tip quite consistently


packages/core/src/sync/index.ts
```diff
@@ -566,180 +579,241 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
-          await Promise.all(
-            args.sources
-              .filter(({ filter }) => filter.chainId === network.chainId)
-              .map(({ filter }) =>
-                args.syncStore.insertInterval({ filter, interval }),
-              ),
+          const events: RawEvent[] = unindexedEvents.filter(
+            ({ checkpoint }) => checkpoint <= to,
+          );
+          unindexedEvents = unindexedEvents.filter(
```
0xOlias, line 636: I find it a bit confusing that this is called `unindexedEvents`, because this service does not actually do the indexing. Am I correct that this is the list of events (across all chains) that the sync service is holding onto, but can't pass up to the runtime because there is another network with a lagging checkpoint? Essentially a buffer?

packages/core/src/sync/index.ts
```diff
@@ -566,180 +579,241 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
+            callTraces: finalizedEventData.flatMap(({ callTraces, block }) =>
+              callTraces.map((callTrace) => ({ callTrace, block })),
+            ),
+            chainId: network.chainId,
+          }),
+          ...args.sources
+            .filter(({ filter }) => filter.chainId === network.chainId)
+            .map(({ filter }) =>
+              args.syncStore.insertInterval({ filter, interval }),
+            ),
```
0xOlias, line 746: Really nice that this is now happening in the same spot - not necessary for this PR, but ultimately I think it would be great to use a database transaction here.

packages/core/src/sync/index.ts
```diff
@@ -566,180 +579,241 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
+        const reorgedHashes = new Set<Hash>();
+        for (const b of event.reorgedBlocks) {
+          reorgedHashes.add(b.hash);
         }
 
-        default:
-          never(event);
+        unindexedEvents = unindexedEvents.filter(
+          (e) => reorgedHashes.has(e.block.hash) === false,
+        );
```
0xOlias, line 797: I'm getting a bit confused by the difference between `unfinalizedEventData` and `unindexedEvents` here. It seems like they contain the same information, why do we need both?

packages/core/src/sync/index.ts
```diff
@@ -784,11 +875,6 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
         historicalSync.kill();
         promises.push(realtimeSync.kill());
       }
-
-      eventQueue.pause();
-      eventQueue.clear();
-      promises.push(eventQueue.onIdle());
```
0xOlias, line 790: Why'd you remove these?

packages/core/src/sync-realtime/index.ts
```diff
@@ -254,12 +276,17 @@ export const createRealtimeSync = (
...
 
-    localChain.push(syncBlockToLightBlock(block));
+    unfinalizedBlocks.push(syncBlockToLightBlock(block));
 
-    args.onEvent({
+    // Make sure `transactions` can be garbage collected
+    // @ts-ignore
+    block.transactions = undefined;
+
+    await args.onEvent({
```
0xOlias, line 285: Why did you add the `await` here?

0xOlias (commented):
Looking great, couple questions.

packages/core/src/sync-realtime/index.ts
```diff
@@ -254,12 +276,17 @@ export const createRealtimeSync = (
       });
     }
 
-    localChain.push(syncBlockToLightBlock(block));
+    unfinalizedBlocks.push(syncBlockToLightBlock(block));
 
-    args.onEvent({
+    // Make sure `transactions` can be garbage collected
+    // @ts-ignore
+    block.transactions = undefined;
```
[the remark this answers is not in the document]
kyscott18: yes

packages/core/src/sync-realtime/index.ts
```diff
@@ -317,51 +389,74 @@ export const createRealtimeSync = (
...
+            }
+          }
+        }
+      }
+    }
+
+    // delete reorged blocks from `factoryLogsPerBlock`
+    for (const { hash } of reorgedBlocks) {
+      factoryLogsPerBlock.delete(hash);
+    }
```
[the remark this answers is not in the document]
kyscott18: it shouldn't matter

packages/core/src/sync-store/index.ts
```diff
@@ -638,13 +638,7 @@ export const createSyncStore = ({
...
               ),
           )
-          .onConflict((oc) =>
-            oc.column("hash").doUpdateSet((eb) => ({
-              blockHash: eb.ref("excluded.blockHash"),
-              blockNumber: eb.ref("excluded.blockNumber"),
-              transactionIndex: eb.ref("excluded.transactionIndex"),
-            })),
-          )
+          .onConflict((oc) => oc.column("hash").doNothing())
```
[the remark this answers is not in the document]
kyscott18: This previous logic was to handle the case when a reorg occurred. It no longer applies because all data in the `sync-store` is finalized.

packages/core/src/sync/events.ts
```diff
@@ -95,6 +116,170 @@ export type CallTraceEvent = {
...
+        }
+        break;
+      }
+
+      case "callTrace": {
+        for (const callTraces of Array.from(traceByTransactionHash.values())) {
+          // Use lexographical sort of stringified `traceAddress`.
+          callTraces.sort((a, b) => {
+            return a.traceAddress < b.traceAddress ? -1 : 1;
+          });
```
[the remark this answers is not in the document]
kyscott18: The traces get bucketed by transaction hash:
https://github.com/ponder-sh/ponder/blob/2d208790a1cd698c9bfc5c99c44359a4d3499911/packages/core/src/sync/events.ts#L153-L158

and then the function process all call traces in a transaction in a batch:
https://github.com/ponder-sh/ponder/blob/2d208790a1cd698c9bfc5c99c44359a4d3499911/packages/core/src/sync/events.ts#L224

packages/core/src/sync/index.ts
```diff
@@ -566,180 +579,241 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
-          while (true) {
-            if (isKilled) return;
-            if (from === to) break;
-            const { events, cursor } = await args.syncStore.getEvents({
-              filters: args.sources.map(({ filter }) => filter),
-              from,
-              to,
-              limit: args.common.options.syncEventsQuerySize,
-            });
+        const blockWithEventData = {
```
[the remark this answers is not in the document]
kyscott18: yes, doesn't contain `type`

packages/core/src/sync/index.ts
```diff
@@ -566,180 +579,241 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
-            args.common.logger.info({
-              service: "sync",
-              msg: `Synced final end block for '${network.name}' (${hexToNumber(syncProgress.end!.number)}), killing realtime sync service`,
+          args
+            .onRealtimeEvent({
+              type: "block",
+              checkpoint: to,
+              events: events.sort((a, b) =>
+                a.checkpoint < b.checkpoint ? -1 : 1,
+              ),
```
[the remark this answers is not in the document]
kyscott18: intra-chain ordering

packages/core/src/sync/index.ts
kyscott18: Yes, exactly. Is `eventBuffer` more clear?

packages/core/src/sync/index.ts
kyscott18: The `sync` module no longer contains a queue. Instead, it is synchronous with `sync-realtime`. This is primarily to maintain consistency for factory addresses. Without this change, it would be almost impossible to maintain `unfinalizedChildAddresses` and pass the correct value for the correct block height to `buildEvents`.

packages/core/src/sync-realtime/index.ts
kyscott18: [See above response](https://github.com/ponder-sh/ponder/pull/1115#discussion_r1793637042)

packages/core/src/sync/index.ts
kyscott18: They do contain the same information, where `unindexedEvents` is derived from `unfinalizedEventData`. However, when building events, point-in-time child addresses are needed. Also, the lifetime of each of these variables is different.

packages/core/src/sync/index.ts
0xOlias: Maybe `bufferedEvents` ?

packages/core/src/sync-realtime/index.ts
```diff
@@ -425,9 +520,9 @@ export const createRealtimeSync = (
         );
       }
 
-      callTraces = traces.filter(
-        (trace) => trace.type === "call",
-      ) as SyncCallTrace[];
+      callTraces = traces
+        .filter((trace) => trace.type === "call")
+        .filter((trace) => trace.result !== null) as SyncCallTrace[];
```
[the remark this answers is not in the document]
0xOlias: Makese sense. Though, this means we'll technically be storing different data in the database in historical vs realtime, correct? In realtime, we're filtering before insertion but the opposite in historical.

packages/core/src/sync/index.ts
```diff
@@ -740,12 +738,17 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
             ),
             chainId: network.chainId,
           }),
-          ...args.sources
+        ]);
+
+        // Add corresponding intervals to the sync-store
+        // Note: this should happen after so the database doesn't become corrupted
```
0xOlias, line 744: Nice catch, I bet this would have bit us.

0xOlias (approved):
LGTM

PR #1116, opened 2024-09-22
Fix `schema` and `config` file path in `ponder-env.d.ts` 

KONFeature:
Fix the generated `schema` and `config` types from `ponder-env.d.ts` when `--config` or `--root` args are used

### DIFF
--- a/packages/core/src/common/codegen.ts
+++ b/packages/core/src/common/codegen.ts
@@ -11,8 +11,8 @@ export const ponderEnv = `// This file enables type checking and editor autocomp
 declare module "@/generated" {
   import type { Virtual } from "@ponder/core";
 
-  type config = typeof import("./ponder.config.ts").default;
-  type schema = typeof import("./ponder.schema.ts").default;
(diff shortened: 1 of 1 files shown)


0xOlias:
Hi, thanks for opening - could you share more details about the workflow / repo structure you're using that inspired this change? Seems likely that we can merge, but want to understand the broader need to see if we can solve it more comprehensively.

KONFeature:
Hey, yep so here is the [repo](https://github.com/frak-id/frak-indexer/tree/main/packages/ponder)

Basically, I got a config folder at the root, within that config I got 3 config files, and 1 global config builder. This help me to fine tune the config depending on the env (local / dev / prod), but share the heavy lifting of the definition via the config builder (contracts definition etc)

typedarray:
Thanks again for opening, but we're going to close for now. We still don't have a super satisfying solution for multiple/dynamic configs, but we're confident that the design will use a static `ponder-env.d.ts` file vs. overwriting the paths.

PR #1120, opened 2024-09-23
experimental offchain tables

kyscott18:


packages/core/src/database/index.ts
```diff
@@ -1083,33 +1097,32 @@ export const createDatabase = async (args: {
...
-        // console.log(tableNames, Object.keys(args.schema));
+        for (let i = 0; i < sqlTableNames.length; i++) {
+          const jsTableName = jsTableNames[i]!;
+          const sqlTableName = sqlTableNames[i]!;
 
-        for (const tableName of tableNames) {
-          const columns = getTableColumns(args.schema[tableName]! as PgTable);
+          const columns = getTableColumns(args.schema[jsTableName]! as PgTable);
 
           const columnNames = Object.values(columns).map(
```
0xOlias, line 1109: Might this also run into JS vs SQL naming issues?

packages/core/src/database/index.ts
```diff
@@ -1120,24 +1133,29 @@ $$ LANGUAGE plpgsql
...
             .execute(qb.internal);
         }
       });
     },
     async revert({ checkpoint }) {
-      const tableNames = Object.values(args.schema)
-        .filter((table): table is PgTable => is(table, PgTable))
-        .map((table) => getTableConfig(table).name);
+      const sqlTableNames = getSQLTableNames();
+      const jsTableNames = getJsTableNames();
```
0xOlias, line 1146: I wonder if an API like `function getTableNames(): { js: string; sql :string }[]` would be safer

PR #1126, opened 2024-10-02
Fix: extra transactions added in realtime

kyscott18:
The realtime sync functions in two stages, one concurrent stage to extract chain information corresponding to a block such as logs, call traces, transactions, and transaction receipts. This stage does as much filtering as possible but doesn't have access to all factory addresses. 

The next stage is serial, and factory addresses are available. The logs and call traces are re-filtered but transactions and transaction receipts are not. This is causing extra transactions to be added to the database.

I fixed this issues by re-filtering transactions and transaction receipts in the second stage `handleBlock()`.

### DIFF
--- a/.changeset/bright-bikes-enjoy.md
+++ b/.changeset/bright-bikes-enjoy.md
@@ -2,4 +2,12 @@
 "@ponder/core": patch
 ---
 
-Fixed a bug introduced in v0.6 where extra transaction may be added to the database in the "realtime" sync when using factory contracts."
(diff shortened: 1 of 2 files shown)


kyscott18:
Any users that were affected by this bug and want to reduce the database size can do so with the query:

```sql
DELETE FROM ponder_sync.transactions WHERE 
  hash NOT IN (SELECT "transactionHash" FROM ponder_sync.logs) 
  AND 
  hash NOT IN (SELECT "transactionHash" FROM ponder_sync."callTraces");
```

kyscott18:
I believe the failing ci is most likely the result of a regression in foundry @0xOlias

PR #1129, opened 2024-10-03
Add missing `chainId` validation to `getEvents` subquery

goums:
Use checkpoint limits in `getEvents` nested queries to pre filter only the required checkpoint block range


kyscott18 (commented):
This looks great! I wonder if we could just use `checkpoint` to filter down the subqueries before the `unionAll`  instead of `blockNumber`?

goums:
> This looks great! I wonder if we could just use `checkpoint` to filter down the subqueries before the `unionAll` instead of `blockNumber`?

Yes that could be possible as well indeed.
I though `blockNumber` would be nice as it is already used for the filter log, and will still be required in case the filter has smaller boundaries than the current checkpoint.

kyscott18 (approved):

PR #1130, opened 2024-10-03
fix: progress >100%

kyscott18:
## Description
This fixes a bug that would result in the progress showing > 100%. This was reported in the telegram a few times.

telegram-cloud-photo-size-4-6039785064532919156-x

## Diagnosis
This is a result of incorrect handling of `finalized` block refetching. The current logic is as follows:
- sync all blocks from `start` to `historicalLast`(minimum of `finalized` and `end`)
- check if `finalized` is stale (last fetched more than 5 mins ago)
- if not stale, enter realtime
- else, re-run this loop

However, `start` should not be used on iterations after the first, because this would lead to blocks being re-synced (they have already been synced on the first iteration) 

## Solution
The solution was to detect when `localHistoricalSyncGenerator` should skip blocks that have already been synced, by checking `syncProgress.current`. I validated this worked on the `project-uniswap-v3` example, which I was also able to recreate the original bug with.

### DIFF
--- a/packages/core/src/sync/index.ts
+++ b/packages/core/src/sync/index.ts
@@ -997,8 +997,15 @@ export async function* localHistoricalSyncGenerator({
   // Cursor to track progress.
   let fromBlock = hexToNumber(syncProgress.start.number);
 
-  // Handle a cache hit by fast forwarding and potentially exiting
-  if (syncProgress.current !== undefined) {
+  /**
+   * Handle a cache hit by fast forwarding and potentially exiting
+   */
+  if (
+    syncProgress.current !== undefined &&
+    (syncProgress.cached === undefined ||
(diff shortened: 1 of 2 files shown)


packages/core/src/sync/index.ts
```diff
@@ -1026,7 +1035,7 @@ export async function* localHistoricalSyncGenerator({
      * time spent syncing ≈ time before indexing function feedback.
      */
     const interval: Interval = [
-      fromBlock,
+      Math.min(fromBlock, hexToNumber(historicalLast.number)),
```
kyscott18, line 1041: This wasn't required to fix the bug, and may be redundant, but it is a nice check to make sure the interval is always ordered correctly

0xOlias (approved):

Issue #1128, opened 2024-10-03
bug: request transaction receipts only for filters with `includeTransactionReceipt: true`


kyscott18:
### Description
Transaction receipts should only be fetched when a specific tx hash corresponds to a filter with `includeTransactionReceipts: true`.

The current logic is not granular enough and is wasteful.

https://github.com/ponder-sh/ponder/blob/5402ddcf3c751b979745c11afd6c56856e49ea68/packages/core/src/sync-realtime/index.ts#L492-L504

khaidarkairbek:
Added the fix in the pull request above, would be perfect to get a feedback on the logic.

PR #1143, opened 2024-10-08
Fetching receipts only for filters with includeTransactionReceipt: true

khaidarkairbek:
In sync-realtime, added filtering for transactions whose tx receipts need to be fetched. Relates to #1128 issue

### DIFF
--- a/packages/core/src/sync-realtime/index.ts
+++ b/packages/core/src/sync-realtime/index.ts
@@ -453,87 +453,60 @@ export const createRealtimeSync = (
       chainId: args.network.chainId,
     });
 
-    /**
-     * key: matched transaction hashes
-     * value: `true` if the corresponding receipt should be fetched
-     */
-    const transactionHashes = new Map<Hash, boolean>();
+    const requiredTransactions = new Set<Hash>();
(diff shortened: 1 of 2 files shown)


kyscott18 (approved):
This looks great, nice work.

packages/core/src/sync-realtime/index.ts
```diff
@@ -453,34 +453,71 @@ export const createRealtimeSync = (
...
 
-    // Remove logs that don't match a filter
-    logs = logs.filter((log) =>
-      logFilters.some((filter) => isLogFilterMatched({ filter, block, log })),
-    );
+    /**
+     * key: matched transaction hashes
+     * value: `true` if the corresponding receipt should be fetched
+     */
+    const transactionHashes = new Map<Hash, boolean>();
```
0xOlias, line 460: I think this would be more readable as two `Set<Hash>` named `requiredTransactionInputs` and `requiredTransactionReceipts`

khaidarkairbek: Agree, was wondering whether duplicating data in two sets would make sense for this purpose.

0xOlias (approved):

PR #1144, opened 2024-10-08
Fixup types and metrics

3commascapital:


packages/core/src/database/index.ts
```diff
@@ -141,67 +141,69 @@ export const createDatabase = (args: {
...
-    //   user: createSqliteDatabase(userFile),
-    //   readonly: createReadonlySqliteDatabase(userFile),
-    //   sync: createSqliteDatabase(syncFile),
-    // };
+    driver = {
+      internal: createPool(poolConfig),
+      user: createPool(poolConfig),
+      readonly: createReadonlyPool(poolConfig),
+      sync: createPool(poolConfig),
+    };
```
0xOlias, line 160: PGlite is single user/connection. And at the moment, it doesn't seem to support opening a database in readonly mode while another database is writing. So for now, we'll need to combine the `internal`, `readonly` and `user` drivers into one that we'll call `user` (those all touch the same tables). The sync can still be its own database and driver.

Here's what I'm thinking:

```ts
import { PGlite } from "pglite";

// ...

const userDir = path.join(args.databaseConfig.directory, "public");
const syncDir = path.join(args.databaseConfig.directory, "sync");

const userPglite = await PGlite.create(userDir)
const syncPglite = await PGlite.create(syncDir)

driver = {
  user: userPglite,
  sync: syncPglite,
}
```

The `driver` object should contain the low-level drivers so that we can safely tear them down them in the `kill` function.

packages/core/src/database/index.ts
```diff
@@ -141,67 +141,69 @@ export const createDatabase = (args: {
...
 
     qb = {
       internal: new HeadlessKysely({
         name: "internal",
         common: args.common,
-        dialect: new SqliteDialect({ database: driver.user }),
+        dialect: new PostgresDialect({ pool: driver.internal }),
```
0xOlias, line 166: Now, when we create the query builder (Kysely) objects, we can use the `kysely-pglite` driver (https://github.com/dnlsandiego/kysely-pglite/tree/main).

We should use the `user` driver for the `internal`, `user`, and `readonly` qb objects. I think this `qb` object should look identical to the one below for postgres except for the `dialect` option.

PR #1146, opened 2024-10-09
support ipv6 by default

sinasab:
Since updating the internal server to hono, it seems this behavior has changed; looks like ponder always defaults to using `0.0.0.0` as the hostname when unspecified, rather than adaptively choosing between that and `::` as node did. See [here](https://github.com/honojs/node-server/blob/main/src/server.ts#L24) for where this behavior lives.

I added a small fix that defaults the `hostname` arg to `::` when unset, and a test that validates this behavior.

As with before, this can easily be worked around via using the --hostname flag (eg. `ponder start --hostname=::`)

### DIFF
--- a/.changeset/curvy-carrots-cough.md
+++ b/.changeset/curvy-carrots-cough.md
@@ -1,0 +1,5 @@
+---
+"@ponder/core": patch
+---
+
+Setting default hostname to "::" to support ipv6
--- a/docs/pages/docs/api-reference/ponder-cli.mdx
+++ b/docs/pages/docs/api-reference/ponder-cli.mdx
@@ -43,7 +43,7 @@ Start the development server with hot reloading
 
 Options:
   -p, --port <PORT>          Port for the web server (default: 42069)
-  -H, --hostname <HOSTNAME>  Hostname for the web server (default: "0.0.0.0" or "::")
+  -H, --hostname <HOSTNAME>  Hostname for the web server (default: "::")
   -h, --help                 display help for command
 ```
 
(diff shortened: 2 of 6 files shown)


0xOlias:
Thanks for opening and apologies for the regression.

Rather than update our default to override theirs, should we try to upstream this change into `@hono/node-server`? @kyscott18 has contributed there before.

https://github.com/honojs/node-server/blob/main/src/server.ts#L24

```diff
-  server.listen(options?.port ?? 3000, options.hostname ?? '0.0.0.0', () => {
+  server.listen(options?.port ?? 3000, options.hostname, () => {
```

The property that Ponder's `hostname` argument behaves just like the Node.js default seems desirable to me. Let me know what you think.

sinasab:
> should we try to upstream this change

Nice, I like that idea. I'd assumed that getting the change into ponder would be significantly quicker than upstreaming it, but you inspired me to submit the upstream PR anyway-- it's here: https://github.com/honojs/node-server/pull/206

> The property that Ponder's hostname argument behaves just like the Node.js default seems desirable to me. Let me know what you think.

Yes definitely, when it works 😜 in light of it not working as expected though, I was thinking the tradeoff of not default-supporting ipv4-only platforms would be stomacheable, esp since most users will be running ponder on railway where private networking is recommended but is breaking opaquely out of the box.

I'm not sure how quickly the upstream PR will be addressed, but once it's landed I can bump the version to get it into ponder/core as well. In the meantime, feel free to leave this PR open or close it, whatever seems appropriate to you 🫡

sinasab:
Upstream change got merged! Closing this in favor of https://github.com/ponder-sh/ponder/pull/1155

PR #1149, opened 2024-10-09
fix status containing unindexed blocks

### LINKED ISSUE #1148: `status` out of sync
The `status` table can show state that has been synced but not indexed.

This bug only applies to the realtime portion of the chain, because of the asynchronous nature of the queue in `run.ts`

kyscott18:
Closes #1148.


0xOlias (approved):
LGTM

Issue #1150, opened 2024-10-09
Validate rpc response consistency


kyscott18:
## Description
Before further processing, we should validate that all referenced data is present. All `blockHash`s and `transactionHash`s should have a corresponding block and transaction. This should happen in both `sync-historical` and `sync-realtime`.

## Related 
Several users have reported that `transaction` is `undefined` in their ponder indexing functions. After further investigation this is a result of the database being corrupted, with a `log.transactionHash` referencing a transaction that doesn't exist

khaidarkairbek:
What should be the proper handling if the log references the transaction that is not in the block? Should it throw an error?

kyscott18:
> What should be the proper handling if the log references the transaction that is not in the block? Should it throw an error?

Great question - I think the `sync-historical` logic should log an error containing exactly what is inconsistent and then fail. `sync-realtime` should log a similar warning, and then enter this retry loop that is already used to handle some rpc-related errors.

 https://github.com/ponder-sh/ponder/blob/8790396ad61e49d284e838572c347b7c97ec9898/packages/core/src/sync-realtime/index.ts#L644-L654
 
The reason for the different behavior in `sync-historical` vs `sync-realtime` is that json rpc providers are generally less reliable (because of load balancing or reorg handilng logic or whatever reason) when the requested data is closer to the tip of the chain but eventually able to serve the correct data after retrying. Farther back in the historical range the providers are more reliable but errors are persistent.

khaidarkairbek:
I see, can i take over this issue btw?

kyscott18:
Yes, that would be great. I think you should start with the `sync-realtime` cause it should be more straightforward than the `sync-historical` and make a draft pr when you finish that part.

kyscott18:
@khaidarkairbek Here's another spot in the same function we would want to update that does some rpc response validation and throws and error:
https://github.com/ponder-sh/ponder/blob/6d45c52ed32791be782181f4c04fd63fe5e01d9a/packages/core/src/sync-realtime/index.ts#L391-L395

PR #1152, opened 2024-10-09
Validate RPC log/callTrace response

khaidarkairbek:
Added a check to validate whether filtered logs and calltraces point to valid transactions in the block for realtime-sync engine. Relates to #1150 issue.


khaidarkairbek:
@kyscott18 Added a check after filtering transactions, was wondering if this would be sufficient for realtime-sync

packages/core/src/sync-realtime/index.ts
```diff
@@ -497,6 +497,13 @@ export const createRealtimeSync = (
       requiredTransactions.has(hash),
     );
 
+    // Validate that filtered logs/callTraces point to valid transaction in the block
+    if (transactions.length !== requiredTransactions.size) {
+      throw new Error(
+        `Received callTrace/log in the block referencing transaction that is not in current head block '${block.hash}'`,
+      );
```
0xOlias, line 504: Not blocking, but I'd advocate for a more specific error message which includes the suspicious transaction hashes. That would make it much easier for us to submit bug reports to chains / RPC providers.

Also, what if the lengths are different because we have an _extra_ required transaction? That scenario would suggest a Ponder bug rather than a faulty RPC.

khaidarkairbek: 1) With regards to error message, do you think it would be better to include the list of suspicious tx hashes in the message? 
2) We would expect requiredTransactions size to be larger than transactions length only in the case if some logs or callTraces that is matched with filter are pointing to more transactions, than there are transactions filtered by those logs/traces. 
`const transactions = block.transactions.filter(({ hash }) =>
      requiredTransactions.has(hash),
    );`

kyscott18: 1. yes
2. I think the transaction validation should be more precise. Something like:
```ts
const blockTransactionsHashes = new Set(
  block.transactions.map((t) => t.hash),
);
for (const hash of Array.from(requiredTransactions)) {
  if (blockTransactionsHashes.has(hash) === false) {
    // throw error, include `hash`
  }
}
```

I also think it would be nice to have something that compares `blockHash` in `logs` and `callTraces` to `block.hash`

khaidarkairbek:
@kyscott18 @0xOlias I have added the validation checks in sync-realtime for logs response: 
 - log.transactionHash in block.transactions
 - log.blockHash === block.hash
and for callTraces response: 
 - trace.transactionHash in block.transactions
 - trace.blockHash === block.hash (already existed)
 
Similarly, for sync-historical I have added the following checks: 
 - log.transactionHash in block.transactions
 - trace.transactionHash in block.transactions
 
Since logs and traces are fetched by batches in the sync-historical, it's not feasible to check if the logs and traces returned by the rpc point to the right block. I suggest adding the check if the block that log/trace point to is in the range of blocks fetched from rpc.

packages/core/src/sync-realtime/index.ts
```diff
@@ -393,6 +393,15 @@ export const createRealtimeSync = (
           `Detected invalid '${args.network.name}' eth_getLogs response.`,
         );
       }
+
+      // Check that logs refer to the correct block
+      for (const log of logs) {
```
kyscott18, line 398: This is great

kyscott18 (approved):

PR #1155, opened 2024-10-10
chore: bumping hono/node-server to support ipv6

sinasab:
See context here: https://github.com/ponder-sh/ponder/pull/1146

Upstream change to hono/node-server was included @ https://github.com/honojs/node-server/pull/206

This PR bumps the version of hono/node-server to include the upstream change. I also added a test to ensure ipv6 works, which should help avoid future regressions as well!

### DIFF
--- a/.changeset/stupid-cooks-wash.md
+++ b/.changeset/stupid-cooks-wash.md
@@ -2,4 +2,4 @@
 "@ponder/core": patch
 ---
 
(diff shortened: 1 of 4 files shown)


kyscott18 (approved):

0xOlias (approved):
LGTM, great work and thanks for the writing the test.

PR #1173, opened 2024-10-16
Fix `block.hash` and `log.blockHash` comparaison during sync

KONFeature:
The hash comparaison was done on the hex string values, thus failing in some case due to case sensitivity or leading zero.

For our use case we where indexing with alchemy + envio rpcs, and every version after the v0.6.11 was failing to index our events due to this condition and case sensitivity.

Using BigInt wrapping ensure we are comparing the real hex value (it's also possible to use `hexToBytes(...)` or `hexToBigint(...)`from viem, but it would be more consuming in term of memory and cpu)

### DIFF
--- a/packages/core/src/sync-realtime/index.ts
+++ b/packages/core/src/sync-realtime/index.ts
@@ -491,7 +491,7 @@ export const createRealtimeSync = (
 
       // Check that logs refer to the correct block
       for (const log of logs) {
-        if (log.blockHash !== block.hash) {
+        if (BigInt(log.blockHash) !== BigInt(block.hash)) {
           throw new Error(
             "Detected invalid eth_getLogs response. `log.blockHash` does not match requested block hash.",
           );


kyscott18:
This check in the realtime service is meant to validate against invalid rpc responses.

https://github.com/ponder-sh/ponder/blob/e6a0ab8e696bee7135e93fdc74b8e8792b341681/packages/core/src/sync-realtime/index.ts#L492-L499

The execution specification says that all hash responses should be lowercase. So any non-lowercase hash values are invalid, and the validation + error is working as expected. Seems like one of the rpc providers is non-conforming or has a bug.

https://github.com/ethereum/execution-apis/blob/4140e528360fea53c34a766d86a000c6c039100e/src/schemas/base-types.yaml#L63-L66

KONFeature:
Ok, so a PR to disable this check would be ok in that case?

PR #1174, opened 2024-10-16
Feature: network option to disable rpc request validation

KONFeature:
Related to #1173  

Add a network option to disable rpc request validation for the comparaison between block hash of logs and traces results


0xOlias:
Hi @KONFeature thanks for opening, appreciate the quick action.

We just merged #1175 and released it in `0.6.16`. That PR adds the specific block hash values to the validation error message, which should make it much easier to identify what's going on + report the bug to the RPC provider. My hunch is that this is a casing issue.

For context - if we added custom logic to accommodate every non-compliant RPC provider behavior, we would likely introduce a bug OR ingest/cache invalid data. We've found that the only sane approach is to stay pretty disciplined about adhering to the specs. Hope you can understand our stance on this.

Regarding this config option - if we allow invalid data, a different validation (further along in the control flow) will likely fail and crash the app anyway.

PR #1175, opened 2024-10-16
fix: validation error messages

0xOlias:


packages/core/src/sync-historical/index.ts
```diff
@@ -242,7 +242,7 @@ export const createHistoricalSync = async (
         undefined
       ) {
         throw new Error(
-          `Detected inconsistent RPC responses. Log with transaction hash ${log.transactionHash} not found in block ${block.hash}.`,
+          `Detected inconsistent eth_getLogs response. 'log.transactionHash' ${log.transactionHash} not found in 'block.transactions' ${block.hash}`,
```
kyscott18, line 245: The reason I called in "inconsistent RPC response" instead of "inconsistent eth_getLogs response" is that it's entirely possible that eth_getBlockByNumber is the method returning the bad data.

packages/core/src/sync-realtime/index.ts
```diff
@@ -493,7 +493,7 @@ export const createRealtimeSync = (
       for (const log of logs) {
         if (log.blockHash !== block.hash) {
           throw new Error(
-            "Detected invalid eth_getLogs response. `log.blockHash` does not match requested block hash.",
+            `Detected inconsistent eth_getLogs response. 'log.blockHash' ${log.blockHash} does not match 'block.hash' ${block.hash}`,
```
kyscott18, line 496: In this case, I think it's safe to say that this request is "invalid" because the eth_getLogs is called with `blockHash`, so it's expected the result matches

kyscott18 (approved):
Looks great

PR #1205, opened 2024-10-30
Implement types for `debug_` rpc namespace

kyscott18:


packages/core/src/utils/debug.ts
```diff
@@ -0,0 +1,110 @@
...
+  /** The topics of the log. */
+  topics: LogTopic[];
+  /** The data of the log. */
+  data: Hex;
+  /** The position of the log relative to subcalls within the same trace. */
+  position: number;
+};
+
+/** The configuration for the call tracer. */
+type CallConfig = {
```
0xOlias, line 66: I think I've seen this named "TracerConfig" before

PR #1207, opened 2024-10-30
feat: use more accurate cache size

### LINKED ISSUE #1171: feat: improved accuracy of `indexingCacheMaxBytes`
## Background
With #929, Ponder uses an in-memory cache during the historical backfill to improve the performance of indexing functions. This cache occasionally flushes to the database to avoid an out of memory error. 

The cache size is determined by `indexingCacheMaxBytes`, which is currently computed based on `os.freemem()`.
https://github.com/ponder-sh/ponder/blob/e6a0ab8e696bee7135e93fdc74b8e8792b341681/packages/core/src/common/options.ts#L108-L116

## Description
**Having a more correct (larger) cache size can significantly improve performance.**

The current calculation isn't very accurate. This is reproducible by running a Ponder app with `ponder dev -v` and looking for the log. `indexingCacheMaxBytes` on a local machine should almost certainly be multiple GB.

```sh
7:29:16 PM DEBUG indexing   Using a 64 MB indexing cache
```

Solving this issue involves a research project into which memory and then a small pr implementing the new logic, which would be (maybe) only scoped to `common/option.ts`. Ideally, we can determine the memory size at which node is going to OOM and use some percentage of that as the cache size.

### Notes
https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-mib is what we currently use to increase memory for some hosting environments.

kyscott18:
Closes #1171

### DIFF
--- a/packages/core/src/common/options.ts
+++ b/packages/core/src/common/options.ts
@@ -1,5 +1,5 @@
-import os from "node:os";
 import path from "node:path";
+import v8 from "node:v8";
 import type { CliOptions } from "@/bin/ponder.js";
 import type { LevelWithSilent } from "pino";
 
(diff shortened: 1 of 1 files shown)


packages/core/src/common/options.ts
```diff
@@ -103,11 +103,18 @@ export const buildOptions = ({ cliOptions }: { cliOptions: CliOptions }) => {
 
     factoryAddressCountThreshold: 1_000,
 
-    // os.freemem() / 4, bucketed closest to 64, 128, 256, 512, 1024, 2048 mB
+    // v8.getHeapStatistics().heap_size_limit / 4, bucketed closest to 128, 256, 512, 1024, 2048 mB
```
0xOlias, line 106: LGTM, though I might move this to it's own line above the return statement like we do with `port`.

0xOlias (approved):

0xOlias:
Related idea - in the past week, we've seen a couple users confused that the database is missing data because none has been flushed yet. I wonder if we should flush more frequently when using `ponder dev`, or at least do a better job documenting the flush behavior. Can address in a follow-up PR.

0xOlias (approved):

Issue #1198, opened 2024-10-28
feat: filter types


kyscott18:
Tracking issue for all filter related features.

Filters should be deliberately designed, expressive, and modeled after the fundamentals of the Ethereum blockchain, not necessarily the Ethereum standard JSON-RPC. Raw blockchain data that is matched by a filter becomes an event, the arguments to an indexing function.

### Proposed filter types

```ts
type LogFilter = {
  type: "log";
  chainId: number;
  address: Address | Address[] | Factory | undefined;
  topics: LogTopic[];
  fromBlock: number;
  toBlock: number | undefined;
  includeTransactionReceipts: boolean;
};

type BlockFilter = {
  type: "block";
  chainId: number;
  interval: number;
  offset: number;
  fromBlock: number;
  toBlock: number | undefined;
};

type TransferFilter = {   
  type: "transfer";   
  chainId: number;
  fromAddress: Address | Address[] | Factory | undefined;
  toAddress: Address | Address[] | Factory | undefined;
  fromBlock: number;
  toBlock: number | undefined;   
  includeTransactionReceipt: boolean;
};

type TransactionFilter = {
  type: "transaction";
  chainId: number;
  fromAddress: Address | Address[] | Factory | undefined;
  toAddress: Address | Address[] | Factory | undefined;
  functionSelectors: Hex | Hex[] | undefined;
  includeInner: boolean;
  includeFailed: boolean;
  callType: "call" | "staticcall" | "delegatecall" | "selfdestruct" | "create" | "create2" | "callcode" | undefined;
  fromBlock: number;
  toBlock: number | undefined;  
  includeTransactionReceipt: boolean;
};
```

Few things I'm not sure about:
- is `includeTransactionReceipt` necessary? Does in need to include the `logs` property?
- should transactions and traces be separate filters?

### `ponder.config.ts`

The config should provide high level abstractions, such as `contracts` and `accounts`. These high-level `sources` also have the ability to post-process a filter result and add decoded args, etc. 

### `factory()`

I think it could be nice to have a helper function that can be used in any `address` field.

```ts
import { createConfig, factory } from "@ponder/core";

export default createConfig({
  contracts: {
    UniswapV3Pool: {
      network: "mainnet",
      abi: UniswapV3PoolAbi,
      address: factory({ 
        address: "0x1F98431c8aD98523631AE4a59f267346ea31F984",
        event: getAbiItem({ abi: UniswapV3FactoryAbi, name: "PoolCreated" }),
[...]

PR #1210, opened 2024-11-01
accounts + native transfers

### LINKED ISSUE #1157: [Feature] Native transfers
We should add an intuitive high-level API for native transfers, such that you can write indexing functions that fire when a specific account (or list of accounts specified by a factory) sends or receives any amount of the native token.

kyscott18:
Closes #1157 #1198 #1205 

### `sync`
- [x] debug rpc types
- [x] `TransactionFilter` type
- [x] `TransferFilter` type
- [x] `TraceFilter` type
- [x] `_debug_traceBlockByNumber` + `_debug_traceBlockByHash`
- [x] `TransactionFilter` fragments
- [x] `TransferFilter` fragments
- [x] `TraceFilter` fragments
- [x] `isTransactionFilterMatched`
- [x] `isTransferFilterMatched`
- [x] `isTraceFilterMatched`
- [x] create `traces` table, drop `callTraces`
- [x] `decodeEvents`
- [x]  `buildEvents`
- [x] `getEvents`
### `sync-historical`
- [x] `syncTrace`
- [x] `syncTransaction`
- [x] order traces `checkpoint.eventIndex`
- [x] determine failed traces
- [x] detect inconsistent traces
- [x] factory support
- [x] avoid refetching with `traceCache`
### `sync-realtime`
- [x] `fetchBlockEventData`
- [x] `handleBlock`
- [x] factory support
### API
- [x] `AccountSource`
- [x] `ponder.config.ts`
- [x] `buildConfigAndIndexingFunctions`


packages/core/src/sync-historical/index.ts
```diff
@@ -513,15 +717,29 @@ export const createHistoricalSync = async (
                       break;
                     }
 
-                    case "callTrace":
+                    case "transaction":
+                    case "transfer": {
+                      const blocks = Array.from(
+                        { length: interval[1] - interval[0] + 1 },
+                        (_, i) => interval[0] + i,
+                      );
```
kyscott18, line 508: I would feel a little better about this if we just used an array and a for loop

packages/core/src/sync-historical/index.ts
```diff
@@ -291,103 +293,305 @@ export const createHistoricalSync = async (
...
 
-      if (block.hash !== callTrace.blockHash) {
+    for (const hash of transactionHashes) {
+      // Validate that trace points to a transaction inside a block
+      if (block.transactions.find((t) => t.hash === hash) === undefined) {
         throw new Error(
-          `Detected inconsistent RPC responses. 'trace.blockHash' ${callTrace.blockHash} does not match 'block.hash' ${block.hash}`,
+          `Detected inconsistent RPC responses. 'trace.transactionHash' ${hash} not found in 'block.transactions' ${block.hash}`,
         );
       }
```
kyscott18: nice

packages/core/src/sync-historical/index.ts
```diff
@@ -291,103 +293,305 @@ export const createHistoricalSync = async (
...
-      toBlock: interval[1],
-    }).then(
-      (traces) =>
-        traces.flat().filter((t) => t.type === "call") as SyncCallTrace[],
-    );
-
-    if (isKilled) return;
+    let traces = await _debug_traceBlockByNumber(args.requestQueue, {
+      blockNumber: toHex(blockNumber),
+    });
```
kyscott18: When I was evaluating the debug rpc methods, I notice that `traceBlockByHash` was often way faster than `traceBlockByNumber`. It would require a different algorithm because block hash isn't immediately available, but something we should try before shipping

packages/core/src/sync-historical/index.ts
```diff
@@ -291,103 +297,84 @@ export const createHistoricalSync = async (
...
-      toAddress,
-      fromBlock: interval[0],
-      toBlock: interval[1],
-    }).then(
-      (traces) =>
-        traces.flat().filter((t) => t.type === "call") as SyncCallTrace[],
-    );
+    const syncTraces = await _debug_traceBlockByNumber(args.requestQueue, {
+      blockNumber: toHex(blockNumber),
+    });
```
kyscott18, line 306: We should start to think about how we can avoid requesting the same block twice if there are multiple transfer and transaction filters

packages/core/src/sync-realtime/index.ts
```diff
@@ -587,19 +683,38 @@ export const createRealtimeSync = (
...
+    //         requiredTransactionReceipts.add(callTrace.transactionHash);
+    //       }
+    //     }
+    //   }
+
+    //   return isCallTraceMatched;
+    // });
+
+    // Remove traces that dont match transaction or transfer filter
+    traces = traces.filter((trace) => {
```
kyscott18, line 702: I think we need to do a depth first search over all the traces, either in this block or above where `traces` is declared. 

It performs three functions:
- processes child traces (each call can have several sub-calls)
- orders traces for `checkpoint.eventIndex`
- determine failed traces

See `sync-historical` https://github.com/ponder-sh/ponder/blob/2117cec2f8e477d5446ab0242b07a82d054f50e1/packages/core/src/sync-historical/index.ts#L324-L368

packages/core/src/sync-realtime/index.ts
```diff
@@ -643,15 +758,15 @@ export const createRealtimeSync = (
...
+    // callTraces = callTraces.filter(
+    //   (trace) => revertedTransactions.has(trace.transactionHash) === false,
+    // );
 
     return {
       block,
       logs,
       factoryLogs,
-      callTraces,
+      traces,
```
kyscott18, line 769: Should this type be the same as ?? https://github.com/ponder-sh/ponder/blob/2117cec2f8e477d5446ab0242b07a82d054f50e1/packages/core/src/sync-historical/index.ts#L313-L317

packages/core/src/sync-realtime/index.ts
```diff
@@ -513,32 +582,59 @@ export const createRealtimeSync = (
...
 
-    const shouldRequestTraces = callTraceFilters.length > 0;
+    const shouldRequestTraces =
+      transactionFilters.length > 0 || transferFilters.length > 0;
 
-    let callTraces: SyncCallTrace[] = [];
+    let traces: SyncTrace[] = [];
     if (shouldRequestTraces) {
-      const traces = await _trace_block(args.requestQueue, {
+      traces = await _debug_traceBlockByNumber(args.requestQueue, {
```
kyscott18, line 590: Should use `_debug_traceBlockByHash`. It's more robust when reorgs occur because it unambiguously refers to the exact block

packages/core/src/sync-realtime/index.ts
```diff
@@ -571,44 +628,74 @@ export const createRealtimeSync = (
...
-      requiredTransactions.has(hash),
-    );
+    const transactions = block.transactions.filter((transaction) => {
+      let isMatched = requiredTransactions.has(transaction.hash);
+      for (const filter of transactionFilters) {
+        if (isTransactionFilterMatched({ filter, block, transaction })) {
+          if (filter.includeReverted === false) {
+            // or TODO: includeTransactionReceipt
+            requiredTransactionReceipts.add(transaction.hash);
+            return true;
```
kyscott18, line 692: Are you sure we want to `return true` here?

packages/core/src/sync-realtime/index.ts
```diff
@@ -233,10 +276,22 @@ export const createRealtimeSync = (
...
+      for (const filter of transactionFilters) {
+        if (isTransactionFilterMatched({ filter, block, transaction })) {
+          matchedFilters.add(filter);
+          isMatched = true;
+        }
+      }
+      return isMatched;
+    });
+
     transactions = transactions.filter((t) => transactionHashes.has(t.hash));
```
kyscott18, line 295: I don't think we need this line anymore because of line 285

packages/core/src/sync-realtime/index.ts
```diff
@@ -233,10 +276,22 @@ export const createRealtimeSync = (
...
     transactionReceipts = transactionReceipts.filter((t) =>
```
kyscott18, line 296: We should revisit this once we implement the `includeTransactionReceipt` feature

packages/core/src/sync-realtime/index.ts
khaidarkairbek: Yes, since we have matched the transaction filter and we have also added it to the required transactionReceipts, we can return early.

packages/core/src/sync-realtime/index.ts
khaidarkairbek: Agreed, that was a typo

packages/core/src/sync-realtime/filter.ts
```diff
@@ -189,6 +189,13 @@ export const isTraceFilterMatched = ({
     return false;
   }
 
+  if (
+    filter.includeReverted === false &&
+    (trace.revertReason !== undefined || trace.error !== undefined)
+  ) {
+    return false;
+  }
```
kyscott18, line 197: It's not super clear, this function returns true for the case "should this trace be saved in the database" and not "does this trace exactly match the filter such that it will cause an indexing function to run". The reason for the distinction is that we don't want the cache to be completely useless when the filter type slightly changes. For that reason this check should be removed. 

In this example, if the user decides that they want to include reverted traces after they have already fully synced their app, every block will need to be resynced.

Maybe we could add some comments acknowledging the properties that are ignored.

packages/core/src/sync-realtime/filter.ts
```diff
@@ -234,6 +241,10 @@ export const isTransferFilterMatched = ({
     return false;
   }
 
+  if (trace.revertReason !== undefined && trace.error !== undefined) {
+    return false;
+  }
```
kyscott18, line 246: Same as above

kevinfaveri:
Hello guys; I have a pressing timeline on getting something for native transfers and would like to explore using this here... Do you think it will be ready within next 1/2 weeks? Any timeline in mind?

Thank you in advance.

kyscott18:
> Hello guys; I have a pressing timeline on getting something for native transfers and would like to explore using this here... Do you think it will be ready within next 1/2 weeks? Any timeline in mind?

We're hoping to get a prerelease out in the next few days, but don't want to make any promises on when we will release v0.8.

packages/core/src/sync-realtime/index.ts
```diff
@@ -173,6 +208,36 @@ export const createRealtimeSync = (
       }
     }
 
+    const isFactoryAddressMatched = ({
```
kyscott18, line 211: Could get rid of this if we update `is[type]FilterMatched` to handle factories

packages/core/src/sync-realtime/index.ts
```diff
@@ -571,44 +687,70 @@ export const createRealtimeSync = (
...
     });
 
     ////////
     // Transactions
     ////////
 
-    const transactions = block.transactions.filter(({ hash }) =>
-      requiredTransactions.has(hash),
-    );
+    const transactions = block.transactions.filter((transaction) => {
```
0xOlias, line 744: My understanding is that the `fetchBlockEventData` function should be very conservative with filtering out data, because only the `handleBlock` function has access to the newest child addresses.

I'm not quite able to follow whether or not we're accomplishing that as expected here, so worth a double check. Seems like a plausible way to introduce a missing event regression that our tests might not catch.

@kyscott18 and @khaidarkairbek  I think once we're ready to ship `0.8`, we should do one final comprehensive review of these two functions.

packages/core/src/sync-store/index.ts
```diff
@@ -566,64 +523,155 @@ export const createSyncStore = ({
...
+          qb.where("error", "is", null),
+        )
+        .$if(filter.callType !== undefined, (qb) =>
+          qb.where("type", "=", filter.callType!),
+        )
+        .$if(filter.functionSelector !== undefined, (qb) => {
+          if (Array.isArray(filter.functionSelector)) {
+            return qb.where((eb) =>
+              eb.or(
+                (filter.functionSelector as Hex[]).map((fs) =>
```
0xOlias, line 659: I wonder if we should add a `functionSelector` column + index to the database - I could see this `like` performing poorly on large tables.

packages/core/src/sync-store/migrations.ts
```diff
@@ -1082,6 +1083,144 @@ AND ponder_sync."rpcRequestResults"."blockNumber" <= 9223372036854775807;
...
+        .addColumn("checkpoint", "varchar(75)", (col) => col.notNull())
+        .addColumn("type", "text", (col) => col.notNull())
+        .addColumn("transactionHash", "varchar(66)", (col) => col.notNull())
+        .addColumn("blockNumber", "numeric(78, 0)", (col) => col.notNull())
+        .addColumn("blockHash", "varchar(66)", (col) => col.notNull())
+        .addColumn("from", "varchar(42)", (col) => col.notNull())
+        .addColumn("to", "varchar(42)")
+        .addColumn("gas", "numeric(78, 0)", (col) => col.notNull())
+        .addColumn("gasUsed", "numeric(78, 0)", (col) => col.notNull())
+        .addColumn("input", "text", (col) => col.notNull())
```
0xOlias, line 1108: Are you sure this is not null? I've definitely seen traces with an "empty" input, but maybe it's an empty string.

packages/core/src/sync-store/migrations.ts
```diff
@@ -1082,6 +1083,144 @@ AND ponder_sync."rpcRequestResults"."blockNumber" <= 9223372036854775807;
...
+        .addColumn("output", "text")
+        .addColumn("error", "text")
+        .addColumn("revertReason", "text")
+        .addColumn("value", "numeric(78, 0)")
+        .addColumn("index", "integer", (col) => col.notNull())
+        .addColumn("subcalls", "integer", (col) => col.notNull())
+        .execute();
+
+      // `getEvents` benefits from an index on
```
0xOlias, line 1117: Hmm, this definitely feels over-indexed, but I'm happy to leave it as-is until we tackle the sync tables overhaul.

packages/core/src/sync/events.ts
```diff
@@ -423,6 +564,71 @@ export const decodeEvents = (
         break;
       }
 
+      case "account": {
+        switch (source.filter.type) {
+          case "transaction": {
+            // TODO(kyle) what if toAddress and fromAddress are both undefined?
```
0xOlias, line 570: Not sure what you mean here

packages/core/src/types/eth.ts
```diff
@@ -193,41 +191,48 @@ export type TransactionReceipt = {
...
+  /** Error message, if any. */
+  error?: string;
+  /** Why this call reverted, if it reverted. */
+  revertReason?: string;
+  /** Value transferred. */
+  value: bigint | null;
+  /** Index of this trace in the transaction. */
+  traceIndex: number;
+  /** Number of subcalls. */
+  subcalls: number;
```
0xOlias, line 227: Won't this still count non-`CALL` traces? eg

```ts
  /** Number of child traces produced during the execution of this trace. */
  subtraces: number;
```

packages/core/src/utils/rpc.ts
```diff
@@ -147,59 +147,144 @@ export const _eth_getTransactionReceipt = (
...
-    .then((traces) => traces as unknown as SyncTrace[]);
+      method: "debug_traceBlockByHash",
+      params: [hash, { tracer: "callTracer" }],
+    })
+    .then((traces) => {
+      const result: SyncTrace[] = [];
+      let index = 0;
+      // all traces that weren't included because the trace has an error
+      // or the trace's parent has an error, mapped to the error string
+      const failedTraces = new Map<
```
0xOlias, line 244: Seems fine but I think this Map will contain frames across multiple transactions within the block

packages/core/src/sync-store/migrations.ts
kyscott18: it definitely felt like overkill but not sure which ones should be left out

packages/core/src/sync-store/migrations.ts
kyscott18: https://github.com/alloy-rs/alloy/blob/5fcbb2d87100923f4274a13e3ddb54cde3bbd681/crates/rpc-types-trace/src/geth/call.rs#L24

packages/core/src/sync-realtime/filter.ts
```diff
@@ -60,70 +75,231 @@ export const isLogFilterMatched = ({
...
+      return false;
+    }
+  }
+
+  if (filter.callType !== trace.type) {
+    return false;
+  }
+
+  if (
+    isValueMatched(filter.functionSelector, trace.input.slice(0, 10)) === false
```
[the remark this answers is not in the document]
kyscott18: `isValueMatched` handles casing

packages/core/src/sync/events.ts
kyscott18: How do we know where do display `{sourceName}:transaction:from` or `${sourceName}:transaction:to` if the user creates an account with `{ address: undefined }`. I think we should actually validate against this

packages/core/src/sync-realtime/index.ts
kyscott18: Yes, it is pretty complex. I've reread it a couple of times. We want to filter out as much as possible so that dependent requests (`eth_getTransactionReceipt`) can be skipped. Otherwise we could filter out no data in `fetchBlockEventData` and only in `handleBlock`

kevinfaveri:
> > Hello guys; I have a pressing timeline on getting something for native transfers and would like to explore using this here... Do you think it will be ready within next 1/2 weeks? Any timeline in mind?
> 
> We're hoping to get a prerelease out in the next few days, but don't want to make any promises on when we will release v0.8.

@kyscott18 Hey, curious on the technical implementation here. I see you use debug RPC APIs (trace by block number). Why a simple block by number wouldn't suffice (like, by filtering msg.value > 0 as a transfer of native and using tx.origin as the "from")? I suppose it is because you want the internal calls, but what information are you missing on a native transfer that you would need the debug/trace RPC vs the normal get block by number?

I can think of one transfer that Alice > Bob > Charlie would be interpreted as Alice > Charlie; but too practical effects, on that transaction, the money went from Alice to Charlie at the end of the block confirmation, right? What is the imporstance of knowing bob in the middle of the internal call?

Is this for example some custom contract that would send part of a deposit from user to 2 users, which could create the Alice > Bob > Charlie situation? And then if you only consider in / out of block you wouldnt see that Bob got a cut too?

Issue #810, opened 2024-04-17
[Feature] Effect functions


0xOlias:
We should support an alternative kind of function that uses a similar registration / context API, but supports side effects.

vimtor:
Why can't you just call the effect on the indexer function? Would there be problems with block reorgs?

0xOlias:
> Why can't you just call the effect on the indexer function? Would there be problems with block reorgs?

A few reasons:
1. **Performance** - Side effect workloads often take a long time (e.g. CPU-bound image processing, waiting for HTTP responses from other APIs, analytical SQL queries). If these effects run in the indexing "hot loop", it can make reindexing take forever, and for high-throughput use cases it prevents the app from keeping up with chain tip. Also, these workloads can usually run in parallel.
2. **Errors** - If an unrecoverable error occurs during indexing, the Ponder instance crashes. This makes sense because the only "approved" actions within an indexing function are 1) SQL operations and 2) RPC requests like `readContract`. If either of these fail despite many retries, something is very wrong and demands the developer's immediate attention (database or RPC provider is down). With side effects, errors are much more common and are often tolerable, because the functions often call arbitrary external services. So, side effect errors probably should not crash the instance.
4. **Reorgs** - Ponder automatically handles reorgs for indexing, which is possible because we've added a few guardrails at the SQL layer. With side effects, there's no way for the framework to know what's happening, so we likely need to expose an "undo" function that will run on a reorg.
5. **State** - Today (without side effects) Ponder maintains no state between hot reloads / redeployments, because the database state can be deterministically reconstructed on demand. With side effects, you very likely want to keep track of which events have already been handled, and not re-run all side effects on each redeployment. So, we probably need new semantics for specifying when a specific side effect should run (full backfill, or only in realtime? only after finalization?). There are also some gnarly issues here for zero-downtime deployments.

With these requirements in mind, it becomes pretty clear that we need a job queue pattern for side effects, hopefully with some blockchain/Ponder secret sauce to make it really intuitive.

ind-igo:
Excited for this feature. Curious what kinds of designs you are referring to, wrt to the EVM/Ponder sauce?

tk-o:
@typedarray, I'm keen to support the Ponder team in getting the Effects feature built 🙃

I appreciate all the thinking going into making things work correctly. The nature of blockchain networks requires a lot of intellectual gymnastics to get right outcomes.

[...]

PR #1214, opened 2024-11-03
feat: add indexingEnv to distinguish between historical and realtime events in indexing handlers

andreasasprou:
This PR adds an `indexingEnv` field to the indexing context to allow indexing functions to distinguish between historical and realtime events. This enables developers to implement different logic based on whether an event is being processed during historical backfilling or realtime indexing.

Example usage in an indexing function, with my use case

```ts
ponder.on(
	"ConditionalTokens:ConditionResolution",
	async ({ event, context }) => {
		// Skip processing for historical events since conditions are already
		// backfilled during setup
		if (context.indexingEnv === "historical") {
			return;
		}

		const conditionId = event.args.conditionId;

		// Fetch current market data from Polymarket API
		// Note: While we could derive this from on-chain events, that would require
		// indexing all Polygon events. Since we're only monitoring specific traders
		// in a recent block range, the API call is more cost-effective
		const market = await getMarketByIdFromPolymarketAPI(conditionId);

		if (!market) {
			logger.error("ConditionResolution: market not found in Polymarket API", {
				conditionId,
				event,
			});
			return;
		}

		await context.db.Condition.upsert({
			id: conditionId,
		});

		await Promise.all(
			market.tokens.map((token, index) =>
				context.db.Token.upsert({
					id: BigInt(token.token_id),
					create: {
						conditionId,
						index,
						price: token.price,
					},
					update: {
						price: token.price,
					},
				}),
			),
		);
	},
);
```

### DIFF
--- a/.changeset/metal-hairs-grin.md
+++ b/.changeset/metal-hairs-grin.md
@@ -1,0 +1,5 @@
+---
+"@ponder/core": minor
+---
+
+Added indexingEnv feature to distinguish between historical and realtime events
--- a/packages/core/src/bin/utils/run.ts
+++ b/packages/core/src/bin/utils/run.ts
@@ -103,7 +103,7 @@ export async function run({
       switch (event.type) {
         case "block": {
           const result = await handleEvents(
-            decodeEvents(common, sources, event.events),
+            decodeEvents(common, sources, event.events, "realtime"),
             event.checkpoint,
           );
 
@@ -180,7 +180,7 @@ export async function run({
       end = checkpoint;
 
       const result = await handleEvents(
-        decodeEvents(common, sources, events),
+        decodeEvents(common, sources, events, "historical"),
         checkpoint,
       );
       await metadataStore.setStatus(sync.getStatus());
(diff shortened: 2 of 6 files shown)


0xOlias:
Thanks for opening!

We've received this feature request a few times, and have historically decided against it. You can read more about the reasoning [here](https://github.com/ponder-sh/ponder/issues/810#issuecomment-2450234890). Essentially, as far as we've seen the only "good" reason to need this feature is to run an effect.

That issue comment does a good job of explaining why we need a new pattern for "write" effects (send email, write data to non-ponder database, etc). Your use case is more of a "read" effect, so I can understand why it seems reasonable. Nonetheless, our long-term vision (reindex even the largest apps in <1 minute, keep up with the highest-throughput chains) precludes "read effects" within indexing functions as well, so the disciplined move here is to not add this feature.

We're still working out the long-term solution for this use case (fetching price data) - my best guess is an effect function that fetches and stores the price in an offchain table, then use a JOIN at the query layer.

PR #1215, opened 2024-11-03
replace @/generated shim with global variable

kyscott18:
Addresses https://t.me/ponder_sh/7662

Not certain this will fix it because the virtual module method was working locally as well.

### DIFF
--- a/packages/core/src/_test/setup.ts
+++ b/packages/core/src/_test/setup.ts
@@ -28,7 +28,7 @@ import type { RequestQueue } from "@/utils/requestQueue.js";
 import pg from "pg";
 import { rimrafSync } from "rimraf";
 import type { Address } from "viem";
-import { type TestContext, afterAll, vi } from "vitest";
+import { type TestContext, afterAll } from "vitest";
 import { deploy, simulate } from "./simulate.js";
 import {
   getConfig,
@@ -137,11 +137,8 @@ const defaultDatabaseServiceSetup: DatabaseServiceSetup = {
   indexing: "historical",
 };
 
-vi.mock("@/generated", async () => {
-  return {
-    instanceId: "1234",
-  };
-});
+// @ts-ignore
+globalThis.INSTANCE_ID = "1234";
 
 export async function setupDatabaseServices(
   context: TestContext,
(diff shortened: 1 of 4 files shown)


packages/core/src/_test/setup.ts
```diff
@@ -137,11 +137,8 @@ const defaultDatabaseServiceSetup: DatabaseServiceSetup = {
   indexing: "historical",
 };
 
-vi.mock("@/generated", async () => {
-  return {
-    instanceId: "1234",
-  };
-});
+// @ts-ignore
+globalThis.INSTANCE_ID = "1234";
```
0xOlias, line 141: I think we should use something like `__PONDER_INSTANCE_ID` to reduce the chance of a collision. See [Next.js example](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/setup-http-agent-env.ts).

packages/core/src/drizzle/index.ts
```diff
@@ -225,7 +219,8 @@ export const onchainTable = <
   extra: extra;
   dialect: "pg";
 }> => {
-  if (instanceId === undefined) {
+  // @ts-ignore
+  if (globalThis.INSTANCE_ID === undefined) {
```
0xOlias, line 223: It should be safe to do `const instanceId = globalThis.INSTANCE_ID` in the file scope, right?

packages/core/src/drizzle/index.ts
```diff
@@ -225,6 +219,8 @@ export const onchainTable = <
   extra: extra;
   dialect: "pg";
 }> => {
+  // @ts-ignore
+  const instanceId: string | undefined = globalThis.__PONDER_INSTANCE_ID;
```
0xOlias, line 223: Could add a `declare global` somewhere in `packages/core` to annotate this as `string | undefined`, would remove the need for `@ts-ignore` everywhere

https://stackoverflow.com/a/69429093

0xOlias (approved):
LGTM, non-blocking nit

packages/core/src/drizzle/index.ts
kyscott18: Hmm this wasn't working for me

PR #1216, opened 2024-11-03
fix: duplicate db client breaks pglite

kyscott18:


packages/core/src/bin/commands/dev.ts
```diff
@@ -79,9 +80,18 @@ export async function dev({ cliOptions }: { cliOptions: CliOptions }) {
       if (result.status === "success") {
         metrics.resetIndexingMetrics();
 
+        database = await createDatabase({
```
0xOlias, line 83: FWIW I'm pretty sure this no longer needs to be async

0xOlias (commented):
Looks good, nice fix.

I wonder if the database cleanup should now be `dev`/`start`/`serve` concern though, instead of `run`. I'm guessing to fix this for `serve`, it will hang on shutdown because we only use `runServer` there and that one doesn't do database cleanup.

kyscott18:
Ah nice catch, it looks like the `database.kill` function does two things: 

1. Updates the metadata table

https://github.com/ponder-sh/ponder/blob/f3e383fa3e00e0156ccc1c6eb404e3ad41d80c56/packages/core/src/database/index.ts#L1171-L1179

2. Tears down the underlying client

https://github.com/ponder-sh/ponder/blob/f3e383fa3e00e0156ccc1c6eb404e3ad41d80c56/packages/core/src/database/index.ts#L1181-L1197

The "updating the metadata" step should only happen when `database.setup` is called, which doesn't happen in the `serve` case. I also think it makes sense that the function that creates the database is in charge of cleaning it up. Will do another commit.

0xOlias:
> The "updating the metadata" step should only happen when `database.setup` is called, which doesn't happen in the `serve` case. I also think it makes sense that the function that creates the database is in charge of cleaning it up. Will do another commit.

Makes sense. Yea, the WIP work I have locally actually separates the database driver(s) (and associated cleanup logic) from the database service. Was not able to get it to a clean conclusion though.

packages/core/src/database/index.ts
```diff
@@ -1167,17 +1168,20 @@ $$ LANGUAGE plpgsql
         ),
       );
     },
-    async kill() {
+    async unlock() {
```
0xOlias, line 1171: Nice, this makes way more sense.

0xOlias (approved):

PR #1217, opened 2024-11-03
feat: combine all intervals into one table

kyscott18:
This design uses a concept I've been calling "adjacent intervals". It moves the logic of deciding whether or not a fragment is valid for a certain interval to before the query occurs. 

For example, if a log filter has `topic2: "abc"`, then fragments with `topic2: "abc"` and `topic2: null` will be queried to see if they exist in the cache.

This makes it so filter specific information can be completely encoded into the `fragment_id` column of the `interval` table, and no other tables are needed.

## Other notes
- multirange columns for `interval` table
- `getIntervals` and `insertIntervals` were both made plural and use only one query
- `disableCache` avoids the interval table


packages/core/src/sync-realtime/filter.ts
```diff
@@ -12,7 +7,31 @@ import {
...
+  // match all
+  if (filterValue === null || filterValue === undefined) return true;
+
+  // missing value
+  if (eventValue === undefined) return false;
+
+  // array
+  if (
+    Array.isArray(filterValue) &&
+    filterValue.some((v) => v === toLowerCase(eventValue))
```
0xOlias, line 25: Is `filterValue` already expected to be lowercase here?

packages/core/src/sync-store/encoding.ts
```diff
@@ -304,86 +304,12 @@ type RpcRequestResultsTable = {
...
-  offset: number;
-};
-
-type BlockFilterIntervalsTable = {
+type IntervalTable = {
   id: Generated<number>;
-  blockFilterId: string;
-  startBlock: ColumnType<string, string | bigint, string | bigint>;
-  endBlock: ColumnType<string, string | bigint, string | bigint>;
+  fragment_id: string;
```
0xOlias, line 309: Love the move to snake case for internal SQL names.

packages/core/src/sync-store/encoding.ts
```diff
@@ -304,86 +304,12 @@ type RpcRequestResultsTable = {
...
+  chain_id: number;
+  start_block: ColumnType<string, string | bigint, string | bigint>;
+  end_block: ColumnType<string, string | bigint, string | bigint>;
```
0xOlias, line 312: What do you think about migrating to multirange data types in the same PR/database migration? Should simplify this table a bit (can use fragment_id as PK), but comes with some risk/uncertainty. 

https://www.postgresql.org/docs/current/functions-range.html

packages/core/src/sync-store/index.ts
```diff
@@ -171,291 +163,76 @@ export const createSyncStore = ({
...
-                    ]),
-                  );
-            }
-          }
-          break;
-
-        default:
-          never(filter);
-      }
+      const fragments = getFragmentIds(filter);
```
0xOlias, line 184: Epic diff lol

packages/core/src/sync-store/index.ts
```diff
@@ -171,291 +163,76 @@ export const createSyncStore = ({
...
             }
 
             if (
               mergedIntervalRows.length === common.options.syncStoreMaxIntervals
             ) {
-              // This occurs when there are too many non-mergeable ranges with the same logFilterId. Should be almost impossible.
+              // This occurs when there are too many non-mergeable ranges with the same fragment_id. Should be almost impossible.
               throw new NonRetryableError(
-                `'${table}Intervals' table for chain '${fragment.chainId}' has reached an unrecoverable level of fragmentation.`,
+                `'interval' table for chain '${filter.chainId}' has reached an unrecoverable level of fragmentation.`,
```
0xOlias, line 235: Multirange types should eliminate this codepath

packages/core/src/sync-store/index.ts
```diff
@@ -1273,212 +1048,19 @@ export const createSyncStore = ({
...
-          )
-          .deleteFrom("logFilterIntervals")
-          .where(
-            "logFilterId",
-            "in",
-            ksql`(SELECT "logFilterId" FROM ${ksql.table("deleteLogFilter")})`,
-          )
+          .deleteFrom("interval")
+          .where("chain_id", "=", chainId)
+          .where("start_block", ">=", fromBlock.toString())
```
0xOlias, line 1053: This query would be slightly more complex with multirange types

0xOlias (commented):
Excited for this

packages/core/src/sync/fragments.ts
```diff
@@ -166,93 +184,71 @@ function parseTopics(topics: (Hex | Hex[] | null)[] | undefined) {
...
+
+        const adjacent = [id];
+
+        if (_toAddress !== null) {
+          adjacent.push(`${chainId}_${null}_${_fromAddress}`);
+        }
+
+        if (_fromAddress !== null) {
+          adjacent.push(`${chainId}_${_toAddress}_${null}`);
+        }
```
khaidarkairbek, line 246: Is this a typo? Do we expect all the ids to be `${chainId}_${_fromAddress}_${_toAddress}` order for traces?

kyscott18: Might be misunderstanding your question. Are you wondering about the lack of factory contracts or the lack of fields like "error" or "function selector"?

kyscott18: Or the `null`

khaidarkairbek: no, in the first id you have fromAddress_toAddress order, in the second id you have fromAddress second , and in the third id you have toAddress first. Is there a reason behind switched order of addresses with respect to first one?

kyscott18: Oh that seems like a typo. Really nice catch - would have taken a while to debug this

packages/core/src/sync-realtime/filter.ts
kyscott18: yes

packages/core/src/sync-store/encoding.ts
kyscott18: I think we are fine to ship without it. Specifically because numeric(78, 0) is not a supported range type, we would have to define our own.

 https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-DEFINING

packages/core/src/sync-store/migrations.ts
```diff
@@ -832,6 +832,157 @@ const migrations: Record<string, Migration> = {
...
+        .expression(
+          db
+            .selectFrom("traceFilters as tf")
+            .innerJoin(
+              "traceFilterIntervals as tfi",
+              "tf.id",
+              "tfi.traceFilterId",
+            )
+            .select([
+              "tf.id as fragment_id",
```
0xOlias, line 912: While we're doing a migration, I think it would be worthwhile to document each of these fragment ID "schemas", and perhaps think ahead towards allowing factory configs to be included in more than just `log.address` and `trace.toAddress`, as specified [here](https://github.com/ponder-sh/ponder/issues/1198). Maybe they are already future-proof, not sure.

These sync store database migrations are "expensive" (currently a one-way door for a given database) so we should aim to minimize the # required.

packages/core/src/sync-store/index.ts
kyscott18: I ended up just changing the behavior of `disableCache` so that it doesn't ever read or write the `interval` table

packages/core/src/sync-store/index.ts
```diff
@@ -166,329 +161,75 @@ export const createSyncStore = ({
...
+        | undefined;
+
+      for (let i = 0; i < filters.length; i++) {
+        const filter = filters[i]!;
+        const fragments = getFragmentIds(filter);
+        for (const fragment of fragments) {
+          const _query = db
+            .selectFrom("interval")
+            .where("fragment_id", "in", fragment.adjacent)
+            .select(["blocks", sql<string>`${i}`.as("filter")]);
```
0xOlias, line 208: I think there's a small bug here where the adjacent range information is getting lost. Ideally we could union the multiranges for all the adjacent fragments that make up this fragment right here in this select statement. Maybe the `SUM` operator works?

https://www.postgresql.org/docs/current/functions-range.html#MULTIRANGE-OPERATORS-TABLE

packages/core/src/sync-store/index.ts
```diff
@@ -166,329 +161,75 @@ export const createSyncStore = ({
...
-            Number(startBlock),
-            Number(endBlock),
-          ]),
-        );
+      for (let i = 0; i < filters.length; i++) {
+        const filter = filters[i]!;
+        const intervals = rows
+          .filter((row) => row.filter === `${i}`)
+          .map((row) =>
+            intervalUnion(JSON.parse(`[${row.blocks.slice(1, -1)}]`)),
```
0xOlias, line 226: See above, but ideally this is not necessary.

packages/core/src/sync/fragments.test.ts
```diff
@@ -1,107 +1,53 @@
...
-      id: "115511_0xb_0xd_null_0xe_null_0",
-      address: "0xb",
-      topic0: "0xd",
-      topic1: null,
-      topic2: "0xe",
-      topic3: null,
-      includeTransactionReceipts: 0,
-    },
-  ]);
+  expect(logFilterFragments[0]!.id).toBe("log_115511_0xa_0xc_null_0xe_null_0");
```
0xOlias, line 43: Let's do one more review of these fragment ID encodings before we ship to main - want to make sure they are future-proofed for the more advanced filter types we've been chatting about.

Part of me thinks we should use a JSON design similar to the RPC request body encoding, and/or include a version key.

0xOlias (commented):
Looking great, one likely bug as we discussed

PR #1224, opened 2024-11-05
simplify dev server

kyscott18:
Provides a more long term solution to the problem fixed in https://github.com/ponder-sh/ponder/commit/160d1e2f9742304ed169acc5e3e6c23f35be094c

### DIFF
--- a/packages/core/src/build/service.ts
+++ b/packages/core/src/build/service.ts
@@ -68,11 +68,34 @@ export type ApiBuild = BaseBuild & {
   routes: PonderRoutes;
 };
 
-export type IndexingBuildResult =
+export type BuildResult =
+  | {
+      status: "success";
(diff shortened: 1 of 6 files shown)


packages/core/src/build/service.ts
```diff
@@ -68,11 +68,34 @@ export type ApiBuild = BaseBuild & {
...
+      indexingBuild: IndexingBuild;
+      apiBuild: ApiBuild;
+    }
+  | {
+      status: "success";
+      kind: "api";
+      indexingBuild?: never;
+      apiBuild: ApiBuild;
+    }
+  | { status: "error"; kind: "indexing" | "api"; error: Error };
```
0xOlias, line 92: Nice simplification, I like it.

PR #1227, opened 2024-11-05
feat: `context.client` supports all rpc methods

kyscott18:
Adds support for caching all available rpc methods. Several new methods are available off of `context.client`, including `simulateCall`, `getTransactionCount`, `getTransactionReceipt`, and `getBlockTransactionCount`.


cicadaweb3:
Would be awesome to have this!

packages/core/src/sync-store/migrations.ts
```diff
@@ -840,6 +840,70 @@ const migrations: Record<string, Migration> = {
         .execute();
     },
   },
+  "2024_11_04_0_request_cache": {
+    async up(db: Kysely<any>) {
+      await db.schema
+        .createTable("rpc_request_results")
+        .addColumn("request", "text", (col) => col.notNull())
+        .addColumn("blockNumber", "numeric(78, 0)")
+        .addColumn("chainId", "integer", (col) => col.notNull())
```
0xOlias, line 849: snake case

packages/core/src/utils/order.ts
```diff
@@ -0,0 +1,13 @@
+export const orderObject = <T extends { [key: string]: any }>(obj: T): T => {
+  const newObj = {} as T;
+  for (const key of Object.keys(obj).sort()) {
```
0xOlias, line 3: I think it would be more accurate to handle JSON arrays differently, because their order is already strict (two JSON arrays with a different order are expected to be different, unlike object keys).

packages/core/src/utils/order.ts
```diff
@@ -0,0 +1,15 @@
+export const orderObject = <T extends { [key: string]: any }>(obj: T): T => {
+  if (Array.isArray(obj)) return obj;
```
0xOlias, line 2: Think we need `.map` here

0xOlias (approved):

PR #1243, opened 2024-11-14
fix: `onchainSchema` creating undefined schema

kyscott18:
## Description
- Fixed a bug when using `onchainSchema()`

### DIFF
--- a/packages/core/src/database/index.ts
+++ b/packages/core/src/database/index.ts
@@ -473,25 +473,6 @@ export const createDatabase = (args: {
       });
     },
     async setup() {
-      await qb.internal.wrap({ method: "setup" }, async () => {
-        for (const statement of args.statements.schema.sql) {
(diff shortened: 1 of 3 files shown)


packages/core/src/database/index.ts
```diff
@@ -511,113 +492,154 @@ export const createDatabase = (args: {
...
+          // @ts-ignore
+          .select("table_name")
+          // @ts-ignore
+          .where("table_name", "=", "namespace_lock")
+          .executeTakeFirst()
+          .then((table) => table?.table_name === "namespace_lock");
 
-            await qb.internal
+        if (hasNamespaceLockTable) {
+          await qb.internal.wrap({ method: "migrates" }, async () => {
```
0xOlias, line 506: Maybe `method: "migrate"` ?

PR #1251, opened 2024-11-16
fix: OOM issues with memory optimized flush

kyscott18:
Use the "COPY" statement for a more performant database insertion. Memory usage is optimized for `flush` by allowing for garbage collection as soon as an entry is determined to be "released" (not saved in the cache). 

Some quick performance benchmarks showed the flush query is about 3.5x faster.

### Other notes
- Uses a more conservative value for `indexingCacheMaxBytes`


packages/core/src/indexing-store/historical.ts
```diff
@@ -753,133 +793,169 @@ export const createHistoricalIndexingStore = ({
...
 
-          const batchSize = Math.round(
-            common.options.databaseMaxQueryParameters /
-              Object.keys(getTableColumns(table)).length,
-          );
+        const promises: Promise<void>[] = [];
 
-          const insertValues = entries
+        for (const [table, tableCache] of cache) {
+          const removableInsertValues = Array.from(tableCache.values())
```
0xOlias, line 808: I'm curious if lines 808-851 could be done in a single pass over `tableCache.values()` (for loop + mutation vs. copies with map and filter).

Might not make a measurable difference though, not sure.

packages/core/src/indexing-store/historical.ts
```diff
@@ -753,133 +793,169 @@ export const createHistoricalIndexingStore = ({
...
+            nonRemovableInsertValues,
+          );
+          const updateValues = removableUpdateValues.concat(
+            nonRemovableUpdateValues,
+          );
+
+          const insertSize = insertValues.length;
+          const updateSize = updateValues.length;
+          const insertStream = getReadableStream(table, insertValues);
+          const updateStream = getReadableStream(table, updateValues);
```
0xOlias, line 856: Is there another opportunity to reduce peak mem usage here by handling the inserts first, then the updates? Again, just an idea - hard to tell what's big/small and what's GC'able here.

I guess the concurrency would get slightly worse.

packages/core/src/indexing-store/historical.ts
```diff
@@ -753,133 +793,169 @@ export const createHistoricalIndexingStore = ({
...
+              database.qb.user.wrap(
+                { method: `${tableNameCache.get(table)}.flush()` },
+                async () => {
+                  const client = await (
+                    database.driver as { internal: Pool }
+                  ).internal.connect();
+
+                  try {
+                    await client.query(`
+CREATE TEMP TABLE "${tableNameCache.get(table)}" AS
```
0xOlias, line 911: Is there any chance of a name collision with the temp table and an actual table?

0xOlias (commented):
Looking great

packages/core/src/indexing-store/historical.ts
```diff
@@ -753,133 +793,169 @@ export const createHistoricalIndexingStore = ({
...
+                    await client.query(`
+UPDATE "${getTableConfig(table).schema ?? "public"}"."${getTableName(table)}" as target
+SET ${set}
+FROM "${tableNameCache.get(table)}" as source
+WHERE ${primaryKeys.map(({ sql }) => `target."${sql}" = source."${sql}"`).join(" AND ")};
+`);
+
+                    await client.query(
+                      `DROP TABLE IF EXISTS "${tableNameCache.get(table)}"`,
+                    );
```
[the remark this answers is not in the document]
kyscott18: I thought so too, but when I didn't have this I saw a few instances of table name collisions on the second or third call to `flush()`

packages/core/src/indexing-store/historical.ts
kyscott18: No, the temp tables use a separate namespace that takes precedence over others

packages/core/src/indexing-store/historical.ts
kyscott18: It might, but the case we are really optimizing for is the first call to `flush()`, where all entries are `insert` and none are `find`. 

In this case all entries are serialized into csv, and only 35% of the original entries are garbage collectable. In subsequent call, a good portion of entries will be "find", which don't have to get serialized/copied.

0xOlias (approved):

PR #1259, opened 2024-11-19
GraphQL `totalCount`

0xOlias:
Adds `totalCount` field to GraphQL page type.

Also fixes two bugs that I noticed while implementing this:
- The one-to-many relation query path had a bug where the `orderBy` argument passed to the plural resolver attempted to use the columns of the parent table. Added a test for this, which was failing before the fix.
- The top-level singular resolver was not using dataloader. Now, certain query paths that do self-joins should do one fewer query.

### DIFF
--- a/packages/core/src/graphql/index.ts
+++ b/packages/core/src/graphql/index.ts
@@ -12,6 +12,7 @@ import {
   arrayContained,
   arrayContains,
   asc,
+  count,
   createTableRelationsHelpers,
   desc,
   eq,
@@ -36,7 +37,6 @@ import {
   PgSerial,
   isPgEnum,
 } from "drizzle-orm/pg-core";
-import type { RelationalQueryBuilder } from "drizzle-orm/pg-core/query-builders/query";
 import {
   GraphQLBoolean,
   GraphQLEnumType,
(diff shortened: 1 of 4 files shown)


packages/core/src/graphql/index.ts
```diff
@@ -564,22 +566,33 @@ async function executePluralQuery(
...
-  if (after !== null && before !== null) {
-    throw new Error("Cannot specify both before and after cursors.");
-  }
+  const totalCountPromise = includeTotalCount
+    ? drizzle
+        .select({ count: count() })
+        .from(rawTable)
+        .where(and(...whereConditions, ...extraConditions))
+        .then((rows) => rows[0]?.count ?? null)
+    : Promise.resolve(null);
```
kyscott18, line 584: Does this mean that queries that don't request total count will return data with `totalCount: null` ?

kyscott18 (approved):

packages/core/src/graphql/index.ts
0xOlias: Nope - GraphQL will ignore/strip any unselected properties on the object returned by the resolver.

PR #1261, opened 2024-11-20
Fix invalid schema examples in docs

gskril:


docs/pages/docs/migration-guide.mdx
gskril, line 1: also `t.evmBigint().notNull()` in this file doesn't seem to be valid — is that meant to be `t.bigint().notNull()` ?

kyscott18 (approved):
Thanks!

0xOlias (approved):

PR #1273, opened 2024-11-21
improve error logging for flush errors

0xOlias:


packages/core/src/common/logger.ts
```diff
@@ -116,6 +116,9 @@ const format = (log: Log) => {
       prettyLog.push(`${log.error.name}: ${log.error.message}`);
     }
 
+    if ("where" in log.error) {
+      prettyLog.push(`where: ${log.error.where as string}`);
+    }
```
0xOlias, line 121: Postgres queries often include a useful "where" property.

packages/core/src/indexing-store/realtime.ts
```diff
@@ -137,8 +137,7 @@ export const createRealtimeIndexingStore = ({
                         .onConflictDoNothing()
                         .returning();
                     } catch (e) {
-                      parseSqlError(e);
-                      throw e;
+                      throw parseSqlError(e);
```
0xOlias, line 140: Noticed that we were double-throwing here (parseSqlError was throwing internally). I think it makes more sense to return the error from the helper function and then throw it?

kyscott18 (approved):

PR #1274, opened 2024-11-22
fix: graphql snake case bugs

0xOlias:
See changeset. Note that there are still some cases of ambiguous parsing that we should consider validating against, e.g. if the column names themselves end with any of the where suffixes like `_not`, `_not_in`.


packages/core/src/graphql/index.ts
```diff
@@ -764,16 +769,32 @@ function buildWhereConditions(
       continue;
     }
 
-    const [fieldName, condition_] = whereKey.split(/_(.*)/s);
```
0xOlias, line 767: It's a miracle that this worked at all tbh.

kyscott18 (approved):

PR #1276, opened 2024-11-23
cosmetic improvements

kyscott18:
- `@ponder/core` -> `ponder`
- `ponder:schema`
- `ponder-env.d.ts`
- `@/generated` -> `ponder:registry`
- normalized `event`

### DIFF
--- a/benchmarks/apps/ponder-reth/ponder.config.ts
+++ b/benchmarks/apps/ponder-reth/ponder.config.ts
@@ -1,4 +1,4 @@
-import { createConfig } from "@ponder/core";
+import { createConfig } from "ponder";
 import { http } from "viem";
 
 import { RocketTokenRETHAbi } from "./abis/RocketTokenRETH";
--- a/examples/reference-erc20/ponder-env.d.ts
+++ b/examples/reference-erc20/ponder-env.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="ponder/virtual" />
 
-declare module "ponder:register" {
+declare module "ponder:internal" {
   const config: typeof import("./ponder.config.ts");
   const schema: typeof import("./ponder.schema.ts");
 }
--- a/examples/reference-erc20/ponder-env.d.ts
+++ b/examples/reference-erc20/ponder-env.d.ts
@@ -1,27 +1,15 @@
+/// <reference types="ponder/virtual" />
+
+declare module "ponda" {
+  const config: typeof import("./ponder.config.ts");
+  const schema: typeof import("./ponder.schema.ts");
+}
+
+declare module "ponder:schema" {
+  export * from "./ponder.schema.ts";
+}
+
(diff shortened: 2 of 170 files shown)


benchmarks/apps/ponder-reth/ponder.config.ts
```diff
@@ -1,4 +1,4 @@
-import { createConfig } from "@ponder/core";
+import { createConfig } from "ponder";
```
0xOlias, line 1: satisfied

examples/reference-erc20/ponder-env.d.ts
```diff
@@ -1,27 +1,15 @@
+/// <reference types="ponder/virtual" />
+
+declare module "ponder:register" {
```
0xOlias, line 3: What's the full list of names you considered here? Want to get this right, because it's going to show up in a LOT of code snippets for a long time :)

Edit: Oh, I figured this was the name that would replace "@/generated" but now I see that's different. Do you think we should rename "@/generated" in this release? I lean yes.

examples/with-client/ponder/src/api/index.ts
```diff
@@ -1,7 +1,7 @@
 import { type ApiContext, ponder } from "@/generated";
 import { trpcServer } from "@hono/trpc-server";
-import { eq } from "@ponder/core";
 import { initTRPC } from "@trpc/server";
+import { eq } from "ponder";
```
0xOlias, line 4: Is there any value to using different entrypoints here? 
```ts
import { eq } from "ponder/db";
// or 
import { eq } from "ponder/drizzle";
// or 
import { eq } from "ponder/sql";
```
Everything from top-level seems fine/simple, just curious if you gave it any thought.

packages/core/src/common/telemetry.ts
```diff
@@ -121,8 +121,7 @@ export function createTelemetry({
 
     // Attempt to find and read the users package.json file.
     const packageJson = getPackageJson(options.rootDir);
-    const ponderCoreVersion =
-      packageJson?.dependencies?.["@ponder/core"] ?? "unknown";
+    const ponderCoreVersion = packageJson?.dependencies?.ponder ?? "unknown";
```
0xOlias, line 124: `ponderVersion` ?

0xOlias (commented):
Awesome, just a few question on naming.

examples/reference-erc20/ponder-env.d.ts
kyscott18: Yes I think we should replace `@/generated`, but I haven't done it yet. 

This `ponder:register` is an entirely internal thing that is used to pass type information back into the `packages/core/src/types.d.ts`.

examples/with-client/ponder/src/api/index.ts
kyscott18: Let's take another look on monday

PR #1277, opened 2024-11-23
`factory()`

kyscott18:


examples/feature-factory/ponder.config.ts
```diff
@@ -20,21 +20,21 @@ export default createConfig({
     LlamaCore: {
       network: "sepolia",
       abi: LlamaCoreAbi,
-      factory: {
+      address: factory({
```
0xOlias, line 23: Looks so good!

packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -504,20 +505,13 @@ export async function buildConfigAndIndexingFunctions({
...
       const resolvedAddress = source?.address;
 
-      if (resolvedFactory !== undefined && resolvedAddress !== undefined) {
-        throw new Error(
-          `Validation failed: Contract '${contractMetadata.name}' cannot specify both 'factory' and 'address' options.`,
-        );
-      }
-
-      if (resolvedFactory) {
+      if (typeof resolvedAddress === "object") {
```
0xOlias, line 510: Are we validating the factory arguments at runtime anywhere?

packages/core/src/config/config.ts
```diff
@@ -200,54 +180,36 @@ type ContractsConfig<networks, contracts> = {} extends contracts
...
-            [name in allNetworkNames]?: Prettify<
-              GetAddress<unknown> & TransactionReceiptConfig & BlockConfig
-            >;
-          };
-    };
+> = {
+  /**
+   * Network that this account is deployed to. Must match a network name in `networks`.
+   * Any filter information overrides the values in the higher level "accounts" property.
+   * Factories cannot override an address and vice versa.
```
0xOlias, line 191: Is this comment now irrelevant?

0xOlias (approved):
Looks great! So much clearer this way. Couple nits.

packages/core/src/build/configAndIndexingFunctions.ts
kyscott18: https://github.com/ponder-sh/ponder/blob/kjs/factory/packages/core/src/build/factory.ts

packages/core/src/config/config.ts
kyscott18: nice catch

0xOlias (approved):
LGTM

PR #1278, opened 2024-11-24
add `include` property to filters

kyscott18:
Adds back support for including transaction receipts in indexing function events

### DIFF
--- a/packages/core/src/build/configAndIndexingFunctions.ts
+++ b/packages/core/src/build/configAndIndexingFunctions.ts
@@ -10,11 +10,17 @@ import {
   isRpcUrlPublic,
 } from "@/config/networks.js";
 import { buildAbiEvents, buildAbiFunctions, buildTopics } from "@/sync/abi.js";
-import type {
-  AccountSource,
-  BlockSource,
-  ContractSource,
-  Source,
+import {
+  type AccountSource,
+  type BlockSource,
+  type ContractSource,
+  type Source,
+  defaultBlockFilterInclude,
+  defaultLogFilterInclude,
+  defaultTraceFilterInclude,
+  defaultTransactionFilterInclude,
+  defaultTransactionReceiptInclude,
+  defaultTransferFilterInclude,
 } from "@/sync/source.js";
 import { chains } from "@/utils/chains.js";
 import { toLowerCase } from "@/utils/lowercase.js";
@@ -530,9 +536,13 @@ export async function buildConfigAndIndexingFunctions({
             topic1,
             topic2,
             topic3,
-            // includeTransactionReceipts: source.includeTransactionReceipts,
             fromBlock,
             toBlock,
+            include: defaultLogFilterInclude!.concat(
+              source.includeTransactionReceipts
+                ? defaultTransactionReceiptInclude
+                : [],
+            ),
           },
         } satisfies ContractSource;
 
@@ -549,10 +559,13 @@ export async function buildConfigAndIndexingFunctions({
                 callType: "CALL",
                 functionSelector: registeredFunctionSelectors,
                 includeReverted: false,
-                // includeTransactionReceipts:
-                //   rawContract.includeTransactionReceipts,
                 fromBlock,
                 toBlock,
+                include: defaultTraceFilterInclude!.concat(
+                  source.includeTransactionReceipts
+                    ? defaultTransactionReceiptInclude
+                    : [],
+                ),
               },
             } satisfies ContractSource,
           ];
@@ -593,9 +606,13 @@ export async function buildConfigAndIndexingFunctions({
           topic1,
           topic2,
           topic3,
-          // includeTransactionReceipts: rawContract.includeTransactionReceipts,
           fromBlock,
           toBlock,
+          include: defaultLogFilterInclude!.concat(
+            source.includeTransactionReceipts
+              ? defaultTransactionReceiptInclude
+              : [],
+          ),
         },
       } satisfies ContractSource;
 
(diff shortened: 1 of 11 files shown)


packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -530,9 +536,13 @@ export async function buildConfigAndIndexingFunctions({
             topic1,
             topic2,
             topic3,
-            // includeTransactionReceipts: source.includeTransactionReceipts,
             fromBlock,
             toBlock,
+            include: defaultLogFilterInclude!.concat(
```
0xOlias, line 541: Is this non-null assertion a typo?

packages/core/src/sync-historical/index.ts
```diff
@@ -535,13 +536,32 @@ export const createHistoricalSync = async (
       }),
     ).then((traces) => traces.flat());
 
-    if (traces.length > 0) {
-      await args.syncStore.insertTraces({
-        traces,
+    if (isKilled) return;
+
+    await args.syncStore.insertTraces({
```
0xOlias, line 541: Notice that you removed the `if (traces.length > 0)` check, seems fine but figured I'd flag

packages/core/src/sync-historical/index.ts
```diff
@@ -535,13 +536,32 @@ export const createHistoricalSync = async (
...
+      traces,
+      chainId: args.network.chainId,
+    });
+
+    if (isKilled) return;
+
+    if (shouldGetTransactionReceipt(filter)) {
+      const transactionHashes = new Set(
```
0xOlias, line 549: Totally speculative, but I wonder if it makes sense to initialize this before we call `insertTraces` so that the relatively large `traces` object can be GC'd sooner.

packages/core/src/sync-historical/index.ts
```diff
@@ -535,13 +536,32 @@ export const createHistoricalSync = async (
...
+        traces.map(({ transaction }) => transaction.hash),
+      );
+
+      const transactionReceipts = await Promise.all(
+        Array.from(transactionHashes).map((hash) =>
+          _eth_getTransactionReceipt(args.requestQueue, { hash }),
+        ),
+      );
```
0xOlias, line 557: Definitely a follow-up concern, but I'm guessing we could de-duplicate these requests across filters. And, if we add support for `eth_getBlockReceipts` that would also be suitable for reuse (same way we do it for transaction inputs).

packages/core/src/sync-realtime/index.ts
```diff
@@ -674,10 +675,13 @@ export const createRealtimeSync = (
...
       for (const filter of logFilters) {
         if (isLogFilterMatched({ filter, block, log })) {
-          // TODO: includeTransactionReceipt
-          // return early if includeTransactionReceipt, otherwise continue checking filters
           requiredTransactions.add(log.transactionHash);
           isMatched = true;
+          if (shouldGetTransactionReceipt(filter)) {
+            requiredTransactionReceipts.add(log.transactionHash);
+            // skip to next log
+            break;
```
0xOlias, line 683: If I'm reading correctly, this `break` skips checking this log against all subsequent filters. I guess this is fine because we've already definitely added this `log.transactionHash` to `requiredTransactions` and `requiredTransactionReceipts`, and that's the only side-effect we care about in this loop?

Control flow is a bit difficult to follow, but looks correct to me.

packages/core/src/sync/events.ts
```diff
@@ -308,7 +316,9 @@ export const buildEvents = ({
                   trace: convertTrace(trace),
                   block: convertBlock(block),
                   transaction: convertTransaction(transaction),
-                  transactionReceipt: undefined,
+                  transactionReceipt: shouldGetTransactionReceipt(filter)
```
0xOlias, line 319: IIRC we talked about adding the receipt properties to the `transaction` object - what's your latest thinking?

packages/core/src/sync/source.ts
```diff
@@ -171,3 +210,123 @@ export const getChildAddress = ({
...
+    "transactionReceipt.logs",
+    "transactionReceipt.logsBloom",
+    "transactionReceipt.status",
+    "transactionReceipt.to",
+    "transactionReceipt.transactionHash",
+    "transactionReceipt.transactionIndex",
+    "transactionReceipt.type",
+  ];
+
+const defaultTraceInclude: `trace.${keyof UserTrace}`[] = [
```
0xOlias, line 269: Is this missing `defaultBlockFilterInclude` and `defaultTransactionInclude`?

packages/core/src/sync/fragments.ts
```diff
@@ -204,76 +126,149 @@ export const getTransactionFilterFragmentIds = ({
...
 
   for (const fragmentFromAddress of fromAddressFragmentIds) {
     for (const fragmentToAddress of toAddressFragmentIds) {
-      const id =
-        `transfer_${chainId}_${fragmentFromAddress.id}_${fragmentToAddress.id}` as const;
+      for (const fragmentFunctionSelector of Array.isArray(functionSelector)
+        ? functionSelector
+        : [functionSelector]) {
+        const id =
+          `trace_${chainId}_${fragmentFromAddress.id}_${fragmentToAddress.id}_${callType ?? null}_${fragmentFunctionSelector ?? null}_${includeTransactionReceipts ? 1 : 0}` as const;
```
0xOlias, line 150: Could you explain why `callType` is included in the fragment key and when it might come into play?

0xOlias (commented):
Looking great, few questions

packages/core/src/sync-historical/index.ts
kyscott18: Yes, it's checked in `syncStore.insertTraces()`

packages/core/src/sync/fragments.ts
kyscott18: Gonna remove

packages/core/src/sync/events.ts
kyscott18: Gonna do it in a separate pr

packages/core/src/sync-realtime/index.ts
kyscott18: Yes, @khaidarkairbek can you take a look as well?

khaidarkairbek: Yes, there is no need to iterate further if we have already included the log and added the tx receipt. These are the only side effects so far.

PR #1280, opened 2024-11-24
fix: value returned by realtime store

### LINKED ISSUE #1279: ORM differential behavior between `historical` and `live` indexing
Context: Seeing 0.8 around the corner, decided to upgrade to 0.7 and found a bug that prevents me to upgrade without going into lengthy workaround due to the ORM.

Description: Using ORM functions such as `insert` and `update` have a different return type between `historical` (returns object) and `live` (returns array) indexing.

Reproduction:
- GitHub: https://github.com/robiiinos/ponder-repro
> Clone the project, install dep's (`yarn`) and run the project (`yarn dev`)

Using latest version 0.7.9. Indexing should only take a few minutes as I changed the startBlock to be very close to actual time

The return type is typed correctly, it expects an object, and during `historical`, it returns an object as we can see in the below 2 images

Screenshot 2024-11-24 at 9 57 37 PM

Screenshot 2024-11-24 at 10 01 58 PM

and as soon as the indexer turns into `live` indexing, it starts returning an array of length 1 with the correct object inside
Expected behavior should be to return an object, as during the `historical` indexing

Screenshot 2024-11-24 at 10 02 17 PM

kyscott18:
Closes #1279

### DIFF
--- a/packages/core/src/indexing-store/realtime.ts
+++ b/packages/core/src/indexing-store/realtime.ts
@@ -23,6 +23,7 @@ import {
 import { type PgTable, getTableConfig } from "drizzle-orm/pg-core";
 import { drizzle } from "drizzle-orm/pg-proxy";
 import { createQueue } from "../../../common/src/queue.js";
+import { normalizeColumn } from "./historical.js";
 import { type IndexingStore, parseSqlError } from "./index.js";
 
 /** Throw an error if `table` is not an `onchainTable`. */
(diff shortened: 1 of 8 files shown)


packages/core/src/indexing-store/realtime.ts
```diff
@@ -163,7 +164,10 @@ export const createRealtimeIndexingStore = ({
                               .map(({ js }) => table[js]),
                             set: valuesU,
                           })
-                          .returning();
+                          .returning()
+                          .then((res) =>
+                            Array.isArray(values) ? res : res[0],
```
0xOlias, line 169: So, `res` can either be a single row or an array (matching `values`)?

Seems like we don't have type safety here, so would be good to make sure each of these branches works as expected. Do you think it's worthwhile to update the tests to also validate return objects?

0xOlias (commented):
Makes sense!

0xOlias (approved):

PR #1293, opened 2024-11-27
flush cache during dev

kyscott18:


packages/core/src/bin/utils/run.ts
```diff
@@ -175,7 +175,11 @@ export async function run({
       // Persist the indexing store to the db if it is too full. The `finalized`
       // checkpoint is used as a mutex. Any rows in the reorg table that may
       // have been written because of raw sql access are deleted.
-      if (historicalIndexingStore.isCacheFull() && events.length > 0) {
+      if (
+        (historicalIndexingStore.isCacheFull() ||
+          common.options.command === "dev") &&
```
kyscott18, line 180: @typedarray Should we using a different algorithm than this?

typedarray: Approx how often will this run?

kyscott18: every 10k events

typedarray (approved):
LGTM

PR #1295, opened 2024-11-29
feat(core.indexing-store): introduce mode param to indexing functions

### LINKED ISSUE #1294: feat: pass the indexing store mode into the indexing function
This issue will be solved once the indexing function have access to the indexing store mode, which is either, `historical`, or `realtime`.

tk-o:
Resolves #1294 

This PR allows the indexing function to read the mode of the indexing store.

It's a useful information when sending out notifications to other services (i.e. only sending them out during the realtime indexing phase).

### DIFF
--- a/packages/core/src/indexing-store/historical.test.ts
+++ b/packages/core/src/indexing-store/historical.test.ts
@@ -36,6 +36,8 @@ test("find", async (context) => {
     initialCheckpoint: encodeCheckpoint(zeroCheckpoint),
   });
 
+  expect(indexingStore.mode).toBe("historical");
+
   // empty
 
   let result = await indexingStore.find(schema.account, {
--- a/packages/core/src/indexing-store/historical.ts
+++ b/packages/core/src/indexing-store/historical.ts
@@ -1001,6 +1001,9 @@ export const createHistoricalIndexingStore = ({
     isCacheFull() {
       return cacheBytes > maxBytes;
     },
+    get mode() {
+      return "historical" as const;
+    },
   } satisfies IndexingStore<"historical">;
 
   // @ts-ignore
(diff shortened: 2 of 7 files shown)


typedarray:
Hi, thanks for opening. Unfortunately, Ponder does not support running side effects (e.g. notifications) within indexing functions yet. As [this comment](https://github.com/ponder-sh/ponder/issues/810#issuecomment-2450234890) describes, doing side effects _properly_ is quite difficult.

I fear that if we add & document a "mode" flag, it would an implicit endorsement of side effects. More folks will start using them, and inevitably run into issues because the complexities described above won't be properly handled. 

The support burden resulting from these issues will drain maintainer resources that we'd rather spend elsewhere (like to design & build a proper effects feature!).

I fully expect that effects will be a banger feature when we release it. I just want to make sure we do it with intention. If you respond to the issue thread above describing your notifications use case, that would be a huge help - the faster we can gather real requirements, the sooner we'll be able to ship it.

Closing for now - let's move any further discussion to that issue.

tk-o:
> I fear that if we add & document a "mode" flag, it would an implicit endorsement of side effects. More folks will start using them, and inevitably run into issues because the complexities described above won't be properly handled.

I will have to find a workaround for my use case. Probably will create a small module that will keep track of [the last finalized block](https://docs.tatum.io/docs/evm-block-finality-and-confidence) and will trigger a fire-and-forget notification for other services in my web stack to react to.

tk-o:
Can we please have a go for this one? It would help me greatly testing my side effects processor implementation that I've discussed in:
- #810 

Also, it's an opt-in, so not everyone has to use it. But it's very important for testing out a possible solution to handling effects.

Please reconsider merging this one in, @typedarray.

Issue #1291, opened 2024-11-27
feat: refactor `rpc` module


kyscott18:
## Description
Refactor the `rpc` module to support advanced features such as websockets, introspection, supported methods, load balancing, and rate limiting. The rpc module should still support the eip1193 `request` api, but potentially with support for extra options as a second parameter.

Current module for reference:
https://github.com/ponder-sh/ponder/blob/afba4c361ec70b3ab5de60e6b82c8112331cc185/packages/core/src/utils/requestQueue.ts

## Other notes
The config should be updated as well

### Related 
#1290

Issue #1290, opened 2024-11-27
feat: Use websockets for `sync-realtime`


kyscott18:
## Description
Using websockets to replace the current polling algorithm would be more performant and cheaper.

https://github.com/ponder-sh/ponder/blob/afba4c361ec70b3ab5de60e6b82c8112331cc185/packages/core/src/sync-realtime/index.ts#L810-L855

Currently `enqueue` is used with `setInterval`, but should be swapped out with a websocket request, similar to https://viem.sh/docs/actions/public/watchBlocks

## Other notes
This will require some updates to the rpc infrastructure to separate out websockets and https, and also be able to introspect as to whether a websocket transport is available

PR #1296, opened 2024-11-30
rpc refactoring

khaidarkairbek:
Relates to #1291 and #1290


packages/core/src/sync-realtime/index.ts
```diff
@@ -807,11 +816,63 @@ export const createRealtimeSync = (
...
         try {
-          const block = await _eth_getBlockByNumber(args.requestQueue, {
-            blockTag: "latest",
-          });
+          const block =
+            blockHash === undefined
+              ? await _eth_getBlockByNumber(args.requestQueue, {
+                  blockTag: "latest",
+                })
+              : await _eth_getBlockByHash(args.requestQueue, {
```
kyscott18, line 873: how come we still need to fetch the block? Does `newHeads` not return the full block object?

packages/core/src/sync-realtime/index.ts
```diff
@@ -807,11 +816,63 @@ export const createRealtimeSync = (
...
+
+            // After a certain number of attempts, emit a fatal error.
+            if (++consecutiveErrors === ERROR_TIMEOUT.length) {
+              args.common.logger.error({
+                service: "realtime",
+                msg: `Fatal warning: Failed to subscribe to the latest '${args.network.name}' block after ${ERROR_TIMEOUT.length} attempts.`,
+                error,
+              });
+
+              args.onFatalError(error);
```
kyscott18, line 860: Is there any way we could reuse the same retry logic we have in `enqueue`?

packages/core/src/utils/requestQueue.ts
```diff
@@ -33,6 +36,24 @@ export type RequestQueue = Omit<
...
+
+export type SubscribeParameters = Parameters<
+  ResolvedWebSocketTransport["value"]["subscribe"]
+>[0] & {
+  method: "eth_subscribe";
+};
+
+export type SubscribeReturnType = Awaited<
+  ReturnType<ResolvedWebSocketTransport["value"]["subscribe"]>
+>;
```
kyscott18, line 50: Are these types from viem?

packages/core/src/utils/requestQueue.ts
```diff
@@ -104,10 +125,63 @@ export const createRequestQueue = ({
...
 
+  const subscribe = async (request: Omit<SubscribeParameters, "method">) => {
+    for (let i = 0; i <= RETRY_COUNT; i++) {
+      try {
+        const stopClock = startClock();
+
+        const wsTransport = resolveWebsocketTransport(network.transport);
+
+        if (wsTransport === undefined) {
+          throw new Error(
```
kyscott18, line 136: I think this should be `NonRetryableError`

packages/core/src/utils/requestQueue.ts
```diff
@@ -173,3 +257,28 @@ function shouldRetry(error: Error) {
...
+
+    return wsTransport;
+  }
+
+  if (transport.config.type === "webSocket") {
+    return transport as ResolvedWebSocketTransport;
+  }
+
+  return undefined;
+}
```
kyscott18, line 284: What do you think about getting away from using the `Transport` type in this module?

In the future this would let us implement features that aren't available in the transports like request cancellation. I also feel confident that `ponder.config.ts` should use rpc urls, not transports.

packages/core/src/sync-realtime/index.ts
khaidarkairbek: `newHeads` returns the block header, which does not include transactions. So, we still need to fetch via _eth_getBlockByHash to access complete block data.

packages/core/src/utils/requestQueue.ts
khaidarkairbek: agree, it should always be defined by the way we make use of _eth_subscribe in sync-realtime.

packages/core/src/utils/requestQueue.ts
khaidarkairbek: Yes, `ResolvedWebSocketTransport` is based on the `WebSocketTransport` type defined in viem. 
```ts
type ResolvedWebSocketTransport = Omit<
  ReturnType<WebSocketTransport>,
  "value"
> & {
  value: NonNullable<ReturnType<WebSocketTransport>["value"]>;
};
```

packages/core/src/utils/requestQueue.ts
khaidarkairbek: I think this makes sense, and we can make this change once we get to config and switch to rpcUrls instead of transports.

packages/core/src/rpc/index.ts
```diff
@@ -135,6 +205,11 @@ export const createRequestQueue = ({
     ) => {
       const stopClockLag = startClock();
 
+      return requestQueue.add({ request: params, stopClockLag });
+    },
+    subscribe: (params: SubscribeParameters) => {
+      const stopClockLag = startClock();
+
       return requestQueue.add({ request: params, stopClockLag });
```
kyscott18, line 213: Thinking about `subscribe()` not adding to the queue. I think it's safe to assume that we aren't going to get a 429 with the requests sent to it, because there should just be one per chain.

packages/core/src/rpc/index.ts
```diff
@@ -116,15 +176,25 @@ export const createRequestQueue = ({
...
         { method: task.request.method, network: network.name },
         task.stopClockLag(),
       );
 
-      return await fetchRequest(task.request);
+      if (task.request.method === "eth_subscribe") {
+        return await withRetry({
+          fn: subscribe,
+          request: task.request,
+        });
```
kyscott18, line 191: What are the difference between the errors throw in `subscribe()` vs the errors passed to `onError()` in the subscribe params

packages/core/src/sync-realtime/index.ts
```diff
@@ -833,28 +910,16 @@ export const createRealtimeSync = (
...
+          onError(_error as Error);
         }
       };
 
-      interval = setInterval(enqueue, args.network.pollingInterval);
+      if (isPolling) {
+        interval = setInterval(enqueue, args.network.pollingInterval);
+      } else {
+        // Check on the connection every second
+        interval = setInterval(watchWebsocket, 1000);
```
kyscott18, line 921: Do we need to do this? Looking at the viem implementation of `watchBlocks` seems like they are handling it differently

https://github.com/wevm/viem/blob/main/src/actions/public/watchBlocks.ts

packages/core/src/rpc/index.ts
khaidarkairbek: I was confused a bit too, as far as I know, the onError handles the errors thrown during the active websocket connection, while the errors thrown in subscribe happen when establishing this connection. Feel free to correct me here.

kyscott18: I'm not sure either but that makes sense

packages/core/src/sync-realtime/index.ts
khaidarkairbek: My idea behind this implementation was to have persistency of websocket connection, so if error happens during active connection, we drop it and reconnect again. From the watchblocks implementation, they seem to leave this flexibility via onError callback.

packages/core/src/rpc/index.ts
khaidarkairbek: Yeah that makes sense.

PR #1297, opened 2024-11-30
remove transactionReceipt.logs

kyscott18:


typedarray (approved):
LGTM, let's make sure we include this in the migration guide

PR #1328, opened 2024-12-11
Improve type checking for resolvedAddress in configAndIndexingFunctions

3commascapital:
- Add check to ensure resolvedAddress is not an array
- Remove unnecessary type assertion when spreading resolvedAddress


kyscott18 (approved):
Nice fix!

typedarray (approved):

PR #1335, opened 2024-12-13
fix: reorgs with multiple chains causing missing events

kyscott18:
### Description
Fixes an issues when handling reorgs with a multi chain configuration. This bug occurs when a reorg on one chain caused an event on a different chain to be reverted and never re-run.

### Example
1. event occurs on chain A, get handled with an indexing function
2. reorg detected on chain B
3. event on chain A gets reverted, because it is chronologically ahead of the reorg
4. chain B gets reprocessed, but not chain A

### DIFF
--- a/packages/core/src/sync/index.ts
+++ b/packages/core/src/sync/index.ts
@@ -226,14 +226,15 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
       syncProgress: SyncProgress;
       historicalSync: HistoricalSync;
       realtimeSync: RealtimeSync;
-      unfinalizedEventData: BlockWithEventData[];
+      unfinalizedEventData: (BlockWithEventData & {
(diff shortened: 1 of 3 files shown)


typedarray (approved):
LGTM

packages/core/src/sync/index.ts
```diff
@@ -723,57 +711,67 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
             service: "sync",
             msg: `Finalized block for '${network.name}' has surpassed overall indexing checkpoint`,
           });
+          return;
```
kyscott18, line 714: @typedarray Let me know what you think about this. Because we coupled events with raw block data, we must not flush either until it is known that the events have been processed. Another solution is to decouple these concepts, but I think the implementation of this will be harder to reason about and less maintainable.

PR #1342, opened 2024-12-16
feat: replacing getTransactionReceipt with getBlockReceipts

### LINKED ISSUE #1197: feature: Using eth_getBlockReceipts
Currently in sync-historical, to check for reverted transactions and to include tx receipts, we fetch receipts via eth_getTransactionReceipt method on ethereum API. However, when the number of transactions after filtering is large, would it be more efficient to use eth_getBlockReceipts to lower the number of requests made to the RPC? 

```ts
// Request transactionReceipts to check for reverted transactions.
    const transactionReceipts = await Promise.all(
      Array.from(transactionHashes).map((hash) =>
        _eth_getTransactionReceipt(args.requestQueue, {
          hash,
        }),
      ),
    );
```

khaidarkairbek:
Closes #1197 . Includes the new helper function (`_eth_getBlockReceipts`) and changes to sync-realtime / sync-histroical. This change should lower the number of requests to RPC and CU costs.

### DIFF
--- a/packages/core/src/sync-historical/index.ts
+++ b/packages/core/src/sync-historical/index.ts
@@ -18,7 +18,7 @@ import {
   shouldGetTransactionReceipt,
 } from "@/sync/source.js";
 import type { Source, TransactionFilter } from "@/sync/source.js";
-import type { SyncBlock, SyncLog, SyncTrace } from "@/types/sync.js";
+import type { SyncBlock, SyncLog, SyncTrace, SyncTransactionReceipt } from "@/types/sync.js";
 import {
   type Interval,
   getChunks,
@@ -32,7 +32,6 @@ import {
   _eth_getBlockByNumber,
   _eth_getBlockReceipts,
   _eth_getLogs,
-  _eth_getTransactionReceipt,
 } from "@/utils/rpc.js";
 import { getLogsRetryHelper } from "@ponder/utils";
 import {
@@ -83,6 +82,12 @@ export const createHistoricalSync = async (
    * Note: All entries are deleted at the end of each call to `sync()`.
    */
   const transactionsCache = new Set<Hash>();
+  /**
+   * Block transaction receipts that have already been fetched. 
+   * Note: All entries are deleted at the end of each call to `sync()`.
+   */
+  const blockReceiptsCache = new Map<Hash, Promise<SyncTransactionReceipt[]>>(); 
+
   /**
    * Data about the range passed to "eth_getLogs" for all log
    *  filters and log factories.
@@ -289,6 +294,18 @@ export const createHistoricalSync = async (
     }
   };
 
+  const syncBlockReceipts = async (block: Hash) => {
+    if (blockReceiptsCache.has(block)) {
+      return await blockReceiptsCache.get(block)!; 
+    } else {
+      const blockReceipts = _eth_getBlockReceipts(args.requestQueue, {
+        blockHash: block, 
+      }); 
+      blockReceiptsCache.set(block, blockReceipts); 
+      return await blockReceipts; 
+    }
+  };
+
   /** Extract and insert the log-based addresses that match `filter` + `interval`. */
   const syncLogFactory = async (filter: LogFactory, interval: Interval) => {
     const logs = await syncLogsDynamic({
@@ -388,12 +405,12 @@ export const createHistoricalSync = async (
     if (isKilled) return;
 
     if (shouldGetTransactionReceipt(filter)) {
-      const requiredBlockReceipts = new Set(blocks.map((b) => b.number));
+      const requiredBlockReceipts = new Set(blocks.map((b) => b.hash));
       // Request transactionReceipts to check for reverted transactions.
       const blockReceipts = await Promise.all(
-        [...requiredBlockReceipts].map((blockNumber) =>
-          _eth_getBlockReceipts(args.requestQueue, { blockNumber }).then(
-            (receipts) => ({ blockNumber, receipts }),
+        Array.from(requiredBlockReceipts).map((blockHash) =>
+          syncBlockReceipts(blockHash).then(
+            (receipts) => ({ blockHash, receipts }),
           ),
         ),
       );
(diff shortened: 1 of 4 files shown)


packages/core/src/sync-realtime/index.ts
```diff
@@ -779,12 +780,23 @@ export const createRealtimeSync = (
...
     ////////
 
-    const transactionReceipts = await Promise.all(
-      block.transactions
-        .filter(({ hash }) => requiredTransactionReceipts.has(hash))
-        .map(({ hash }) =>
-          _eth_getTransactionReceipt(args.requestQueue, { hash }),
-        ),
+    const blockReceipts = await _eth_getBlockReceipts(args.requestQueue, {
+      blockNumber: block.number,
```
kyscott18, line 784: Should use `blockHash` whenever possible

packages/core/src/sync-realtime/index.ts
```diff
@@ -779,12 +780,23 @@ export const createRealtimeSync = (
...
+    const blockReceiptsTransactionHashes = new Set(
+      blockReceipts.map((r) => r.transactionHash),
+    );
+    for (const hash of Array.from(requiredTransactions)) {
+      if (blockReceiptsTransactionHashes.has(hash) === false) {
+        throw new Error(
+          `Detected inconsistent RPC responses. Transaction receipt with transactionHash ${hash} is missing in \`blockReceipts\`.`,
+        );
+      }
+    }
```
kyscott18, line 797: Nice check

packages/core/src/utils/rpc.ts
```diff
@@ -146,6 +146,24 @@ export const _eth_getTransactionReceipt = (
       return receipt as SyncTransactionReceipt;
     });
 
+/**
+ * Helper function for "eth_getBlockReceipts" request.
+ */
+export const _eth_getBlockReceipts = (
```
kyscott18, line 152: Can `_eth_getTransactionReceipts` be removed?

packages/core/src/sync-historical/index.ts
```diff
@@ -387,14 +388,38 @@ export const createHistoricalSync = async (
     if (isKilled) return;
 
     if (shouldGetTransactionReceipt(filter)) {
-      const transactionReceipts = await Promise.all(
-        Array.from(transactionHashes).map((hash) =>
-          _eth_getTransactionReceipt(args.requestQueue, { hash }),
+      const requiredBlockReceipts = new Set(blocks.map((b) => b.number));
+      // Request transactionReceipts to check for reverted transactions.
+      const blockReceipts = await Promise.all(
+        [...requiredBlockReceipts].map((blockNumber) =>
```
kyscott18, line 394: `Array.from(requiredBlockReciepts)`

packages/core/src/sync-historical/index.ts
```diff
@@ -465,11 +492,30 @@ export const createHistoricalSync = async (
...
-    const transactionReceipts = await Promise.all(
-      Array.from(transactionHashes).map((hash) =>
-        _eth_getTransactionReceipt(args.requestQueue, { hash }),
-      ),
-    );
+    const blockReceipts = (await Promise.all(
+      [...requiredBlockReceipts].map((blockNumber) => _eth_getBlockReceipts(args.requestQueue, {
+        blockNumber,
+      }).then(receipts => ({ blockNumber, receipts })))
+    ));
```
kyscott18, line 499: Curious if you think block receipts would benefit from using the same "cache" concept that is used in this file for block and traces?

kevoconnell:
This would be a great change - am getting rekd by getTransactionReceipt ATM

packages/core/src/sync-realtime/index.ts
khaidarkairbek: What do you think of the checks in sync-historical?

packages/core/src/utils/rpc.ts
khaidarkairbek: Some test files are using `_eth_getTransacationReceipt`, but I will remove this method before merging

packages/core/src/sync-historical/index.ts
khaidarkairbek: I think it would definitely benefit. Included in the commit below.

packages/core/src/utils/rpc.ts
```diff
@@ -146,6 +146,20 @@ export const _eth_getTransactionReceipt = (
...
+  requestQueue: RequestQueue,
+  { blockHash }: { blockHash: Hash },
+): Promise<SyncTransactionReceipt[]> =>
+  requestQueue
+    .request({
+      method: "eth_getBlockReceipts",
```
typedarray, line 158: Just did some research on chain support.

- `eth_getBlockReceipts` is non-standard
- Erigon added it in April 2021 ([pr](https://github.com/erigontech/erigon/pull/1735)) and it started becoming popular
- Geth only added support in September 2023 ([release](https://github.com/ethereum/go-ethereum/releases/tag/v1.13.0))
- It's still somewhat [controversial](https://github.com/base-org/node/issues/125#issuecomment-1758344540)

TL;DR I strongly suspect we will encounter chains that don't support it, and we'll need to include a fallback that uses `eth_getTransactionReceipt`. Also may make sense to use that method anyway if the # of required transactions in the block is less than some threshold (2?).

kyscott18: They are currently priced the same on alchemy

https://docs.alchemy.com/reference/compute-unit-costs#standard-evm-json-rpc-methods

kyscott18:
If eth_getBlockReceipts proves to be successful, I think it would be reasonable to greatly simplify `syncRealtime.fetchBlockEventData()`

kyscott18 (approved):

packages/core/src/sync-historical/index.ts
```diff
@@ -288,6 +311,70 @@ export const createHistoricalSync = async (
...
+        ),
+      );
+
+      return transactionReceipts;
+    }
+
+    let blockReceipts: SyncTransactionReceipt[];
+    try {
+      blockReceipts = await syncBlockReceipts(block);
+    } catch {
```
typedarray, line 331: I think it makes sense to log this error at `debug` or maybe `warn`

typedarray (approved):
Looks good - last note is that it would be nice to see a manual test verifying that this works, eg run one of the example apps indexing accounts and show that the `/metrics` response includes eth_getBlockReceipts requests.

packages/core/src/sync-historical/index.ts
```diff
@@ -328,7 +328,16 @@ export const createHistoricalSync = async (
...
-    } catch {
+    } catch (_error) {
+      const error = _error as Error;
+      args.common.logger.warn({
+        service: "sync",
+        msg: `Caught eth_getBlockReceipts error on '${
+          args.network.name
+        }', switching to eth_getTransactionReceipt method.`,
+        error,
+      });
```
kyscott18, line 339: great log

PR #1351, opened 2024-12-16
api functions refactor

kyscott18:
## Description
- `import { db } from "ponder:api"`
- remove `ponder.get`
- removed `schema.graphql` from codegen
- requires `src/api/index.ts`
- requires default export from `src/api/index.ts`
- `getNextAvailablePort()`


docs/pages/docs/migration-guide.mdx
```diff
@@ -7,6 +7,83 @@ import { FileTree, Steps, Tabs, Callout } from "nextra/components";
 
 # Migration guide
 
+## 0.9.0-next.0
+
+### API functions
+
```
typedarray, line 13: High-level overview would be helpful:
```
This release makes Ponder's API layer less opinionated. Now, you just need to default export a Hono `App` object from the `src/api/index.ts` file, and Ponder will serve it.
```

docs/pages/docs/migration-guide.mdx
```diff
@@ -7,6 +7,83 @@ import { FileTree, Steps, Tabs, Callout } from "nextra/components";
...
+});
+
+export default app;
+```
+
+</div>
+
+#### `import { db } from "ponder:api"`
+
+The `db` object is now injected into the API function context. `c.db` is removed.
```
typedarray, line 61: The `c.db` object was removed from the Hono context. Now, use the `"ponder:api"` virtual module to access a readonly Drizzle database object.

docs/pages/docs/query/api-functions.mdx
```diff
@@ -31,11 +31,15 @@ API functions are available starting from version `0.5.0`. Read the [migration g
...
+import { Hono } from "hono";
 
-ponder.get("/hello", (c) => {
+const app = new Hono();
+
+app.get("/hello", (c) => {
   return c.text("Hello, world!");
 });
+
+export default app;
```
typedarray, line 42: So much better!

docs/pages/docs/migration-guide.mdx
```diff
@@ -7,6 +7,83 @@ import { FileTree, Steps, Tabs, Callout } from "nextra/components";
...
+import { graphql } from "ponder";
+
+const app = new Hono();
+
+app.use("/graphql", graphql({ db, schema }));
+
+export default app;
+```
+
+#### Removed `ponder.get()`
```
typedarray, line 31: Removed `ponder.get()`, `post()` and `use()`

Now, simply use Hono's built-in routing system.

docs/pages/docs/query/graphql.mdx
```diff
@@ -12,14 +12,42 @@ import { Callout } from "nextra/components";
...
+```ts filename="src/api/index.ts" {4,8-9}
+import { db } from "ponder:api";
+import schema from "ponder:schema";
+import { Hono } from "hono";
+import { graphql } from "ponder";
+
+const app = new Hono();
+
+app.use("/", graphql({ db, schema }));
+app.use("/graphql", graphql({ db, schema }));
```
typedarray, line 35: This may actually cause issues (instantiating the middleware twice)?

examples/feature-accounts/src/api/index.ts
```diff
@@ -0,0 +1,10 @@
+import { db } from "ponder:api";
+import schema from "ponder:schema";
+import { Hono } from "hono";
+import { graphql } from "ponder";
+
+const app = new Hono();
+
+app.use("/graphql", graphql({ db, schema }));
```
typedarray, line 8: Hmm - I suspect we'll want to register this at the root too for all these examples?

packages/core/src/build/index.ts
```diff
@@ -31,6 +31,11 @@ import { safeBuildPre } from "./pre.js";
 import { safeBuildSchema } from "./schema.js";
 import { parseViteNodeError } from "./stacktrace.js";
 
+declare global {
+  var PONDER_DATABASE_SCHEMA: string | undefined;
+  var PONDER_READONLY_DB: Drizzle<Schema>;
+}
```
typedarray, line 37: Will including this here make those properties appear in userland autocomplete?

packages/core/src/build/index.ts
```diff
@@ -197,201 +174,197 @@ export const createBuild = async ({
...
+          contentHash,
+        },
+      };
+    },
+    async executeApi({ database }): Promise<ApiResult> {
+      global.PONDER_READONLY_DB = database.drizzle;
+
+      if (!fs.existsSync(common.options.apiFile)) {
+        const error = new BuildError(
+          `API function file not found. Create a file at ${common.options.apiFile}. Read more: https://ponder.sh/docs/getting-started/database#database-schema"`,
```
typedarray, line 320: Docs link needs update I think

packages/core/src/build/index.ts
```diff
@@ -197,201 +174,197 @@ export const createBuild = async ({
...
-      result: {
-        app: exports.ponder.hono,
-        routes: exports.ponder.routes,
-      },
-    };
-  };
+      const app = executeResult.exports.default;
 
-  let namespace = common.options.schema ?? process.env.DATABASE_SCHEMA;
+      if (app instanceof Hono === false) {
```
typedarray, line 351: Any possibility of Hono version issues here?

packages/core/src/graphql/middleware.test.ts
```diff
@@ -48,7 +39,10 @@ test("middleware serves request", async (context) => {
     bigint: 0n,
   });
 
-  const app = new Hono().use(contextMiddleware).use("/graphql", graphql());
+  const app = new Hono().use(
+    "/graphql",
+    graphql({ schema, db: database.drizzle }),
```
typedarray, line 44: Should this be the readonly drizzle? Or is that in a different PR 😅

packages/core/src/graphql/middleware.ts
```diff
@@ -35,37 +46,42 @@ export const graphql = (
...
+  fs.writeFileSync(
+    path.join(process.cwd(), "generated", "schema.graphql"),
+    printSchema(graphqlSchema),
+    "utf-8",
+  );
+
+  // common.logger.debug({
+  //   service: "codegen",
+  //   msg: "Wrote new file at generated/schema.graphql",
+  // });
```
typedarray, line 61: This seems fine to include?

packages/core/src/graphql/middleware.ts
```diff
@@ -35,37 +46,42 @@ export const graphql = (
     maxOperationAliases: 30,
   },
 ) => {
-  let yoga: YogaServerInstance<any, any> | undefined = undefined;
+  const graphqlSchema = buildGraphQLSchema(schema);
```
typedarray, line 49: So, if this errors (which it can) that will bubble up when we execute the API functions during the build, correct?

Want to make sure we aren't running these effects only when the user hits the endpoint.

packages/core/src/server/index.ts
```diff
@@ -81,14 +75,6 @@ export async function createServer({
...
-  const contextMiddleware = createMiddleware(async (c, next) => {
-    c.set("db", database.drizzle);
-    c.set("metadataStore", metadataStore);
-    c.set("graphqlSchema", schemaBuild.graphqlSchema);
-    await next();
-  });
-
   const hono = new Hono()
     .use(metricsMiddleware)
     .use(cors({ origin: "*", maxAge: 86400 }))
```
typedarray, line 80: Are we still validating that the user hasn't registered routes that would override the builtins? `metrics`, `health`, and `ready` ?

packages/core/src/server/index.ts
```diff
@@ -189,12 +119,12 @@ export async function createServer({
     const httpServer = serve(
       {
         fetch: hono.fetch,
-        createServer: createServerWithNextAvailablePort,
-        port,
+        createServer: http.createServer,
+        port: apiBuild.port,
```
typedarray, line 123: Fine to ship, but flagging again that this can fail here if the port has been taken between build and this line.

typedarray (commented):
Looking great, mostly nits

packages/core/src/graphql/middleware.ts
kyscott18: No access to `common`

packages/core/src/graphql/middleware.test.ts
kyscott18: different pr

packages/core/src/graphql/middleware.ts
kyscott18: Correct

docs/pages/docs/query/graphql.mdx
kyscott18: I'm not sure, we've been doing this for a while, haven't seen any issues

PR #1353, opened 2024-12-16
create a readonly database role

kyscott18:


packages/core/src/database/index.ts
```diff
@@ -215,22 +232,64 @@ export const createDatabase = ({
...
+    await internal.query(`CREATE SCHEMA IF NOT EXISTS "${preBuild.namespace}"`);
+    const hasRole = await internal
+      .query("SELECT FROM pg_roles WHERE rolname = $1", [role])
+      .then(({ rows }) => rows[0]);
+    if (hasRole) {
+      await internal.query(`DROP OWNED BY ${role}`);
+      await internal.query(`DROP ROLE IF EXISTS ${role}`);
+    }
+    await internal.query(
+      `CREATE ROLE ${role} WITH LOGIN PASSWORD '${connection.password}'`,
```
typedarray, line 257: We should probably use a different password than the one from the main/admin/superuser connection string. Looks like PG requires a non-null/non-empty password, so we could use something like "public" or "readonly", or generate a less predictable one. cc @jaylmiller

packages/core/src/database/index.ts
```diff
@@ -215,22 +232,64 @@ export const createDatabase = ({
...
+    );
+    await internal.query(
+      `GRANT CONNECT ON DATABASE "${connection.database}" TO ${role}`,
+    );
+    await internal.query(`REVOKE ALL ON SCHEMA public FROM ${role}`);
```
typedarray, line 262: Will this fail if the `public` schema doesn't exist?

packages/core/src/database/index.ts
```diff
@@ -106,30 +106,34 @@ type PGliteDriver = {
...
-  /** Used in api functions */
-  readonly: HeadlessKysely<unknown>;
   /** Used to interact with the sync-store */
   sync: HeadlessKysely<PonderSyncSchema>;
+  /** Used in api functions */
+  readonly: HeadlessKysely<unknown>;
+  /** Used in client queries */
+  // client: HeadlessKysely<unknown>;
+  drizzle: Drizzle<Schema>;
+  drizzleClient: Drizzle<Schema>;
```
typedarray, line 125: Would find this clearer as `drizzleReadonly`, "client" makes it seem like it's a database client/connection object (vs a pool).

typedarray (commented):
Looking great

packages/core/src/database/index.ts
jaylmiller: default to something predictable but make option to override would be good

packages/core/src/database/index.ts
```diff
@@ -215,22 +232,64 @@ export const createDatabase = ({
...
+    );
+    await internal.query(
+      `GRANT CONNECT ON DATABASE "${connection.database}" TO ${role}`,
+    );
+    await internal.query(`REVOKE ALL ON SCHEMA public FROM ${role}`);
+    await internal.query(
+      `REVOKE ALL ON ALL TABLES IN SCHEMA public FROM ${role}`,
+    );
+    await internal.query(
+      `GRANT USAGE ON SCHEMA ${preBuild.namespace} TO ${role}`,
```
jaylmiller, line 267: add `"` around namespace

packages/core/src/database/index.ts
```diff
@@ -204,6 +213,14 @@ export const createDatabase = ({
         },
         plugins: [new WithSchemaPlugin("ponder_sync")],
       }),
+      drizzle: drizzlePglite((driver as PGliteDriver).instance, {
+        casing: "snake_case",
```
typedarray, line 217: Is this not the default?

packages/core/src/database/index.ts
```diff
@@ -151,6 +155,11 @@ export const createDatabase = ({
 
     const kyselyDialect = new KyselyPGlite(driver.instance).dialect;
 
+    await driver.instance.query(
+      `CREATE SCHEMA IF NOT EXISTS "${preBuild.namespace}"`,
+    );
+    await driver.instance.query(`SET search_path TO ${preBuild.namespace}`);
```
typedarray, line 161: Do we need to escape the namespace here with double quotes? Same question for a few queries below in the role setup.

packages/core/src/database/index.ts
```diff
@@ -215,22 +232,58 @@ export const createDatabase = ({
...
+    await internal.query(
+      `GRANT CONNECT ON DATABASE "${connection.database}" TO ${role}`,
+    );
+    await internal.query(
+      `GRANT USAGE ON SCHEMA "${preBuild.namespace}" TO ${role}`,
+    );
+    await internal.query(
+      `ALTER DEFAULT PRIVILEGES IN SCHEMA "${preBuild.namespace}" GRANT SELECT ON TABLES TO ${role}`,
+    );
+    await internal.query(
```
typedarray, line 266: Is there a way to do a multi-statement raw query here? Would be easier to read and might be faster.

packages/core/src/database/index.ts
```diff
@@ -215,22 +232,58 @@ export const createDatabase = ({
...
       readonly: createPool({
         ...preBuild.databaseConfig.poolConfig,
+        connectionString: undefined,
         application_name: `${preBuild.namespace}_readonly`,
         max: readonlyMax,
+        user: role,
+        password: "pw",
+        host: connection.host ?? undefined,
+        port: Number(connection.port!),
+        database: connection.database ?? undefined,
```
typedarray, line 286: If this is actually nullable, we might have a problem above in the role name. But I'm pretty sure PG conn strings must explicitly set a database name, so might just be a bad type.

packages/core/src/database/index.ts
```diff
@@ -106,30 +106,34 @@ type PGliteDriver = {
...
+  /** Used in api functions */
+  readonly: HeadlessKysely<unknown>;
+  /** Used in client queries */
+  // client: HeadlessKysely<unknown>;
+  drizzle: Drizzle<Schema>;
+  drizzleReadonly: Drizzle<Schema>;
 };
 
-export const createDatabase = ({
+export const createDatabase = async ({
```
typedarray, line 128: Small nit - it's getting difficult to remember what happens in `createDatabase` vs `database.setup`. Maybe need a new name for setup - `prepareNamespace` ?

packages/core/src/database/index.ts
kyscott18: I don't think so, from what I understand the client only allows one statement per query

packages/core/src/database/index.ts
kyscott18: I don't think it is

PR #1354, opened 2024-12-17
support duplicate fragments

kyscott18:
Fixes a bug where duplicate fragments (defining the same block filter twice) would cause Ponder to crash


packages/core/src/sync-store/index.ts
```diff
@@ -230,7 +250,7 @@ export const createSyncStore = ({
       for (let i = 0; i < filters.length; i++) {
         const filter = filters[i]!;
         const intervals = rows
-          .filter((row) => row.filter === `${i}`)
+          .filter((row) => +row.filter === i)
```
kyscott18, line 253: @typedarray I can't figure out why this `row.filter` is a now a number

typedarray (approved):
LGTM

PR #1355, opened 2024-12-17
fix get intervals and github ci

kyscott18:


packages/core/src/graphql/index.test.ts
```diff
@@ -217,7 +217,7 @@ test("scalar, scalar not null, scalar array, scalar array not null", async (cont
   await cleanup();
 });
 
-test("enum, enum not null, enum array, enum array not null", async (context) => {
```
kyscott18, line 220: @typedarray I'm not sure what the pglite test setup is doing to make this fail

PR #1361, opened 2024-12-19
Add configurable block range limits for network syncing

3commascapital:
- Introduce maxBlockRange and minBlockRange options in NetworkConfig
- Update buildConfigAndIndexingFunctions to use new options
- Modify localHistoricalSyncGenerator to use configurable block range limits
- Set default values: maxBlockRange (100,000) and minBlockRange (25)


kyscott18:
We're gonna stick with the dynamic logic, if there is a provider that is not working we can fix it

https://github.com/ponder-sh/ponder/blob/c2f280b9425bb7303b07b5dfb5fc099861875cd2/packages/utils/src/getLogsRetryHelper.ts

Issue #1286, opened 2024-11-26
fix: future end block causes error


kyscott18:
## Description
Ponder allows for passing an end block that is greater than the current chain head. However, when this happens the end block is incorrectly requested. Requesting a block in the futures causes a "not found" error.

https://github.com/ponder-sh/ponder/blob/afba4c361ec70b3ab5de60e6b82c8112331cc185/packages/core/src/sync/index.ts#L936-L943

Solving this will require an update to the checkpointing logic (`getChainCheckpoint` + `getOmnichainCheckpoint`)

PR #1370, opened 2024-12-21
feat: future endBlock

khaidarkairbek:
Relates to #1286. This fix handles the future endBlock by considering it as 
```ts
{
  number: toHex(end),
  hash: "0x",
  parentHash: "0x",
  timestamp: toHex(maxCheckpoint.blockTimestamp),
} as LightBlock
```

### DIFF
--- a/.changeset/weak-lemons-travel.md
+++ b/.changeset/weak-lemons-travel.md
@@ -1,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Fixed a bug causing future end blocks to error.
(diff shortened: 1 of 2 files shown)


typedarray:
Opened an issue related to this: https://github.com/ponder-sh/ponder/issues/1377

kyscott18 (approved):
This is really clever. The end block is also used to potentially shutdown the realtime service once it is reached, and I think what you came up with even works for that codepath.

Issue #1372, opened 2024-12-23
Contract log filters


khaidarkairbek:
### Version

0.8.8

### Current behavior

Currently, filter field for contracts is restrictive, allowing only either filtering multiple events with no arguments or only one event with or without arguments. This restricts the following functionality: filtering multiple different events with arguments or filtering one event with multiple combination of arguments. 

### Expected behavior

I am suggesting to build on existing filter field adding a functionality to specify the array of filters, with identical api. So it would look something as follows: 
```ts
{
  filter?: {
    event: string; 
    args?: <arg types>
  } | {
    event: string[]; 
  } | ({
    event: string; 
    args?: <arg types>
  } | {
    event: string[];
  })[]
}
```

kyscott18:
While we are at it I think we should remove the `event: string[]` option. It doesn't actually do anything because Ponder looks at which indexing functions are registered and filters down based on that.

khaidarkairbek:
> While we are at it I think we should remove the `event: string[]` option. It doesn't actually do anything because Ponder looks at which indexing functions are registered and filters down based on that.

That sounds good. Currently, we are validating that no indexing function is defined for the events excluded by the filter. If we are looking into filtering certain event by the arguments, it would be a bit inconvenient from devx to include other events with no particular argument filtering in the config. Do you mean we are to include the logSources for the event if there is an indexing function defined for it? Then, we would need to remove that validation.

kyscott18:
I think that we should use registered indexing functions as topic0 filtering, and anything defined in the `filters` option as topic1/2/3. 

Also, if there is a filter defined for an event that doesn't have a registered indexing function, that should probably be an error

PR #1373, opened 2024-12-23
Contract source filter

khaidarkairbek:
Relates to #1372. Added a support for more flexible contract source filters in build. Currently, I am assuming the type of the filter as defined in #1372. If we land on this design for an api, the next step would be dealing with `GetEventFilter` type in the config.


packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -361,36 +361,59 @@ export async function buildConfigAndIndexingFunctions({
...
-        const topic0FromFilter = topics.topic0;
-        topic1 = topics.topic1;
-        topic2 = topics.topic2;
-        topic3 = topics.topic3;
+        const filteredTopicsArray = buildTopics(
+          source.abi,
+          source.filter,
+        ).reduce(
+          (acc, cur) => {
+            if (acc.includes(cur) === false) {
```
kyscott18, line 369: Not sure if this `.includes()` works how you expect on objects

packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -322,32 +322,32 @@ export async function buildConfigAndIndexingFunctions({
...
+      }[] = [
+        {
+          topic0: registeredEventSelectors,
+          topic1: null,
+          topic2: null,
+          topic3: null,
+        },
+      ];
 
       if (source.filter !== undefined) {
```
kyscott18, line 339: I'm finding this code block a little hard to follow. I think you have it right: multiple filters in the config should result to multiple sources. Might be helpful to try to limit the use of `.reduce`

packages/core/src/config/eventFilter.ts
```diff
@@ -1,53 +1,50 @@
...
+// Filters are only for topic1/2/3 filtering
 export type GetEventFilter<
   abi extends Abi,
   contract,
   ///
   safeEventNames extends string = SafeEventNames<abi>,
 > = contract extends {
   filter: {
-    event: infer event extends readonly string[] | string;
+    event: infer event extends string;
```
kyscott18, line 13: Need to add a branch for array of filters

packages/core/src/config/eventFilter.test-d.ts
```diff
@@ -27,81 +27,89 @@ test("event filter with no event", () => {
...
+        args: {
+          arg: "0x1",
+        },
+      },
+    ],
   });
 });
 
-test("event filter with event and args", () => {
+test("event filters with event only filter", () => {
```
kyscott18, line 97: Nice test

kyscott18 (approved):

PR #1383, opened 2024-12-30
Granular `sync-historical`

kyscott18:


packages/core/src/sync-historical/index.ts
```diff
@@ -592,71 +597,96 @@ export const createHistoricalSync = async (
...
+        ];
+
+        const completedIntervals = intervalsCache.get(filter)!;
+        const requiredIntervals: [Fragment, Interval[]][] = [];
+
+        for (const {
+          fragment,
+          intervals: fragmentIntervals,
+        } of completedIntervals) {
+          const requiredFragmentIntervals = intervalDifference(
```
khaidarkairbek, line 629: Can this be improved with Promise.all as there are no side effects in the loop?

packages/core/src/sync-historical/index.ts
```diff
@@ -592,71 +597,96 @@ export const createHistoricalSync = async (
...
+        if (
+          (filter.fromBlock !== undefined && filter.fromBlock > _interval[1]) ||
+          (filter.toBlock !== undefined && filter.toBlock < _interval[0])
+        ) {
+          continue;
+        }
+
+        const interval: Interval = [
+          Math.max(filter.fromBlock ?? 0, _interval[0]),
+          Math.min(filter.toBlock ?? Number.POSITIVE_INFINITY, _interval[1]),
```
khaidarkairbek, line 619: A bit of side note, I was thinking of resolving fromBlock and toBlock values to numbers for filters at the sync module before the ponder goes to sync-historical or sync-realtime. This should simplify the logic, as currently we are just mapping undefined values to 0 or Number.POSITIVE_INFINITY.

kyscott18: I think I could get behind it: removing `| undefined` from all the filter types and having `Number.POSITIVE_INFINITY` represent a filter that never ends? 

@typedarray

packages/core/src/sync-historical/index.ts
kyscott18: I don't it would be a different because it is synchronous

packages/core/src/sync-historical/index.ts
khaidarkairbek: > I think I could get behind it: removing `| undefined` from all the filter types and having `Number.POSITIVE_INFINITY` represent a filter that never ends?
> 
> @typedarray

I have started on this cleanup at multiple block range pr if its not an issue.

Issue #1377, opened 2024-12-26
feat: more than one block range for a contract


typedarray:
### Problem / use case

Sometimes, it's useful to skip certain block ranges when indexing a contract. For example, the Hyperliquid testnet appears to have some block ranges where the RPC does not work.

Currently you can only pass `startBlock` and `endBlock`.

### Proposed solution

We could replace `startBlock` and `endBlock` with a `blocks` field that supports multiple ranges. Also, this could be a good time to introduce tags like `"latest"` which would represent the latest block when the server starts up (very common for block intervals / cron).

#### Simple bounded range
```diff
-  startBlock: 1000,
-  endBlock: 2000,
+  blocks: [1000, 2000],
```

#### Multiple ranges
```diff
+  blocks: [[1000, 2000], [2050, "latest"]],
```

#### No endBlock (live indexing)
```diff
-  startBlock: 1000,
+  blocks: [1000, undefined],
```

### Open questions

1. How should the (very common) case of `endBlock: undefined` look? Perhaps we could come up with a tag like "latest". Could be "live", "realtime", or "latest" if we use something else to mean "the latest block when the server started up".

PR #1384, opened 2025-01-02
Feat: Multiple Block Ranges

khaidarkairbek:
Relates to #1377. Handles multiple block ranges by pointing each resolved block range to its separate source.


packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -24,11 +24,15 @@ import {
...
 import { chains } from "@/utils/chains.js";
+import { type Interval, intervalUnion } from "@/utils/interval.js";
 import { toLowerCase } from "@/utils/lowercase.js";
+import { _eth_getBlockByNumber } from "@/utils/rpc.js";
 import { dedupe } from "@ponder/common";
-import type { Hex, LogTopic } from "viem";
+import { type Hex, type LogTopic, hexToNumber } from "viem";
 import { buildLogFactory } from "./factory.js";
 
+type BlockRange = [number | "latest", number | "realtime" | "latest"];
```
kyscott18, line 34: Could you add some comments here about what each of these values mean?

packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -59,6 +63,37 @@ const flattenSources = <
...
+        ? blocks.map(([fromBlock, toBlock]) => [
+            fromBlock === "latest" ? latest : fromBlock,
+            toBlock === "latest" ? latest : toBlock,
+          ])
+        : [
+            [
+              blocks[0] === "latest" ? latest : blocks[0],
+              blocks[1] === "latest" ? latest : blocks[1],
+            ],
+          ];
```
kyscott18, line 83: Looks like there is a type error in here

packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -113,6 +150,14 @@ export async function buildConfigAndIndexingFunctions({
         finalityBlockCount: getFinalityBlockCount({ chainId }),
         disableCache: network.disableCache ?? false,
       } satisfies Network;
+
+      const latest: Hex = await network.transport({ chain }).request({
+        method: "eth_blockNumber",
+      });
```
kyscott18, line 156: Does this request happen even if there are no tags used in `blocks`?

khaidarkairbek: fixed it

packages/core/src/build/configAndIndexingFunctions.ts
khaidarkairbek: fixed it

PR #1385, opened 2025-01-02
Update api-functions.mdx

jagnani73:
Fixed breaking imports for generated -> registry and core->ponder


kyscott18:
It looks like the docs page you updated was for v0.6, so it shouldn't be updated.

PR #1397, opened 2025-01-05
fix: duplicate logs

kyscott18:
## Description
The factory child address logic was not deduplicating its results, leading to `syncLogsDynamic` being called with many duplicate values. This function has some address batching logic, which would cause multiple replica eth_getLogs requests to be made. The resulting logs failed when being inserted into the database.

## Example
- factory contract with many duplicated child addresses

### DIFF
--- a/.changeset/stale-timers-remain.md
+++ b/.changeset/stale-timers-remain.md
@@ -1,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Fixed a bug resulting in `error: ON CONFLICT DO UPDATE command cannot affect row a second time`.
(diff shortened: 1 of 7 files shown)


tk-o:
Awesome update, @kyscott18 🚀

typedarray:
Nice find. To clarify - I see two changes here:
1. When a user passes an array of hard-coded addresses in their config, de-duplicate the array while building the source/filter.
2. When fetching child addresses from the sync store, use `distinct` to de-duplicate the result array (because its possible for a factory to emit multiple logs specifying the same child address).

The user bug reports were caused by the factory case during the historical sync - we would fetch an array of child addresses and include it in an eth_getLogs request. If that array contains duplicates, a compliant RPC should not return duplicate logs. But, we actually separate the child addresses into batches (50?) for eth_getLogs. So, we had a situation where the same address would be included in multiple batches, leading to duplicate logs even for compliant RPCs. Is that correct?

kyscott18:
Correct

PR #1408, opened 2025-01-06
live queries

kyscott18:


packages/client/src/index.ts
```diff
@@ -43,19 +53,68 @@ export const createClient = <schema extends Schema>(
...
+          query: JSON.stringify(builtQuery),
+        }).toString()}`,
+      );
+
+      sse.onmessage = (event) => {
+        // @ts-ignore
+        callback(prepared.mapResult(JSON.parse(event.data), true).rows);
+      };
+
+      sse.onerror = () => {
```
typedarray, line 107: Perhaps the API should have an onError callback? Does the EventSource API include the event object here?

packages/client/src/index.ts
```diff
@@ -43,19 +53,68 @@ export const createClient = <schema extends Schema>(
...
+
+      const sse = new EventSource(
+        `${url}/client/live?${new URLSearchParams({
+          query: JSON.stringify(builtQuery),
+        }).toString()}`,
+      );
+
+      sse.onmessage = (event) => {
+        // @ts-ignore
+        callback(prepared.mapResult(JSON.parse(event.data), true).rows);
```
typedarray, line 104: Any error handling to consider here as well?

packages/core/src/client/index.ts
```diff
@@ -1,31 +1,110 @@
...
+
+      return streamSSE(
+        c,
+        async (stream) => {
+          stream.onAbort(() => {
+            stream.close();
+          });
+
+          while (stream.closed === false) {
+            const query = JSON.parse(c.req.query("query")!) as QueryWithTypings;
```
typedarray, line 62: Should probably return a 4xx status code if the `query` query param is not present or invalid. And should do this before you enter `streamSSE` I think.

packages/core/src/client/index.ts
```diff
@@ -1,31 +1,110 @@
...
+            const result = await session
+              .prepareQuery(query, undefined, undefined, false)
+              .execute();
+
+            // no-op if result is the same
+
+            if (
+              liveQueries.has(streamId) &&
+              // @ts-ignore
+              hash(result.rows) === liveQueries.get(streamId)!.resultHash
```
typedarray, line 73: Up for debate, but I think it makes more sense to send the result even if the result is unchanged, or make it opt-in. With this implementation, I think there's a higher chance it will look "broken" to users.

packages/core/src/client/index.ts
```diff
@@ -1,31 +1,110 @@
...
 
-    // @ts-ignore
-    const res = await db._.session
-      .prepareQuery(
-        {
-          sql: body.sql,
-          params: body.params,
-          // @ts-ignore
-          typings: body.typings,
+      const streamId = id++;
```
typedarray, line 52: What does the streamId do?

packages/core/src/client/index.ts
```diff
@@ -1,31 +1,110 @@
...
+const POLL_INTERVAL = 500;
 
 export const client = ({ db }: { db: ReadonlyDrizzle<Schema> }) => {
+  let id = 0;
+
   return createMiddleware(async (c, next) => {
-    if (c.req.path !== "/client") {
-      return next();
+    // @ts-ignore
+    const session: PgSession = db._.session;
```
typedarray, line 14: What does this do? Why not use `db` directly?

packages/client/src/index.ts
```diff
@@ -43,19 +53,68 @@ export const createClient = <schema extends Schema>(
...
+  // @ts-ignore
+  const dialect: PgDialect = noopDatabase.dialect;
+
+  // @ts-ignore
+  const session: PgSession = noopDatabase.session;
+
+  return {
+    db: drizzle(
+      async (sql, params, method, typings) => {
+        const result = await fetch(`${url}/client/db`, {
```
typedarray, line 72: You might want to use the `URL` Web API here to handle cases where `url` has a trailing slash or already has query/search params.

packages/client/src/index.ts
```diff
@@ -39,23 +76,122 @@ export const status = pgTable("_ponder_status", (t) => ({
...
+          },
+        };
+      } else {
+        return client.live(
+          (db) => db.select().from(status),
+          () => {
+            _query(client.db).then(onData).catch(onError);
+          },
+          onError,
+        );
```
typedarray, line 189: Slightly strange block lol but I follow

packages/core/src/bin/commands/dev.ts
```diff
@@ -255,7 +276,9 @@ export async function dev({ cliOptions }: { cliOptions: CliOptions }) {
   });
 
   let database: Database | undefined;
+  let listenConnection: ListenConnection | undefined;
 
+  build.initNamespace({ isSchemaRequired: false });
```
typedarray, line 281: duplicate

packages/core/src/bin/commands/start.ts
```diff
@@ -52,9 +52,16 @@ export async function start({ cliOptions }: { cliOptions: CliOptions }) {
   const cleanup = async () => {
     await cleanupReloadable();
     await cleanupReloadableServer();
+    if (listenConnection) {
+      if (listenConnection.dialect === "postgres") {
+        listenConnection.connection.release();
+      }
+    }
```
typedarray, line 59: Why is listenConnection not a property of `database` ? They seem to be passed around together.

packages/core/src/client/index.ts
```diff
@@ -1,31 +1,113 @@
...
+
+      try {
+        const result = await session
+          .prepareQuery(query, undefined, undefined, false)
+          .execute();
+
+        return c.json(result as object);
+      } catch (error) {
+        return c.text((error as Error).message, 500);
+      }
```
typedarray, line 78: May want to strip some of the details from this error object before returning, but can add this after prerelease.

packages/core/src/client/index.ts
```diff
@@ -1,31 +1,113 @@
...
+            });
+          } catch (error) {
+            await stream.writeSSE({
+              data: JSON.stringify({
+                status: "error",
+                error: (error as Error).message,
+              }),
+            });
+          }
+          statusResult = await statusResolver.promise;
```
typedarray, line 106: This waits for the next status update?

packages/core/src/client/index.ts
```diff
@@ -1,31 +1,113 @@
...
+ * const app = new Hono();
+ *
+ * app.use(client({ db }));
+ *
+ * export default app;
+ * ```
+ */
 export const client = ({ db }: { db: ReadonlyDrizzle<Schema> }) => {
+  // @ts-ignore
+  const session: PgSession = db._.session;
```
typedarray, line 36: Does this mean all queries will use the same connection? Seems important to use a pool here.

packages/core/src/database/index.ts
```diff
@@ -1115,8 +1008,116 @@ export const createDatabase = async ({
...
+
+        await sql
+          .raw(`
+      CREATE OR REPLACE TRIGGER ${trigger}
+      AFTER INSERT OR UPDATE OR DELETE
+      ON "${preBuild.namespace}"._ponder_status
+      FOR EACH STATEMENT
+      EXECUTE PROCEDURE ${notification};`)
+          .execute(qb.internal);
+      });
```
typedarray, line 1039: My instinct is that these should happen in the same spot that we create the `_ponder_status` table (if not exists).

packages/core/src/database/index.ts
```diff
@@ -1115,8 +1008,116 @@ export const createDatabase = async ({
...
+
+      if (dialect === "postgres") {
+        return {
+          dialect: "postgres",
+          connection: await (driver as PostgresDriver).internal.connect(),
```
typedarray, line 1044: What happens if this connection drops?

typedarray (commented):
Looks great, nothing blocking prerelease.

packages/core/src/client/index.ts
kyscott18: yes

packages/core/src/database/index.ts
kyscott18: The problem is that the function that creates the status table (`migrate()`) isn't always run when `getListenConnection()` is run. For example the `serve` command. We might be able to come up with a better solution than I have now but that was my original reason for making this change.

packages/core/src/client/index.ts
kyscott18: Variable wasn't being used, deleted

Issue #1376, opened 2024-12-26
Failed 'Nft.flush()'


pursue-wind:
### Version

0.7.17

### Current behavior

**This error only occurs on the Linux server. I've tried versions 0.7 and 0.8, and both have this error. Using a new database doesn't solve the problem either.**
It had been working properly all along, but suddenly this error occurred when starting it up. However, it starts up normally on my local Mac. 
```
7:10:19 AM ERROR indexing   Internal error occurred while flushing cache. Please report this error here: https://github.com/ponder-sh/ponder/issues
7:10:19 AM WARN  user       Failed 'Nft.flush()' database method 
7:10:19 AM ERROR process    Caught unhandledRejection event
FlushError: invalid byte sequence for encoding "UTF8": 0x00
    at file:///home/ubuntu/nft_index_2_test/node_modules/@ponder/common/src/queue.ts:184:15
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at Object.flush (file:///home/ubuntu/nft_index_2_test/node_modules/@ponder/core/src/indexing-store/historical.ts:754:7)
    at start3 (file:///home/ubuntu/nft_index_2_test/node_modules/@ponder/core/src/bin/utils/run.ts:188:9)

7:10:19 AM WARN  process    Received unhandledRejection, starting shutdown sequence
7:10:19 AM ERROR process    Caught unhandledRejection event
FlushError: invalid byte sequence for encoding "UTF8": 0x00
    at file:///home/ubuntu/nft_index_2_test/node_modules/@ponder/common/src/queue.ts:184:15
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at Object.flush (file:///home/ubuntu/nft_index_2_test/node_modules/@ponder/core/src/indexing-store/historical.ts:754:7)
    at start3 (file:///home/ubuntu/nft_index_2_test/node_modules/@ponder/core/src/bin/utils/run.ts:188:9)
```

typedarray:
Thanks for opening - need a bit more information:
1. Can you share the indexing function body where you insert `Nft` records? (if there are multiple, the most common one)
2. And if so, can you share a link to the contract that emitted the event?

We believe the null characters are coming from the decoded event body, for example in a free-text field of the event like "author" or "graffiti". I'm looking for a link to the specific event that might include the null characters so we can reproduce and fix this. Thanks.

pursue-wind:
[...]

PR #1414, opened 2025-01-07
fix: null bytes

### LINKED ISSUE #1378: bug: null characters throw error when inserted into `text` columns
### Related

https://github.com/ponder-sh/ponder/issues/1376

### Version

0.8.8, 0.7.17

### Current behavior

The server crashes the error (or a similar error for ASCII encoding):
```
FlushError: invalid byte sequence for encoding "UTF8": 0x00
```

### Expected behavior

Flush succeeds regardless of null characters

### Steps to reproduce

Don't have a repro handy, but in general this occurs because EVM `string` values can contain null characters, but the Postgres `text` data type does not allow them. We've seen null characters in decoded event `data` (very likely the source of the user reports we're seeing) as well as raw RPC fields like `trace.revertReason`.

### Link to repository

_No response_

### Anything else?

Two possible solutions:
1. Ensure that any values within the `event` object, particularly decoded event arguments (e.g. `event.args.{xxx}`), do not contain null characters by sanitizing them with `str.replace(/\0/g, '')`.
2. Sanitize any value being inserted into a `text` column at the driver level.

I prefer 1, at least as a starting point. On the raw RPC side of things, we should probably sanitize specific free text fields like revertReason before insertion.

typedarray:
It's possible for decoded values for `string` ABI parameters to contain null characters. Null characters are ignored by almost all software systems, except the postgres `TEXT` data type does not accept them ([context](https://www.commandprompt.com/blog/null-characters-workarounds-arent-good-enough)).

This PR removes null characters from all decoded ABI strings.

Fixes #1378 #1376

### DIFF
--- a/packages/core/src/sync/events.ts
+++ b/packages/core/src/sync/events.ts
@@ -500,7 +500,7 @@ export const decodeEvents = (
 
                 event: {
                   name: safeName,
-                  args,
+                  args: removeNullCharacters(args),
                   log: event.log!,
                   block: event.block,
                   transaction: event.transaction!,
@@ -556,8 +556,8 @@ export const decodeEvents = (
                 name: `${source.name}.${safeName}`,
 
                 event: {
-                  args,
-                  result,
+                  args: removeNullCharacters(args),
+                  result: removeNullCharacters(result),
                   trace: event.trace!,
                   block: event.block,
                   transaction: event.transaction!,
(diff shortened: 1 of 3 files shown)


packages/core/src/sync/events.ts
```diff
@@ -500,7 +500,7 @@ export const decodeEvents = (
 
                 event: {
                   name: safeName,
-                  args,
+                  args: removeNullCharacters(args),
```
typedarray, line 503: @kyscott18 lmk if this is the correct place to do this - both historical and realtime events pass through this, yes?

kyscott18: Yes

kyscott18 (approved):

PR #1418, opened 2025-01-08
fix: removing unnecessary _eth_getBlockReceipts requests

khaidarkairbek:
Adding a check to not process _eth_getBlockReceipts if there are no transaction receipts needed for this block.


typedarray (approved):
LGTM, needs a changeset like

```
Fixed a bug introduced in 0.8.14 which caused more `eth_getBlockReceipts` requests to be made than is necessary.
```

PR #1424, opened 2025-01-08
fix: indexing error regression

typedarray:


packages/core/src/indexing/service.ts
```diff
@@ -377,17 +379,23 @@ const executeSetup = async (
...
+    } else {
+      error.meta = [toErrorMeta(event)];
+    }
+
     common.metrics.ponder_indexing_has_error.set(1);
 
     common.logger.error({
       service: "indexing",
-      msg: `Error while processing '${event.name}' event in '${networkByChainId[event.chainId]!.name}' block ${decodedCheckpoint.blockNumber}`,
+      msg: `Error while processing event '${event.name}' event in '${networkByChainId[event.chainId]!.name}' block ${decodedCheckpoint.blockNumber}`,
```
kyscott18, line 398: too many uses of "event"

packages/core/src/indexing/service.ts
```diff
@@ -489,3 +470,80 @@ const executeEvent = async (
...
+const transactionText = (transaction: Transaction) =>
+  `Transaction:\n${prettyPrint({
+    hash: transaction.hash,
+    from: transaction.from,
+    to: transaction.to,
+  })}`;
+
+const logText = (log: Log) =>
+  `Log:\n${prettyPrint({
+    id: log.id,
```
kyscott18, line 490: Why `id`

PR #1426, opened 2025-01-09
fix: null bytes in `trace.error` and `trace.revertReason`

kyscott18:


typedarray (approved):
LGTM. Could add a changeset. Any other free-text fields in the RPC data model?

PR #1438, opened 2025-01-12
Added `ponder_realtime_latency` metric

kyscott18:


packages/core/src/common/metrics.ts
```diff
@@ -151,10 +152,17 @@ export class MetricsService {
...
     this.ponder_realtime_reorg_total = new prometheus.Counter({
       name: "ponder_realtime_reorg_total",
-      help: "Count of how many re-orgs have occurred.",
+      help: "Count of how many re-orgs have occurred",
       labelNames: ["network"] as const,
       registers: [this.registry],
     });
+    this.ponder_realtime_latency = new prometheus.Histogram({
+      name: "ponder_realtime_latency",
+      help: "Duration of block processing",
```
kyscott18, line 161: "Time elapsed between receiving a block and fully processing it"

packages/core/src/sync/index.ts
```diff
@@ -679,6 +679,22 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
               }
+
+              // update `ponder_realtime_latency` metric
+              for (const network of args.networks) {
+                for (const { block, endClock } of perNetworkSync.get(network)!
+                  .unfinalizedBlocks) {
+                  const checkpoint = encodeCheckpoint(
+                    blockToCheckpoint(block, network.chainId, "up"),
+                  );
+                  if (checkpoint > from && checkpoint <= to && endClock) {
```
typedarray, line 690: So this means, if the checkpoint for this block is within the checkpoint range that we just finished indexing, record the duration measurement?

kyscott18: Yes, meaning that the block has been fully processed

typedarray (approved):

PR #1441, opened 2025-01-14
isolated chain ordering

kyscott18:


packages/core/src/sync/index.ts
```diff
@@ -431,483 +308,487 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
-              target: getEventsMaxBatchSize,
-              result: events.length,
-              min: 10,
-              max: 86_400,
-              prev: estimateSeconds,
-              maxIncrease: 1.08,
-            });
+      for (const network of params.indexingBuild.networks) {
+        updateHistoricalStatus({ events, checkpoint, network });
+      }
```
typedarray, line 380: This could do all chains at once, some duplicate work happening

packages/core/src/sync/index.ts
```diff
@@ -431,483 +308,487 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
+      yield events;
+      cursor = checkpoint;
+    }
+  }
+
+  const getOnRealtimeSyncEvent = () => {
```
typedarray, line 386: Do you need this closure / scope or could the `checkpoints` and `latencyTimers` move to the upper scope and it could just be `const onRealtimeSyncEvent = (event: ...) => { ... }`

packages/core/src/sync/index.ts
```diff
@@ -431,483 +308,487 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
-            consecutiveErrors = 0;
-            yield { events, checkpoint: to };
-            from = cursor;
-          } catch (error) {
-            // Handle errors by reducing the requested range by 10x
-            estimateSeconds = Math.max(10, Math.round(estimateSeconds / 10));
+            status[network.name]!.block = {
+              timestamp: hexToNumber(event.block.timestamp),
+              number: hexToNumber(event.block.number),
+            };
```
typedarray, line 433: Is it premature to update the status here? Should it be in the .then below?

packages/core/src/sync/index.ts
```diff
@@ -431,483 +308,487 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
-          chainId: network.chainId,
-          blocks: event.reorgedBlocks,
-        });
+      const perChainOnRealtimeSyncEvent = getPerChainOnRealtimeSyncEvent({
+        common: params.common,
+        network,
+        sources,
+        syncStore: params.syncStore,
+        syncProgress,
+        realtimeSync,
```
typedarray, line 739: It's a bummer that the `perChainOnRealtimeSyncEvent` needs the `realtimeSync`, creates a bit of indirection.

packages/core/src/sync/index.ts
```diff
@@ -936,243 +817,442 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
+        );
+
+        unfinalizedBlocks = unfinalizedBlocks.filter(
+          ({ block }) =>
+            hexToNumber(block.number) > hexToNumber(event.block.number),
+        );
+
+        // Add finalized blocks, logs, transactions, receipts, and traces to the sync-store.
+
+        await Promise.all([
```
typedarray, line 984: This ought to use a database transaction (including the interval insert statement).

Can tackle alongside the sync store improvements.

packages/core/src/sync/index.ts
```diff
@@ -936,243 +817,442 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
-        intervalIntersectionMany(
-          fragmentIntervals.map(({ intervals }) => intervals),
+  params.common.logger.debug({
+    service: "sync",
+    msg: `Initialized '${params.network.name}' extract query for timestamps [${decodeCheckpoint(params.from).blockTimestamp}, ${decodeCheckpoint(params.to).blockTimestamp}]`,
+  });
+
+  for await (const syncCheckpoint of bufferAsyncGenerator(
+    params.localSyncGenerator,
+    Number.POSITIVE_INFINITY,
```
typedarray, line 1140: So, this is what tells the historical sync to just keep cooking regardless of indexing progress?

packages/core/src/sync/index.ts
```diff
@@ -936,243 +817,442 @@ export const createSync = async (args: CreateSyncParameters): Promise<Sync> => {
...
-    )
-  ) {
-    return _eth_getBlockByNumber(requestQueue, {
-      blockNumber: minCompletedBlock,
-    });
-  }
+        consecutiveErrors = 0;
+        cursor = queryCursor;
+        yield { events, checkpoint: cursor };
+      } catch (error) {
```
typedarray, line 1186: I think we should almost certainly log this error, right? Or is it getting logged elsewhere already?

packages/core/src/sync/index.ts
```diff
@@ -1193,110 +1273,91 @@ export async function* localHistoricalSyncGenerator({
...
+        )} cached`,
+      });
     }
 
-    fromBlock = hexToNumber(syncProgress.cached.number) + 1;
+    cursor = hexToNumber(syncProgress.current.number) + 1;
+  } else {
+    common.logger.info({
+      service: "historical",
+      msg: `Started '${network.name}' historical sync`,
```
typedarray, line 1312: Does this mean there's no progress cached? Would be cool to add that to the log

packages/core/src/utils/zipper.ts
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * @returns The merged array.
+ *
+ * @example
+ * ```ts
+ * const result = zipper([1, 3, 5], [2, 4, 6]);
+ * // result = [1, 2, 3, 4, 5, 6]
+ * ```
+ */
+export const zipper = <T>(
```
typedarray, line 16: Would prob just have one zipper func?

typedarray (commented):
Brilliant!

packages/core/src/sync/index.ts
kyscott18: yes

packages/core/src/sync/index.ts
kyscott18: Agreed

packages/core/src/database/index.ts
```diff
@@ -1012,9 +1008,8 @@ export const createDatabase = async ({
       }, common.options.databaseHeartbeatInterval);
     },
     async recoverCheckpoint() {
-      if (checkpoint !== undefined) {
-        return checkpoint;
-      }
+      // new tables are empty
+      if (createdTables) return ZERO_CHECKPOINT_STRING;
```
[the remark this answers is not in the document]
kyscott18: yes

packages/core/src/bin/utils/run.ts
```diff
@@ -143,81 +227,127 @@ export async function run({
...
+              );
+              common.metrics.ponder_indexing_timestamp.set(
+                { network: network.name },
+                checkpoint.blockTimestamp,
+              );
+            }
+          }
+
+          // Note: allows for terminal and logs to be updated
+          await new Promise(setImmediate);
```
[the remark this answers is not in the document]
kyscott18: yes

PR #1450, opened 2025-01-17
dont drop role if already exists, make sure database name exists, conditional temp_file_limit based on superuser status

jaylmiller:


packages/core/src/database/index.ts
```diff
@@ -240,37 +240,38 @@ export const createDatabase = async ({
     const connection = (parse as unknown as typeof parse.parse)(
       preBuild.databaseConfig.poolConfig.connectionString!,
     );
+    if (!connection.database) {
```
jaylmiller, line 243: we reference connection.database further down in the code so it doesn't make sense to ignore it not existing.

kyscott18 (approved):

PR #1466, opened 2025-01-27
docs: update railway healthcheck timeout

jagnani73:
Railway's maximum accepted healthcheck timeout is 1 hour (3600s), not 1 day (86400s)


typedarray:
Thanks!

PR #1470, opened 2025-01-29
fix: clarify subgraph deployment id, not subgraph id

0xnikter:
When trying to migrate from Uniswap V2 subgraph, got this error: 
image

Actually, we should use the Deployment ID, not the subgraph ID:
image


typedarray:
Thanks!

PR #1485, opened 2025-02-04
config: latest tag

khaidarkairbek:
Added latest tag to the config for startBlock and endBlock of the account, contract and block sources. This would be resolved to the `blockNumber` of the latest block per network at the startup.


packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -73,6 +74,36 @@ export async function buildConfigAndIndexingFunctions({
...
+    }
+
+    if (Number.isNaN(blockNumberOrTag)) {
+      return undefined;
+    }
+    if (blockNumberOrTag === "latest") {
+      if (perNetworkLatestBlockNumber.has(network.name)) {
+        return perNetworkLatestBlockNumber.get(network.name)!;
+      } else {
+        const blockPromise = network.transport
```
typedarray, line 94: We should be thoughtful about error handling here. This is the first risky async work we're doing in this function. Worth a mocked test IMO.

.changeset/late-peas-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Added support for block tags (`startBlock: "latest"`) in `ponder.config.ts`.
```
typedarray, line 5: Added support for passing `"latest"` to `startBlock` and `endBlock` in `ponder.config.ts`. With this option, Ponder will fetch the latest block during the build step and use that block number.

examples/feature-multichain/ponder.config.ts
```diff
@@ -1,62 +1,44 @@
...
       transport: http(process.env.PONDER_RPC_URL_137),
-      pollingInterval: 15_000,
+      pollingInterval: 1_000,
     },
   },
   contracts: {
     weth9: {
       abi: weth9Abi,
       address: "0x4200000000000000000000000000000000000006",
+      startBlock: "latest",
```
typedarray, line 33: Awesome diff in this file.

typedarray (commented):
Looking good.

1. I'd like to add two tests, one for success case to make sure it inserts the latest block number and one for an RPC error when fetching the latest block (eg block not found).

2. With this impl, does the build ID stay the same across reloads when using the "latest" tag?

PR #1486, opened 2025-02-05
wip: build core with tsc

typedarray:
- Migrates `core` to use `tsc` via a custom build script, similar to what we did for `@ponder/react`
- Removes private `common` package, copy/pastes required utility function into `core/src/utils` and `utils/src/utils`
- Use a custom tsconfig-paths replacer script
- Improves sourcemaps, go-to-definition now goes to the actual ts source


packages/core/src/bin/ponder.ts
```diff
@@ -14,7 +14,7 @@ import { start } from "./commands/start.js";
 dotenv.config({ path: ".env.local" });
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
-const packageJsonPath = resolve(__dirname, "../../package.json");
+const packageJsonPath = resolve(__dirname, "../../../package.json");
```
kyscott18, line 17: Was the on purpose?

PR #1489, opened 2025-02-05
chore: improve GraphQL enum not found error message

jaydenwindle:
This error is often thrown when an enum has not been exported from the Ponder schema. This PR updates the error message to make the fix more obvious.


kyscott18 (approved):
Thanks!

PR #1491, opened 2025-02-07
fix: broken link in docs

kanvgupta:
Fixed broken link to Call Traces(/docs/call-traces) on Contracts & Networks page(/docs/contracts-and-networks).


typedarray (approved):
Thanks!

PR #1494, opened 2025-02-08
Improve shutdown behavior and eth_getLogs retry estimate

kyscott18:


packages/core/src/sync/index.ts
```diff
@@ -1167,6 +1167,10 @@ export async function* getLocalEventGenerator(params: {
         cursor = queryCursor;
         yield { events, checkpoint: cursor };
       } catch (error) {
+        if (params.common.shutdown.isKilled) {
+          throw new ShutdownError();
+        }
```
typedarray, line 1172: In what scenario would this error not already be a ShutdownError ?

packages/core/src/sync-historical/index.ts
```diff
@@ -109,18 +110,17 @@ export const createHistoricalSync = async (
   >();
 
   /**
-   * Data about the range passed to "eth_getLogs" for all log
-   *  filters and log factories.
+   * Data about the range passed to "eth_getLogs" share among all log
```
typedarray, line 113: Def agree that this should persist across calls to `sync`, but less sure about having one across all filters. If there's an alchemy-style limit (based on response size rather than block count), and the app has one very dense filter and one very sparse filter, it's a bit suboptimal (right?). Don't feel strongly though.

packages/core/src/sync/index.ts
kyscott18: If the query has a "statement timeout" error

packages/core/src/sync-historical/index.ts
kyscott18: Yes it is a bit suboptimal but I don't think the benefits of trying to have a different range for each filter outweigh the complexity. It is hard because we build filters on each call to `historicalSync.sync()` so we would need to generate a unique id for all the filters.

It's still possible and I don't feel super strongly either way, lmk what you want.

typedarray: Oh I see. Let's ship this as-is.

PR #1499, opened 2025-02-10
feat: database transactions for historical indexing 

kyscott18:


packages/core/src/utils/pg.ts
```diff
@@ -46,9 +46,30 @@ pg.Client.prototype.query = function query(
...
+        });
+      } else {
+        return super
+          .connect()
+          .then(() =>
+            this.query("SET synchronous_commit = off;").then(() => {}),
+          );
+      }
+    }
+  }
```
kyscott18, line 66: @typedarray did I do this correctly?

packages/core/src/database/index.ts
```diff
@@ -1167,45 +1183,156 @@ $$ LANGUAGE plpgsql
...
+            target: PONDER_STATUS.network_name,
+            set: {
+              block_number: sql`excluded.block_number`,
+              block_timestamp: sql`excluded.block_timestamp`,
+              ready: sql`excluded.ready`,
+            },
+          });
+      });
+    },
+    async revert({ checkpoint, db }) {
```
typedarray, line 1244: The `db`/`tx` polymorphism is weird in cases like this, because it's definitely not safe to run this without using a transaction.

packages/core/src/graphql/index.ts
```diff
@@ -65,7 +65,7 @@ import { GraphQLJSON } from "./json.js";
 type Parent = Record<string, any>;
 type Context = {
   getDataLoader: ReturnType<typeof buildDataLoaderCache>;
-  metadataStore: MetadataStore;
+  metadata: Pick<Database, "getStatus">;
```
typedarray, line 68: Would probably just do
```ts
  getDataLoader: ReturnType<typeof buildDataLoaderCache>;
  getStatus: Database["getStatus"];
```

packages/core/src/indexing-store/cache.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    checkpoint: ZERO_CHECKPOINT_STRING,
+  });
+
+  const indexingStore = createHistoricalIndexingStore({
+    common: context.common,
+    database,
+    schemaBuild: { schema },
+    indexingCache,
+    db: database.qb.drizzle,
+  });
```
typedarray, line 43: One idea - you could consider an API that looks like this to manage the `tx` dep injection:
```ts
  const indexingStore = createHistoricalIndexingStore({
    common: context.common,
    database,
    schemaBuild: { schema },
    indexingCache,
    // db: database.qb.drizzle,
  });

await indexingStore.update(database.qb.drizzle, schema.account, { pk: value }).set(...)
// or, in a transaction
await indexingStore.update(tx, schema.account, { pk: value }).set(...)
```

packages/core/src/indexing-store/cache.ts
```diff
@@ -0,0 +1,571 @@
...
+
+type Cache = Map<
+  Table,
+  Map<
+    string,
+    {
+      bytes: number;
+      operationIndex: number;
+      row: { [key: string]: unknown } | null;
+    }
```
typedarray, line 91: Maybe use a tuple? Not sure if it makes a difference

packages/core/src/indexing-store/cache.ts
```diff
@@ -0,0 +1,571 @@
...
+        updateBuffer.get(table)!.get(getCacheKey(table, key)) ??
+        insertBuffer.get(table)!.get(getCacheKey(table, key));
+
+      if (bufferEntry) {
+        return structuredClone(bufferEntry.row);
+      }
+
+      const entry =
+        cache.get(table)!.get(getCacheKey(table, key)) ??
+        spillover.get(table)!.get(getCacheKey(table, key));
```
typedarray, line 290: Is this order correct?

packages/core/src/indexing-store/cache.ts
```diff
@@ -0,0 +1,571 @@
...
+
+      cacheBytes += spilloverBytes;
+      spilloverBytes = 0;
+
+      let cacheSize = 0;
+      for (const { size } of cache.values()) cacheSize += size;
+
+      const flushIndex =
+        totalCacheOps -
+        cacheSize * (1 - common.options.indexingCacheFlushRatio);
```
typedarray, line 517: Should probably be named `indexingCacheEvictRatio`

packages/core/src/indexing-store/historical.ts
```diff
@@ -390,50 +69,50 @@ export const createHistoricalIndexingStore = ({
...
+                      const row = await indexingCache.get({
+                        table,
+                        key: values,
+                        db,
+                      });
 
-                      if (row === null) {
-                        return setCacheEntry(table, values, EntryType.INSERT);
+                      if (row) {
+                        return null;
```
typedarray, line 100: Don't quite follow this diff

packages/core/src/indexing-store/historical.ts
```diff
@@ -687,52 +334,27 @@ export const createHistoricalIndexingStore = ({
...
         ),
       ),
     // @ts-ignore
     sql: drizzle(
       async (_sql, params, method, typings) => {
-        await database.createTriggers();
-        await indexingStore.flush();
-        await database.removeTriggers();
-        isDatabaseEmpty = false;
+        await indexingCache.flush({ db });
```
typedarray, line 350: Do we need to call cache.commit here?

packages/core/src/database/index.ts
```diff
@@ -287,23 +308,12 @@ export const createDatabase = async ({
     await driver.internal.query(`CREATE SCHEMA IF NOT EXISTS "${namespace}"`);
 
     qb = {
-      internal: new Kysely({
+      migrate: new Kysely({
         dialect: new PostgresDialect({ pool: driver.internal }),
```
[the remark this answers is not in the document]
kyscott18: The listen connection for the client comes from this pool

packages/core/src/database/index.ts
```diff
@@ -1049,21 +1074,20 @@ export const createDatabase = async ({
...
             const app = await tx
-              .selectFrom("_ponder_meta")
-              .where("key", "=", "app")
-              .select("value")
-              .executeTakeFirstOrThrow()
-              .then((row) => row.value);
+              .select({ value: PONDER_META.value })
+              .from(PONDER_META)
+              .where(eq(PONDER_META.key, "app"))
+              .then((result) => result[0]!.value);
```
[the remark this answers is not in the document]
kyscott18: Note sure if you are talking about the non-null assertion or the entire query

packages/core/src/indexing-store/historical.ts
kyscott18: No, we only want to call cache.commit when we are sure that we will never have to retry that batch of events

packages/core/src/database/index.ts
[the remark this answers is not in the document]
typedarray: Non-null assertion. I guess before we were using `OrThrow`, so similar outcome. LGTM.

packages/core/src/database/index.ts
```diff
@@ -1072,30 +1096,27 @@ export const createDatabase = async ({
...
               // Remove indexes
 
               for (const indexStatement of schemaBuild.statements.indexes
                 .json) {
-                await tx.schema
-                  .dropIndex(indexStatement.data.name)
-                  .ifExists()
-                  .execute();
+                await tx.execute(
+                  sql.raw(`DROP INDEX IF EXISTS "${indexStatement.data.name}"`),
```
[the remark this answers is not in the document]
kyscott18: nice catch

packages/core/src/indexing-store/cache.test.ts
kyscott18: Not sure I follow, this indexing store almost directly makes it way into user code

packages/core/src/indexing-store/cache.ts
```diff
@@ -347,10 +347,13 @@ export const createIndexingCache = ({
         updateBuffer.get(table)!.delete(getCacheKey(table, key));
 
       if (
-        cache.get(table)!.has(getCacheKey(table, key)) &&
-        cache.get(table)!.get(getCacheKey(table, key))
+        (cache.get(table)!.has(getCacheKey(table, key)) &&
```
typedarray, line 350: Could we do `const cacheKey = getCacheKey(table, key)` once at the top of this function? `getCacheKey` seems quite expensive.

Also, isn't it redundant to do `Map.has(val) && Map.get(val)` to get a boolean?

kyscott18: The `get` may return `null`

PR #1514, opened 2025-02-13
fix: indexing store array parsing

### LINKED ISSUE #1484: .text().array() parsing issues with strings that have backslash literals
### Version

0.9.2

### Current behavior

Attempting to store a string with a backslash literal in a drizzle `.text().array()` column causes serialization/parsing issues. This occurs regardless of pglite or postgres backends, suggesting that the issue is within the drizzle codebase, in how they serialize `pgArray`s.

```ts
const STRING_ARRAY_VALUE = '//U_W_U\\\\'

await context.db.insert(schema.example).values({
  id: '0',
  stringArray: [STRING_ARRAY_VALUE]
})

console.log(await context.db.find(schema.example, {id: '0'}))
```

```ts
{ id: '0', stringArray: [ '//U_W_U' ], name: null }
```

and in the database the raw cell value is `{//U_W_U}`.

### Expected behavior

the string should return unchanged (`//U_W_U\\`) and the database value should represent the same

### Steps to reproduce

Check out the repo and run `pnpm dev`

### Link to repository

https://github.com/shrugs/ponder-string-array-test

### Anything else?

i'm working on tracking this down

kyscott18:
Closes #1484

### DIFF
--- a/.changeset/eleven-bees-smell.md
+++ b/.changeset/eleven-bees-smell.md
@@ -1,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Fixed a bug with parsing array column values.
--- a/packages/core/src/indexing-store/cache.ts
+++ b/packages/core/src/indexing-store/cache.ts
@@ -24,7 +24,7 @@ import {
   is,
   sql,
 } from "drizzle-orm";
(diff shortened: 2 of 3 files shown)


typedarray (approved):
LGTM cc @shrugs

PR #1517, opened 2025-02-13
fix: graphql hex value case sensitivity for arguments to singular fields

typedarray:


packages/core/src/graphql/index.ts
```diff
@@ -991,13 +992,37 @@ export function buildDataLoaderCache({
...
+                ),
+              )
+              // Find the database row corresponding to each normalized row fragment
+              .map((fragment) =>
+                rows.find((row) =>
+                  Object.entries(fragment).every(
+                    ([col, val]) => row[col] === val,
+                  ),
+                ),
+              )
```
kyscott18, line 1024: Are you worried that this is o(n^2)?

typedarray: Yea it's definitely slow, but it's kinda just how dataloader works. Another reason to dislike GraphQL (lol).


> To uphold the constraints of the batch function, it must return an Array of values the same length as the Array of keys, and re-order them to ensure each index aligns with the original keys [ 2, 9, 6, 1 ] ([Source](https://github.com/graphql/dataloader?tab=readme-ov-file#batch-function))


https://github.com/graphql/dataloader/blob/main/examples/SQL.md#using-dataloader-with-sqlite

kyscott18 (approved):

PR #1518, opened 2025-02-13
feat: improve indexing performance by delaying errors

kyscott18:


packages/core/src/indexing-store/cache.ts
```diff
@@ -465,6 +503,28 @@ export const createIndexingCache = ({
             { method: `${getTableName(table)}.flush()` },
             async () =>
               copy(table, text).catch((error) => {
+                // try to recover the event that caused the error
+                if (
+                  "where" in error &&
+                  (error.where as string).match(/line (\d+)/)
```
typedarray, line 509: I think it's possible for `error.where` to be null (or undefined), so we should probably handle that case and throw the original error. Same story for if `where` doesn't match the regex or the matched string cannot be coerced to a number

Maybe
```ts
const where = error.where
if (typeof where !== 'string') throw error 
```

packages/core/src/indexing-store/historical.ts
```diff
@@ -41,6 +41,8 @@ export const createHistoricalIndexingStore = ({
     worker: (fn) => fn(),
   });
 
+  let event: Event | undefined;
```
typedarray, line 44: Worth considering that this pattern of "setting" the current event could be used similarly for injecting the `db`/`tx` and `client` values.

packages/core/src/indexing-store/cache.ts
kyscott18: I think that is how it works if you read line 533

Issue #1535, opened 2025-02-17
Multiple db updates on the same row are overriding each other


KONFeature:
### Version

0.9.14

### Current behavior

During the historical sync (didn't test on realtime sync), I'm indexing three events,, the first one is creating an entry, and the following ones are both updating the same db row (one increasing a bigint value, and the second one decreasing it).

The second event (increasing a bigint value), is discarded once the third event is indexed. And so, I end up with a negative value in the db :/

Each events are emitted on different txs and on different blocks.

I wasn't facing any issue on the v0.8.32.

Here is some log output I added in the indexing function:
```
Post reward added pendingAmount {
  update: 3000000n,
  newValue: 3000000n,
  txHash: '0xb1eb5a4fffd97840006ee9fb61cf1922c57fde8755d309c26d168f2aae13bf2d'
}
6:52:59 PM DEBUG database   Inserted 1 'Reward' rows

Post reward added pendingAmount {
  update: 3000000n,
  newValue: 6000000n,
  txHash: '0x5da680fa3f594736744bd8a1699b36c99bdfd0f798f8c1a13431e6a01010285d'
}
6:52:59 PM DEBUG database   Updated 1 'Reward' rows

Post reward claimed pendingAmount {
  update: -6000000n,
  newValue: -3000000n,
  txHash: '0x3d36685532cdfee5ea37ccb7dcb86b52ce23f4dc2dd99199327b53a0e00b0b3a'
}
6:52:59 PM DEBUG database   Updated 1 'Reward' rows
```

### Expected behavior

The third event doesn't erase the cached second update, so the final state of the reward entity (the `newValue` in the log), is 0.

Currently we have:
0 + 3000000 = 6000000
3000000 + 3000000 = 6000000
6000000 - 6000000 = -3000000 

### Steps to reproduce

Db table:
```
export const rewardTable = onchainTable(
    "Reward",
    (t) => ({
        contractId: t.hex().notNull(),
        user: t.hex().notNull(),
        pendingAmount: t.bigint().notNull(),
        totalReceived: t.bigint().notNull(),
        totalClaimed: t.bigint().notNull(),
    }),
    (table) => ({
        pk: primaryKey({ columns: [table.contractId, table.user] }),
        userIdx: index().on(table.user),
        contractIdx: index().on(table.contractId),
        userContractIdx: index().on(table.user, table.contractId),
    })
);
```

Two event handler like this:
```
[...]

PR #1536, opened 2025-02-17
fix: Cleanup spillover once insert/update are flushed to the cache

KONFeature:
Resolve: https://github.com/ponder-sh/ponder/issues/1535

Not sure it's the best approach, but since the newly updated value are stored in the main cache, it shouldn't impact performance of both `has` and `get` method on the internal cache + it provide the latest data via the `get`.


kyscott18:
Fixed this issue with #1537

PR #1541, opened 2025-02-18
add new metrics

kyscott18:


packages/core/src/bin/commands/serve.ts
```diff
@@ -41,6 +42,19 @@ export async function serve({ cliOptions }: { cliOptions: CliOptions }) {
   const telemetry = createTelemetry({ options, logger, shutdown });
   const common = { options, logger, metrics, telemetry, shutdown };
 
+  const version = parse(cliOptions.version);
```
typedarray, line 45: Should we move this to `options` ?

packages/core/src/bin/ponder.ts
```diff
@@ -94,6 +95,7 @@ const startCommand = new Command("start")
     const cliOptions = {
       ...command.optsWithGlobals(),
       command: command.name(),
+      version: packageJson.version,
```
typedarray, line 98: I think this needs to get copied to a few more spots in this file. Or, does it get included automatically via `command.optsWithGlobals()` ?

packages/core/src/internal/metrics.ts
```diff
@@ -70,6 +78,30 @@ export class MetricsService {
...
+    this.ponder_settings_info = new prometheus.Gauge({
+      name: "ponder_settings_info",
+      help: "Ponder settings information",
+      labelNames: ["ordering"] as const,
+      registers: [this.registry],
+    });
+
+    this.ponder_historical_start_timestamp = new prometheus.Gauge({
+      name: "ponder_historical_start_timestamp",
+      help: "Timestamp at which the historical indexing started",
```
typedarray, line 96: Remove "the"

packages/core/src/bin/ponder.ts
kyscott18: nice catch

PR #1550, opened 2025-02-21
fix: truncate event name on dev UI

farrellh1:


farrellh1:
Hi @kyscott18 can you review it, thanks

kyscott18 (approved):
Nice work!

PR #1557, opened 2025-02-25
fix `flush()` when using raw sql

kyscott18:
Fixes a bug that occurs when calling `flush()` multiple times for the same batch of events (same transaction). This occurs because the "update" logic creates a temp table and isn't accounting for the case when the temp table already exists.

With this pr, the temp table is only created if not exists and truncated at the end of the function. Another solution would be to drop the temp table at the end of the function and recreate it on every call.

### DIFF
--- a/packages/core/src/indexing-store/cache.test.ts
+++ b/packages/core/src/indexing-store/cache.test.ts
@@ -110,7 +110,24 @@ test("flush() update", async (context) => {
 
     await indexingCache.flush({ client });
 
-    const result = await indexingStore.find(schema.account, {
+    let result = await indexingStore.find(schema.account, {
+      address: zeroAddress,
+    });
+
+    expect(result).toStrictEqual({
+      address: zeroAddress,
+      balance: 12n,
+    });
+
+    // flush again to make sure temp tables are cleaned up
+
+    await indexingStore.update(schema.account, { address: zeroAddress }).set({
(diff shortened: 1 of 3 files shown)


packages/core/src/indexing-store/cache.ts
```diff
@@ -646,6 +646,7 @@ export const createIndexingCache = ({
                 .map(({ sql }) => `target."${sql}" = source."${sql}"`)
                 .join(" AND ")};
             `;
+          const truncateQuery = `TRUNCATE TABLE "${getTableName(table)}" CASCADE`;
```
typedarray, line 649: I'd probably opt for dropping the temp table rather than create if not exists and truncate. That way the flush function doesn't leave around any state in the transaction. Not a strong opinion though.

packages/core/src/indexing-store/cache.test.ts
```diff
@@ -110,7 +110,24 @@ test("flush() update", async (context) => {
...
+    });
+
+    expect(result).toStrictEqual({
+      address: zeroAddress,
+      balance: 12n,
+    });
+
+    // flush again to make sure temp tables are cleaned up
+
+    await indexingStore.update(schema.account, { address: zeroAddress }).set({
```
typedarray, line 124: Nice failing test

PR #1567, opened 2025-02-27
fix: fix outdated link in CLI

dschlabach:
Hey all - ran into this link which caused a warning from the browser, updated it to the current docs link


typedarray:
Thanks!

PR #1579, opened 2025-03-04
add `bytes()` column type

### LINKED ISSUE #1487: add support for bytea column type
### Problem / use case

I'd like to store hex/bytes fields in `bytea` PG columns so I can do substring operations on it.

### Proposed solution

_No response_

kyscott18:
Closes #1487

### DIFF
--- a/docs/pages/docs/schema.mdx
+++ b/docs/pages/docs/schema.mdx
@@ -46,7 +46,7 @@ Ponder also includes a few extra column types built specifically for EVM indexin
 | :------------ | :------------------------------------------- | :-------------- | :-------------------- |
 | `bigint{:ts}` | Large integer (holds `uint256` and `int256`) | `bigint`        | `NUMERIC(78,0){:sql}` |
 | `hex{:ts}`    | UTF‐8 character sequence with `0x` prefix    | `0x${string}`   | `TEXT{:sql}`          |
-| `bytes{:ts}`  | Low-level byte array                         | `Uint8Array`    | `bytea{:sql}`         |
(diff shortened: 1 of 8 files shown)


frolic:
🙏

docs/pages/docs/schema.mdx
```diff
@@ -46,6 +46,7 @@ Ponder also includes a few extra column types built specifically for EVM indexin
 | :------------ | :------------------------------------------- | :-------------- | :-------------------- |
 | `bigint{:ts}` | Large integer (holds `uint256` and `int256`) | `bigint`        | `NUMERIC(78,0){:sql}` |
 | `hex{:ts}`    | UTF‐8 character sequence with `0x` prefix    | `0x${string}`   | `TEXT{:sql}`          |
+| `bytes{:ts}`  | Low-level byte array                         | `Uint8Array`    | `bytea{:sql}`         |
```
typedarray, line 49: We have a quick section below for our other custom column types (`bigint` and `hex`), would be nice to add one for `bytes`

Also I think the description could just be "Byte array"

packages/core/src/drizzle/bytes.ts
```diff
@@ -0,0 +1,68 @@
...
+    return new PgBytes<MakeColumnConfig<T, TTableName>>(
+      table,
+      this.config as ColumnBuilderRuntimeConfig<any, any>,
+    );
+  }
+
+  /**
+   * @deprecated Bytes columns cannot be used as arrays
+   */
+  override array(): never {
```
typedarray, line 47: Might be nice to test this

PR #1586, opened 2025-03-07
improve rpc request result performance

kyscott18:
## Description
Improved several performance characteristics surrounding cached rpc requests

- Dropped `request` column from `rpc_request_results` table
- Added index to support `pruneRpcRequestResults`
- Granular cache for `multicall` requests
- Stopped caching "0x" rpc responses, which are sometimes wrongly sent by degraded rpc


packages/core/src/sync-store/index.ts
```diff
@@ -725,55 +728,68 @@ export const createSyncStore = ({
...
-  getRpcRequestResult: async ({ request, chainId }) =>
+    );
+  },
+  getRpcRequestResults: async ({ requests, chainId }) =>
     database.wrap(
-      { method: "getRpcRequestResult", includeTraceLogs: true },
+      { method: "getRpcRequestResults", includeTraceLogs: true },
       async () => {
+        const requestHashes = requests.map((request) =>
+          crypto.createHash("md5").update(request).digest("hex"),
```
kyscott18, line 760: @typedarray Need to be very careful that this matches the previous hashes that we generated by postgres. I added a test for this.

packages/core/src/sync/transport.ts
```diff
@@ -45,7 +60,122 @@ export const cachedTransport = ({
...
+              requests: requests
+                .filter((_, index) => cachedResults[index] === undefined)
+                .map((request, index) => ({
+                  request,
+                  result: multicallResult[index]!,
+                }))
+                // Note: we don't cache request that failed or returned "0x". See more about "0x" below.
+                .filter(
+                  ({ result }) => result?.success && result.returnData !== "0x",
+                )
```
kyscott18, line 142: @typedarray I don't think we should cache failed requests and just re-request them

packages/core/src/sync-store/index.ts
typedarray: Seems good, but we should double check
- Make sure any pre-processing we're doing (casing, sorting keys?) is the same as before
- Casing, we're storing this in a TEXT column
- Any possibility that the strings were changed by the driver before making it to PG (in the old imp)

packages/core/src/sync/transport.ts
typedarray: Brilliant

packages/core/src/sync-store/index.ts
kyscott18: The `request` is preprocessed before making it to this function.

https://github.com/ponder-sh/ponder/blob/689156e6878d62da3cdd7cb646a590ed68726ea8/packages/core/src/sync/transport.ts#L84-L86

packages/core/src/indexing/index.test.ts
```diff
@@ -988,16 +988,22 @@ test.skip("ponderActions multicall()", async (context) => {
...
       {
         abi: erc20ABI,
         functionName: "totalSupply",
         address,
       },
+      {
+        abi: erc20ABI,
+        functionName: "totalSupply",
+        address,
+      },
```
typedarray, line 1006: Duplicate?

packages/core/src/sync-store/index.ts
typedarray: Any reason not to do the processing inside this function? Current pattern could make it easy for a caller to process incorrectly.

packages/core/src/sync-store/migrations.ts
```diff
@@ -1679,6 +1680,64 @@ GROUP BY fragment_id, chain_id
...
+        .execute();
+      await db.schema
+        .alterTable("rpc_request_results")
+        .renameColumn("request_hash_temp", "request_hash")
+        .execute();
+      await db.schema
+        .alterTable("rpc_request_results")
+        .addPrimaryKeyConstraint("rpc_request_results_pkey", [
+          "request_hash",
+          "chain_id",
```
typedarray, line 1709: My instinct would be to use the inverse order here.

packages/core/src/sync-store/migrations.ts
```diff
@@ -1679,6 +1680,64 @@ GROUP BY fragment_id, chain_id
...
+        ])
+        .execute();
+      await db.schema
+        .alterTable("rpc_request_results")
+        .dropColumn("request")
+        .execute();
+      await db
+        .updateTable("rpc_request_results")
+        .set({ block_number: 0 })
+        .where("block_number", "=", maxUint256)
```
typedarray, line 1719: What's the story here?

packages/core/src/sync-store/migrations.ts
```diff
@@ -1679,6 +1680,64 @@ GROUP BY fragment_id, chain_id
...
+      await db.schema
+        .alterTable("rpc_request_results")
+        .alterColumn("block_number", (col) => col.setDataType("bigint"))
+        .execute();
+      await db.schema
+        .alterTable("rpc_request_results")
+        .alterColumn("chain_id", (col) => col.setDataType("bigint"))
+        .execute();
+      await db.schema
+        .createIndex("rpc_request_results_chain_id_block_number_index")
```
typedarray, line 1730: This is used by the reorg prune query, yes?

packages/core/src/sync/transport.ts
```diff
@@ -45,7 +60,126 @@ export const cachedTransport = ({
...
+                    blockNumber === undefined
+                      ? undefined
+                      : blockNumber === "latest"
+                        ? 0
+                        : hexToNumber(blockNumber),
+                  result: JSON.stringify(result.returnData),
+                })),
+              chainId: chain!.id,
+            })
+            .catch(() => {});
```
typedarray, line 159: Seems like we should at least log a warning here, it's not expected that `insertRpcRequestResults` will ever throw, right?

packages/core/src/sync/transport.ts
```diff
@@ -95,10 +221,29 @@ export const cachedTransport = ({
...
+                          ? undefined
+                          : blockNumber === "latest"
+                            ? 0
+                            : hexToNumber(blockNumber),
+                      result: JSON.stringify(response),
+                    },
+                  ],
+                  chainId: chain!.id,
+                })
+                .catch(() => {});
```
typedarray, line 245: Same thought here re: logging

packages/core/src/sync/transport.ts
```diff
@@ -45,7 +60,126 @@ export const cachedTransport = ({
       async request({ method, params }) {
         const body = { method, params };
 
+        // multicall
         if (
+          method === "eth_call" &&
+          params[0]?.data?.startsWith(MULTICALL_SELECTOR)
```
typedarray, line 66: Do we need to handle multiple version of multicall or anything here? Should just make sure it matches viem behavior.

packages/core/src/sync/transport.ts
```diff
@@ -45,7 +60,126 @@ export const cachedTransport = ({
...
+            method: "eth_call",
+            params: [
+              {
+                data: call.callData,
+                to: call.target,
+              },
+              blockNumber,
+            ],
+          })).map((request) =>
+            toLowerCase(JSON.stringify(orderObject(request))),
```
typedarray, line 85: It's probably fine, but it's not super intuitive to me that we include the block tag in the hash. Were we doing that before?

packages/core/src/sync/transport.ts
kyscott18: viem only uses `aggregate3()`

https://github.com/wevm/viem/blob/ea5a0ced1210873df75920b1c9d8afcbc8c6359e/src/actions/public/multicall.ts#L211-L223

packages/core/src/indexing/index.test.ts
kyscott18: On purpose

packages/core/src/sync-store/migrations.ts
kyscott18: yes

packages/core/src/sync/transport.ts
kyscott18: The `database.wrap` inside `insertRpcRequestResults` will log an error

packages/core/src/sync/transport.ts
kyscott18: yes

packages/core/src/sync-store/migrations.ts
kyscott18: We need a way to signify "latest" as a block number in the database. Previously this value was `2**256-1`, but since we migrated the column type from numeric(78,0) to bigint, we need a new number.

PR #1587, opened 2025-03-07
docs: add description for "silent" log level

zkpepe:
I noticed that the "silent" log level didn't have a description in the table. To clarify its meaning, I added the following:

| `silent`         | No logs are produced                             |

This provides a complete understanding of the log levels and their effects.


typedarray:
This looks AI generated. It's clear enough as-is.

PR #1588, opened 2025-03-07
allow nested paths in contract `factory()` event parameter

### LINKED ISSUE #1323: feat: factory event `parameter` within struct
Some factory events include the target/child address within a struct. For example, from [Morpho](https://basescan.org/address/0xBBBBBbbBBb9cC5e90e3b3Af64bdAF62C37EEFFCb#code):

```solidity
struct MarketParams {
    address loanToken;
    address collateralToken;
    address oracle;
    address irm;
    uint256 lltv;
}

/// @notice Emitted when creating a market.
/// @param id The market id.
/// @param marketParams The market that was created.
event CreateMarket(Id indexed id, MarketParams marketParams);
```

At the moment, the `factory` configuration for this ABI item only accepts `"id"` or `"marketParams"`, neither of which will work at all. If possible, we should consider `parameter` more flexible, eg `"marketParams.oracle"` which was the target address in this example.

Perhaps we could support the entire Solidity syntax for how to index into a struct (by property name or array index) at both type-level and runtime.

anukul:
Closes #1323

### DIFF
--- a/packages/core/src/config/address.ts
+++ b/packages/core/src/config/address.ts
@@ -1,10 +1,5 @@
-import type {
-  AbiParameter,
-  SolidityArray,
-  SolidityArrayWithTuple,
-  SolidityTuple,
-} from "abitype";
-import type { AbiEvent } from "viem";
+import type { SolidityArrayWithTuple, SolidityTuple } from "abitype";
+import type { AbiEvent, AbiParameter } from "viem";
 
 type CommonFactoryParams<event extends AbiEvent = AbiEvent> = {
   /** Address of the factory contract that creates this contract. */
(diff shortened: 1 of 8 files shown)


packages/core/src/config/address.ts
```diff
@@ -6,7 +6,7 @@ export type Factory<event extends AbiEvent = AbiEvent> = {
   /** ABI event that announces the creation of a new instance of this contract. */
   event: event;
   /** Name of the factory event parameter that contains the new child contract address. */
-  parameter: Exclude<event["inputs"][number]["name"], undefined>;
+  parameterPath: string;
```
anukul, line 9: Can be updated to avoid a breaking change, where `parameter` is deprecated and later removed

PR #1589, opened 2025-03-09
docs: Fixed incorrect syntax in COPY command example

0xkazak:
Noticed a issue where the `COPY {:sql}` command had an extra `{:sql}` part, which is not valid in SQL syntax.
It should be just `COPY sql`.

This update removes the unnecessary part and ensures the example aligns with correct SQL syntax.


typedarray:
That's how syntax highlighting works in our docs framework. The website looks fine.

PR #1590, opened 2025-03-09
fix: Fix queryKey dependency tracking in React

imbarty:
Noticed that `queryOptions.queryKey` was being passed directly instead of as an array. Wrapped it in `[queryOptions.queryKey]` to ensure React tracks dependencies correctly. This prevents potential stale data issues.


typedarray:
The `QueryKey` value is already an array, so this should be working as expected.
Screen Shot 2025-03-09 at 6 55 45 PM

Did you observe a stale dependency issue?

PR #1596, opened 2025-03-11
indexing store prediction

kyscott18:
Improve the indexing store performance by recording historical cache accesses and using the data to prediction future accesses.

### DIFF
--- a/packages/core/src/indexing-store/cache.test.ts
+++ b/packages/core/src/indexing-store/cache.test.ts
@@ -326,7 +326,11 @@ test("load() loads predicted rows", async (context) => {
     event.event.args.to = BOB;
 
     await indexingCache.flush({ client });
-    await indexingCache.load({ events: [event], db: tx });
+    await indexingCache.prefetch({
+      events: [event],
+      db: tx,
+      eventCount: { "": 1 },
+    });
 
     const result = indexingCache.has({
       table: schema.account,
@@ -337,7 +341,7 @@ test("load() loads predicted rows", async (context) => {
   });
 });
 
-test("load() evicts rows", async (context) => {
+test("prefetch() evicts rows", async (context) => {
   const schema = {
     account: onchainTable("account", (p) => ({
       address: p.hex().primaryKey(),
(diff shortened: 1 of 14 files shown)


packages/core/src/indexing-store/cache.test.ts
```diff
@@ -300,7 +373,7 @@ test("commit() evicts rows", async (context) => {
     });
 
     await indexingCache.flush({ client });
-    indexingCache.commit();
+    await indexingCache.load({ events: [], db: tx });
```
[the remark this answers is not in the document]
kyscott18: It's testing that the call to "load()" evicts rows from the cache to free memory.

packages/core/src/indexing-store/cache.ts
```diff
@@ -626,50 +684,120 @@ export const createIndexingCache = ({
...
             }
           }
         }
+      }
 
-        common.logger.debug({
-          service: "indexing",
-          msg: `Evicted ${rowCount} rows from the cache`,
-        });
+      for (const [table, tableCache] of cache) {
```
typedarray, line 728: So, this block means - if one of our predictions is already in the cache, no need to preload it?

packages/core/src/indexing-store/cache.ts
```diff
@@ -626,50 +684,120 @@ export const createIndexingCache = ({
...
+            const endClock = startClock();
+
+            await db
+              .select()
+              .from(table)
+              .where(or(...conditions))
+              .then((results) => {
+                common.logger.debug({
+                  service: "indexing",
+                  msg: `Loaded ${results.length} '${getTableName(table)}' rows`,
```
typedarray, line 765: Preloaded ?

packages/core/src/indexing-store/profile.ts
```diff
@@ -0,0 +1,351 @@
...
+
+export const getProfilePatternKey = (pattern: ProfilePattern): string => {
+  return Object.values(pattern).join("_");
+};
+
+const eq = (
+  target: bigint | string | number | boolean | null | undefined,
+  value: bigint | string | number | boolean | null | undefined,
+) => {
+  if (!target) return false;
```
typedarray, line 14: Worth double checking this boolean logic, or perhaps commenting which type combos each of these cases handle

packages/core/src/indexing-store/cache.ts
kyscott18: yes

PR #1605, opened 2025-03-18
performance metrics

kyscott18:
Added `ponder_historical_concurrency_group_duration`, `ponder_historical_extract_duration`, `ponder_historical_transform_duration` metrics.

### DIFF
--- a/packages/core/src/sync-store/index.ts
+++ b/packages/core/src/sync-store/index.ts
@@ -747,8 +747,8 @@ export const createSyncStore = ({
           });
         }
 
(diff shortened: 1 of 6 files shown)


packages/core/src/sync-store/index.ts
```diff
@@ -745,6 +747,11 @@ export const createSyncStore = ({
           });
         }
 
+        common.metrics.ponder_historical_extract_duration.inc(
+          { step: "decode" },
```
typedarray, line 751: format

typedarray (approved):

PR #1607, opened 2025-03-19
Fix null bytes behavior

### LINKED ISSUE #1456: Ponder Implicitly Removes Null Bytes from Event Arguments
### Version

0.8.28

### Current behavior

In indexing ENS, some labels with null bytes are emitted, but these labels cannot be stored in postgres (which is likely the reason ponder trims them to begin with). Unfortunately this removes information from the application-layer who might handle these cases differently. In ENS's case, it's important that `'vitalik\x00'.eth` and `vitalik.eth` are differentiable. 

In this specific case, we can use `const hadNullBytes = labelhash(name) !== label` as a proxy for null-byte detection, since the labelhash will be different for the true argument value that includes the null byte.

More information is available here, where I debugged this behavior 👇

https://github.com/namehash/ensnode/issues/36

### Expected behavior

I believe the current behavior of ponder is extremely reasonable, but I worry that the developer integrating ponder loses this information and may accidentally index the wrong data. In our case we were treating these two names as identical, which messed up the resulting data (and would have perhaps enabled someone resolving `vitalik.eth` to receive the imposter's information. We assumed we were getting the raw argument data from ponder, intending to handle the null-byte ourselves, but were actually receiving strings with null bytes already removed.

It seems that the default behavior of postgres is to throw an error if a null byte is inserted, though I'm not sure why this didn't cause subgraph indexers from failing to index these names in the original postmortem.

https://ens.mirror.xyz/9GN77d-MqGvRypm72FcwgxlUnPSuKWhG3rWxddHhRwM

In any case, I think that the default behavior, where postgres throws an unrecoverable error, forcing the user to handle this case, is likely the correct path. Ponder can/should provide null-byte helper functions to make this easier on users, for example a `containsNullCharacters` or `removeNullCharacters` helper fn.

### Steps to reproduce

You don't really need to reproduce this, but here's an offending tx: `0x535bf909ec258054cb03ab03498b4de0b365e1037a43fcdfbdcb28202382a9cd` which you can access via ponder with the following config:

```ts
EthRegistrarControllerOld: {
      network: "mainnet",
      abi: EthRegistrarControllerOld,
      address: "0x283Af0B28c62C092C9727F1Ee09c02CA627EB7F5",
      startBlock: 16369364,
      endBlock: 16369364 + 1,
    },
```

and you can log the arguments provided to your event handler to see that the null byte is removed.

the relevant line in ponder is https://github.com/ponder-sh/ponder/blob/178bcbfd36b5eb418ee118c0e6b02a46c9c7ed6d/packages/core/src/sync/events.ts#L505C25-L505C45

### Link to repository

https://github.com/namehash/ensnode

### Anything else?

_No response_

kyscott18:
Closes #1456

### DIFF
--- a/packages/core/src/drizzle/text.ts
+++ b/packages/core/src/drizzle/text.ts
@@ -62,7 +62,7 @@ export class PgText<
     if (value.match(/\0/g)) {
       globalThis.PONDER_COMMON?.logger.warn({
         service: "indexing",
-        msg: `${getTableName(this.table)}.${this.name} contains an invalid null byte character in ${value}. The null bytes will be removed to prevent a crash.`,
+        msg: `Detected and removed null byte characters from the string '${value}' inserted into the ${getTableName(this.table)}.${this.name} column. Postgres "text" columns do not support null byte characters. Please consider handling this case in your indexing logic.`,
       });
 
       return value.replace(/\0/g, "");
--- a/packages/core/src/drizzle/text.ts
+++ b/packages/core/src/drizzle/text.ts
@@ -1,0 +1,72 @@
+import {
+  type ColumnBaseConfig,
+  type ColumnBuilderBaseConfig,
+  type ColumnBuilderRuntimeConfig,
+  type MakeColumnConfig,
+  entityKind,
+  getTableName,
+} from "drizzle-orm";
+import {
+  type AnyPgTable,
+  PgColumn,
+  PgColumnBuilder,
+  type PgTextConfig,
+} from "drizzle-orm/pg-core";
+
+export type PgTextBuilderInitial<
+  TName extends string,
+  TEnum extends [string, ...string[]],
+> = PgTextBuilder<{
+  name: TName;
+  dataType: "string";
+  columnType: "PgText";
+  data: TEnum[number];
+  enumValues: TEnum;
+  driverParam: string;
+}>;
+
+export class PgTextBuilder<
+  T extends ColumnBuilderBaseConfig<"string", "PgText">,
+> extends PgColumnBuilder<T, { enumValues: T["enumValues"] }> {
+  static override readonly [entityKind]: string = "PgTextBuilder";
+
+  constructor(name: T["name"], config: PgTextConfig<T["enumValues"]>) {
+    super(name, "string", "PgText");
+    this.config.enumValues = config.enum;
+  }
+
+  /** @internal */
+  // @ts-ignore
+  override build<TTableName extends string>(
+    table: AnyPgTable<{ name: TTableName }>,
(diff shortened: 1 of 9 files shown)


packages/core/src/drizzle/text.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  override mapToDriverValue(value: string) {
+    if (value.match(/\0/g)) {
+      globalThis.PONDER_COMMON?.logger.warn({
+        service: "indexing",
+        msg: `${getTableName(this.table)}.${this.name} contains an invalid null byte character in ${value}. The null bytes will be removed to prevent a crash.`,
+      });
+
+      return value.replace(/\0/g, "");
```
typedarray, line 68: replaceAll ?

packages/core/src/drizzle/text.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  getSQLType(): string {
+    return "text";
+  }
+
+  override mapToDriverValue(value: string) {
+    if (value.match(/\0/g)) {
+      globalThis.PONDER_COMMON?.logger.warn({
+        service: "indexing",
+        msg: `${getTableName(this.table)}.${this.name} contains an invalid null byte character in ${value}. The null bytes will be removed to prevent a crash.`,
```
typedarray, line 65: . The Postgres TEXT column type does not support null byte characters. To avoid a crash, the characters were stripped before 


Detected and removed null byte characters from the string '${value}' inserted into the ${getTableName(this.table)}.${this.name} column. Postgres "text" columns do not support null byte characters. Please consider handling this case in your indexing logic.

packages/core/src/sync-store/encoding.ts
```diff
@@ -372,10 +372,8 @@ export const encodeTrace = ({
   type: trace.trace.type,
   gas: hexToBigInt(trace.trace.gas),
   gas_used: hexToBigInt(trace.trace.gasUsed),
-  error: trace.trace.error ? trace.trace.error.replace(/\0/g, "") : null,
-  revert_reason: trace.trace.revertReason
-    ? trace.trace.revertReason.replace(/\0/g, "")
-    : null,
+  error: trace.trace.error ?? null,
+  revert_reason: trace.trace.revertReason ?? null,
```
typedarray, line 376: Hmm, I think we still want these? These aren't passing through the Drizzle text column, right?

packages/core/src/drizzle/text.ts
kyscott18: I think the `g` in the regex makes it so that it replaces all occurrences.

typedarray (approved):
Needs a changeset but this LGTM!

PR #1608, opened 2025-03-19
add `ponder db prune` command

kyscott18:
Also fixed a bug for `ponder db list` when there was only one schema in the database

### DIFF
--- a/docs/pages/docs/api-reference/ponder-cli.mdx
+++ b/docs/pages/docs/api-reference/ponder-cli.mdx
@@ -106,6 +106,19 @@ List all deployments
 │ demo          │       no │      1 day ago │           5 │
 ```
 
+## db prune
+
+Prune all inactive deployments.
+
+```bash
+Usage: ponder db prune [options]
+
+Prune all inactive deployments
+
+Options:
+  -h, --help  display help for command
+```
+
 ## codegen
 
 <Callout type="info">
--- a/docs/pages/docs/api-reference/ponder-cli.mdx
+++ b/docs/pages/docs/api-reference/ponder-cli.mdx
@@ -92,12 +92,12 @@ Options:
 
 ## db list
 
-List all deployments.
+List all Ponder deployments.
 
 ```bash
 Usage: ponder db list
 
-List all deployments
+List all Ponder deployments
 
 │ Schema        │ Active   │ Last active    │ Table count │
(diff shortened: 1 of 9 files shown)


docs/pages/docs/api-reference/ponder-cli.mdx
```diff
@@ -106,6 +106,19 @@ List all deployments
 │ demo          │       no │      1 day ago │           5 │
 ```
 
+## db prune
+
+Prune all inactive deployments.
```
typedarray, line 111: We could probably be more specific here and in the help text, maybe

Drop all database schemas created by Ponder instances that are no longer active.

And a similar description for the `db list` command.

packages/core/src/bin/commands/list.ts
```diff
@@ -60,42 +64,40 @@ export async function list({ cliOptions }: { cliOptions: CliOptions }) {
...
-      .from(getPonderMeta(row.tableSchema))
-      .where(eq(getPonderMeta(row.tableSchema).key, "app")),
+      .from(getPonderMeta(row.schema))
+      .where(eq(getPonderMeta(row.schema).key, "app")),
   );
 
-  // @ts-ignore
-  const result = await unionAll(...queries);
+  if (queries.length === 0) {
+    console.log("No 'ponder start' apps found in this database.\n");
```
typedarray, line 88: Any reason not to use the logger here?

packages/core/src/bin/commands/prune.ts
```diff
@@ -0,0 +1,167 @@
...
+    tablesToDrop.push(`"${schema}"."_ponder_meta"`);
+    tablesToDrop.push(`"${schema}"."_ponder_status"`);
+
+    const tableCount = ponderSchemas.find(
+      (s) => s.schema === schema,
+    )!.tableCount;
+
+    if (schema !== "public" && tableCount <= 2 + value.table_names.length * 2) {
+      schemasToDrop.push(schema);
+    }
```
typedarray, line 137: What's the logic here? Are you hesitant to simply do DROP SCHEMA CASCADE here?

kyscott18: The schema is only dropped if we can be pretty sure there are no other tables besides Ponder tables. 2 + x * 2 is the formula used to account for reorg tables, _ponder_meta and _ponder_status

typedarray (approved):

PR #1623, opened 2025-03-25
fix `timestamp` column

kyscott18:


packages/core/src/indexing-store/cache.ts
```diff
@@ -402,70 +402,72 @@ export const createIndexingCache = ({
...
-                  // @ts-ignore
-                  await client.query("SAVEPOINT flush");
-                },
-              );
 
-              if (result.status === "success") {
-                return;
-              }
+          try {
+            const text = getCopyText(
```
kyscott18, line 411: @typedarray moving `getCopyText` inside of the try catch improved the log when the error was in that function.

typedarray: Nice, makes sense.

packages/core/src/indexing-store/cache.ts
```diff
@@ -526,67 +528,79 @@ export const createIndexingCache = ({
...
-          await new Promise(setImmediate);
-
           const endClock = startClock();
 
+          // @ts-ignore
+          await client.query(createTempTableQuery);
+          // @ts-ignore
+          await client.query("SAVEPOINT flush");
+
           try {
```
typedarray, line 538: Does it make sense for these to be outside the try ? Just notice it's different from before.

packages/core/src/indexing-store/historical.test.ts
```diff
@@ -1035,3 +1035,249 @@ test("text with null bytes", async (context) => {
...
+    await indexingStore.insert(schema.account).values({
+      address: zeroAddress,
+      timestamp: new Date(1742925862000),
+    });
+
+    const result = await indexingStore.find(schema.account, {
+      address: zeroAddress,
+    });
+
+    expect(result).toStrictEqual({
```
typedarray, line 1114: Nice failing test

typedarray (approved):

packages/core/src/indexing-store/cache.ts
kyscott18: yes

PR #1628, opened 2025-03-28
rpc request prediction for `context.client`

kyscott18:
Improve performance of `context.client.readContract()` and `context.client.multicall()` using profiling and prefetching.

### DIFF
--- a/packages/core/src/indexing/index.test.ts
+++ b/packages/core/src/indexing/index.test.ts
@@ -50,11 +50,14 @@ test("createIndexing()", async (context) => {
     schemaBuild: { schema },
   });
 
+  const eventCount = {};
+
   const cachedViemClient = createCachedViemClient({
     common,
     indexingBuild: { networks },
     requestQueues: [createRequestQueue({ network: networks[0]!, common })],
     syncStore,
+    eventCount,
   });
 
   const indexing = createIndexing({
@@ -65,6 +68,7 @@ test("createIndexing()", async (context) => {
       indexingFunctions: {},
     },
     client: cachedViemClient,
+    eventCount,
   });
 
   expect(indexing).toBeDefined();
@@ -76,11 +80,14 @@ test("processSetupEvents() empty", async (context) => {
     schemaBuild: { schema },
   });
 
+  const eventCount = {};
+
   const cachedViemClient = createCachedViemClient({
     common,
     indexingBuild: { networks },
     requestQueues: [createRequestQueue({ network: networks[0]!, common })],
     syncStore,
+    eventCount,
   });
 
   const indexing = createIndexing({
@@ -91,6 +98,7 @@ test("processSetupEvents() empty", async (context) => {
       indexingFunctions: {},
     },
     client: cachedViemClient,
+    eventCount,
   });
 
   const result = await indexing.processSetupEvents({ db: indexingStore });
@@ -108,11 +116,14 @@ test("processSetupEvents()", async (context) => {
     "Erc20:setup": vi.fn(),
   };
 
+  const eventCount = { "Erc20:setup": 0 };
+
   const cachedViemClient = createCachedViemClient({
     common,
     indexingBuild: { networks },
     requestQueues: [createRequestQueue({ network: networks[0]!, common })],
     syncStore,
+    eventCount,
   });
 
   const indexing = createIndexing({
@@ -123,6 +134,7 @@ test("processSetupEvents()", async (context) => {
       indexingFunctions,
     },
     client: cachedViemClient,
+    eventCount,
   });
 
   const result = await indexing.processSetupEvents({ db: indexingStore });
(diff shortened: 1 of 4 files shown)


packages/core/src/indexing/client.ts
```diff
@@ -0,0 +1,998 @@
...
+      // Use profiling metadata + next event batch to determine which
+      // rpc requests are going to be made, and preload them into the cache.
+
+      const prediction: { ev: number; request: Request }[] = [];
+
+      for (const event of events) {
+        if (profile.has(event.name)) {
+          for (const [, { pattern, count }] of profile.get(event.name)!) {
+            // Expected value of times the prediction will be used.
+            const ev = count / eventCount[event.name]!;
```
typedarray, line 584: Are these are both total counts since the beginning of indexing?

packages/core/src/indexing/client.ts
```diff
@@ -0,0 +1,998 @@
...
+      for (const [chainId, requests] of chainRequests.entries()) {
+        const ni = indexingBuild.networks.findIndex(
+          (n) => n.chainId === chainId,
+        );
+        const network = indexingBuild.networks[ni]!;
+        const requestQueue = requestQueues[ni]!;
+
+        const dbRequests = requests.filter(({ ev }) => ev > 0.2);
+
+        const cachedResults = await syncStore.getRpcRequestResults({
```
typedarray, line 619: Maybe worth running concurrently across chains? Or even batching into one db query across all chains?

packages/core/src/indexing/client.ts
```diff
@@ -0,0 +1,998 @@
...
+              returnData: `0x${string}`;
+            }
+          >();
+          const requestsToInsert = new Set<EIP1193Parameters>();
+
+          for (const request of requests) {
+            const cacheKey = getCacheKey(request);
+
+            if (cache.get(network.chainId)!.has(cacheKey)) {
+              const cachedResult = cache.get(network.chainId)!.get(cacheKey)!;
```
typedarray, line 725: Is it possible for something to be in the cache that was _not_ a prediction for the current batch of events?

packages/core/src/indexing/index.test.ts
```diff
@@ -59,20 +51,21 @@ test("createIndexing()", async (context) => {
     schemaBuild: { schema },
   });
 
+  const cachedViemClient = createCachedViemClient({
+    common,
+    indexingBuild: { networks },
+    requestQueues: [createRequestQueue({ network: networks[0]!, common })],
```
typedarray, line 57: Is it an invariant that the indexes of the networks and requestQueues arrays are the same?

packages/core/src/indexing/index.ts
```diff
@@ -179,10 +157,9 @@ export const createIndexing = ({
     const metricLabel = { event: event.name };
 
     try {
-      blockNumber = event.block;
+      // blockNumber = event.block;
```
typedarray, line 160: rm

packages/core/src/indexing/index.ts
```diff
@@ -317,6 +296,10 @@ export const createIndexing = ({
       context.db = db;
       for (let i = 0; i < events.length; i++) {
         const event = events[i]!;
+
+        client.event = event;
+        context.client = client.getClient(networkByChainId[event.chainId]!);
```
typedarray, line 301: I think we're indeed re-running a lot of client initialization code here - might get optimized but worth fixing IMO

typedarray (commented):
Looking great - couple questions

packages/core/src/indexing/index.test.ts
kyscott18: yes

packages/core/src/indexing/client.ts
```diff
@@ -0,0 +1,998 @@
...
+
+    return actionsWithMetrics;
+  };
+
+  return {
+    getClient(network) {
+      const requestQueue =
+        requestQueues[indexingBuild.networks.findIndex((n) => n === network)]!;
+
+      return createClient({
```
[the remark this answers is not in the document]
kyscott18: I think this should be fixed upstream by only calling this function once per chain

packages/core/src/indexing/client.ts
kyscott18: yes

packages/core/src/indexing/client.ts
kyscott18: no

packages/core/src/indexing/client.ts
kyscott18: Gonna just do await Promise.all for now

packages/core/src/indexing/profile.ts
```diff
@@ -0,0 +1,675 @@
...
+export const getProfilePatternKey = (pattern: ProfilePattern): string => {
+  return JSON.stringify(orderObject(pattern), (key, value) => {
+    if (key === "abi") return undefined;
+
+    if (typeof value === "bigint") {
+      return value.toString();
+    }
+    return value;
+  });
+};
```
kyscott18, line 21: @typedarray Can you take a look at this? 

Is this a reasonable way to get a unique key from the pattern?

Issue #1566, opened 2025-02-26
feat: factory start/endBlock


khaidarkairbek:
### Problem / use case

Currently, the factory address tracks child addresses in the interval between start and endBlock of the contract/account config. 

However, in some niche use cases there might be a need to track factory address only for a small subinterval of the contract indexing timeline. The other use case might be when the factory needs to track child addresses before the contract indexing timeline. Ex: collect uniswap pools via factory but only interested in contract activity from a week ago. 

In the current implementation, you would either need to run contract indexing for an unnecessarily big interval or collect more child addresses than needed. 

@kyscott18 @typedarray let me know what you think. 

### Proposed solution

I would propose a solution building on current factory implementation with optional startBlock and endBlock fields: 
```ts
type Factory<event extends AbiEvent = AbiEvent> = {
  /** Address of the factory contract that creates this contract. */
  address: `0x${string}` | readonly `0x${string}`[];
  /** ABI event that announces the creation of a new instance of this contract. */
  event: event;
  /** Name of the factory event parameter that contains the new child contract address. */
  parameter: Exclude<event["inputs"][number]["name"], undefined>;
  /** Optional factory interval */
  startBlock?: number; 
  endBlock?: number; 
};
```

Due to the nature of the factory addresses, the following validations would make sense: 
factory startBlock <= contract/account startBlock: 
    - either error out during validation or default contract startBlock to factory startBlock
factory endBlock <= contarct/account endBlock:
    - either error out during validation of default factory endBlock to contract endBlock

harry-nutsfi:
Exactly the problem that I've raised in the telegram channel!

PR #1629, opened 2025-03-30
feat: factory block intervals

khaidarkairbek:
Relates to #1566.

adds fromBlock and toBlock to factory fields to specify factory specific block range in addition to filter specific block range.

### DIFF
--- a/packages/core/src/build/configAndIndexingFunctions.ts
+++ b/packages/core/src/build/configAndIndexingFunctions.ts
@@ -280,6 +280,48 @@ export async function buildConfigAndIndexingFunctions({
         `Validation failed: Start block for '${source.name}' is after end block (${startBlock} > ${endBlock}).`,
       );
     }
+
+    if (
+      "address" in source &&
+      typeof source.address === "object" &&
+      !Array.isArray(source.address)
+    ) {
+      const factoryStartBlock =
+        (await resolveBlockNumber(source.address.startBlock, network)) ??
+        startBlock;
+
+      const factoryEndBlock =
+        (await resolveBlockNumber(source.address.startBlock, network)) ??
+        endBlock;
+
+      if (
+        factoryStartBlock !== undefined &&
+        (startBlock === undefined || factoryStartBlock > startBlock)
+      ) {
+        throw new Error(
+          `Validation failed: Start block for '${source.name}' is before start block of factory address (${factoryStartBlock} > ${startBlock}).`,
+        );
+      }
+
+      if (
+        endBlock !== undefined &&
+        (factoryEndBlock === undefined || factoryEndBlock > endBlock)
+      ) {
+        throw new Error(
+          `Validation failed: End block for ${source.name}  is before end block of factory address (${factoryEndBlock} > ${endBlock}).`,
+        );
+      }
+
+      if (
+        factoryStartBlock !== undefined &&
+        factoryEndBlock !== undefined &&
+        factoryEndBlock < factoryStartBlock
+      ) {
+        throw new Error(
+          `Validation failed: Start block for '${source.name}' factory address is after end block (${factoryStartBlock} > ${factoryEndBlock}).`,
+        );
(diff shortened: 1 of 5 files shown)


packages/core/src/sync-historical/index.ts
```diff
@@ -459,8 +459,31 @@ export const createHistoricalSync = async (
   const syncAddressFactory = async (
     factory: Factory,
     interval: Interval,
+    filterFromBlock: number | undefined,
   ): Promise<Map<Address, number>> => {
-    await syncLogFactory(factory, interval);
+    /**
+     * We sync the factory if:
+     *  - factory is in the interval (factory.toblock >= interval[0] || factory.toBlock === undefined)
+     *  - this is the first interval for the filter (filterFromBlock === interval[0]) -> sync factory even before the filter block range
```
kyscott18, line 467: Can you explain this second case?

kyscott18:
@khaidarkairbek I imagine we're gonna have to change some stuff regarding which intervals are being synced as well
https://github.com/ponder-sh/ponder/blob/c3d0679eb0ffd3cadb0c3136a1b735ca8d928d11/packages/core/src/sync-historical/index.ts#L772-L792

packages/core/src/sync-historical/index.ts
khaidarkairbek: yeah, this is non ideal workaround, since we can have factory start being defined before the source startBlock, we need to sync all the child addresses from the start of the factory to the start of the filter or end of the factory, whichever is the least. (i am currently thinking of other options.)

khaidarkairbek:
> @khaidarkairbek I imagine we're gonna have to change some stuff regarding which intervals are being synced as well
> 
> https://github.com/ponder-sh/ponder/blob/c3d0679eb0ffd3cadb0c3136a1b735ca8d928d11/packages/core/src/sync-historical/index.ts#L772-L792

In this current implementation, it is not necessary, since the factory is caught up at the start of the filter interval when indexing historically. I think this is a bit ugly, and we need to resolve intervals instead. Also, the other scenario we need to cover is if we have a factory that needs syncing historically, but source starts in realtime.

packages/core/src/sync-historical/index.ts
```diff
@@ -474,6 +483,9 @@ export const createHistoricalSync = async (
     let logs: SyncLog[];
     if (isAddressFactory(filter.address)) {
       const childAddresses = await syncAddressFactory(filter.address, interval);
+
+      if ((filter.fromBlock ?? 0) > interval[1]) return;
```
kyscott18, line 487: Include a comment like 
```ts
// Note: Exit early when only the factory needs to be synced
```

kyscott18 (approved):

packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -323,6 +323,48 @@ export async function buildConfigAndIndexingFunctions({
         `Validation failed: Start block for '${source.name}' is after end block (${startBlock} > ${endBlock}).`,
       );
     }
+
+    if (
```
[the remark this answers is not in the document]
khaidarkairbek: correct, these are the only validations needed as far as i know

packages/core/src/sync-historical/index.ts
```diff
@@ -598,6 +611,9 @@ export const createHistoricalSync = async (
       ? await syncAddressFactory(filter.toAddress, interval)
       : undefined;
 
+    // Note: Exit early when only the factory needs to be synced
+    if ((filter.fromBlock ?? 0) > interval[1]) return;
```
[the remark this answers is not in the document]
khaidarkairbek: I don't think so, the only case when we would enter the sync function is either for factory sync, filter sync or both. Factory syncs above and whether we need to sync the filter is based on whether the interval overlaps with filter, which this check is sufficient for.

PR #1634, opened 2025-04-01
improve factory performance in historical sync

kyscott18:


typedarray (approved):
Needs a changeset but looks good! Should be a nice speedup and db size reduction for large factory apps.

PR #1640, opened 2025-04-03
add retry for ContractFunctionZeroDataError

kyscott18:


packages/core/src/indexing/ponderActions.ts
```diff
@@ -195,13 +195,38 @@ export const getPonderActions = (getBlockNumber: () => bigint) => {
...
+    const getRetryAction = (action: PonderActions[keyof PonderActions]) => {
+      return async (...args: Parameters<typeof action>) => {
+        const RETRY_COUNT = 3;
+        for (let i = 0; i <= RETRY_COUNT; i++) {
+          try {
+            // @ts-ignore
+            return await action(...args);
+          } catch (error) {
+            if (
+              (error as Error)?.message.includes("returned no data") ===
```
kyscott18, line 207: @typedarray Should we try to use `instanceof` instead?

PR #1644, opened 2025-04-07
docs: replace event.log.id with event.id for global uniqueness

barajeel:
In version 0.10, the property `event.log.id` was removed (along with `event.trace.id`). I've updated the code to use the new `event.id`, which serves as the globally unique identifier for events.

This ensures compatibility with the latest changes and maintains event tracking integrity.


typedarray:
The line that you updated is the "Before" section of a before & after comparison for the migration. So, it's written as expected.

PR #1645, opened 2025-04-07
docs: fix variable name in Context example

barajeel:
corrected a mismatch where `context` was referred to as `event` in the comment.


barajeel:
[typedarray](https://github.com/typedarray) pls check it

typedarray:
The fix is correct, so I'm accepting this - but FYI your contribution smells like AI-assisted reputation farming (lol).

PR #1646, opened 2025-04-07
new `ponder_sync` migrations

kyscott18:
This pr introduces a new migration system for the `ponder_sync` database schema.

### Description
The main difference is that each minor version will have its own database schema. For example, all Ponder apps using v0.11 will use the `ponder_sync_0_11` schema. Previous database schemas will not be deleted.

This system allows for backwards compatibility between minor Ponder versions. It also allows for running two deployments of Ponder against the same database with different minor versions.

Migration files in `sync-store/sql` define migrations. Migrations are generally split into two parts, one part for copying data from the previous schema, and another part for performing data migrations against the copied data.

### Locking
It's important to prevent a deadlock when performing a migration while another deployment is running, or failing gracefully when two deployments are attempting to perform a migration at the same time.

An important detail is that a migration does not write to any rows in the existing schema, it only reads from the existing schema and writes to the new one.

Because of our conservative use of Postgres statements, a running deployment will never conflict with a migrating deployment. More info https://www.postgresql.org/docs/current/explicit-locking.html.

When performing a migration, Ponder uses a repeatable read transaction. This transaction isolation level uses a snapshot of the data taken at the beginning of the transaction, essentially guaranteeing the copied data will be consistent.
https://github.com/ponder-sh/ponder/blob/48673dba3478f589a1fb4d7da8546f1bcb8d80ef/packages/core/src/database/index.ts#L687-L695

### Legacy migrations
Currently this pr requires users to first migrate to v0.10 (or use an empty database) and throws an error otherwise.
https://github.com/ponder-sh/ponder/blob/48673dba3478f589a1fb4d7da8546f1bcb8d80ef/packages/core/src/database/index.ts#L657-L671


packages/core/src/sync-store/encode.ts
```diff
@@ -0,0 +1,152 @@
...
+import { hexToBigInt, hexToNumber } from "viem";
+import type * as ponderSyncSchema from "./schema.js";
+
+export const encodeBlock = ({
+  block,
+  chainId,
+}: {
+  block: SyncBlock;
+  chainId: number;
+}): typeof ponderSyncSchema.blocks.$inferInsert => ({
```
typedarray, line 19: Better to use satisfies?

packages/core/src/sync-store/index.ts
```diff
@@ -536,136 +548,166 @@ export const createSyncStore = ({
...
+            address: ponderSyncSchema.logs.address,
+            topic0: ponderSyncSchema.logs.topic0,
+            topic1: ponderSyncSchema.logs.topic1,
+            topic2: ponderSyncSchema.logs.topic2,
+            topic3: ponderSyncSchema.logs.topic3,
+            data: ponderSyncSchema.logs.data,
+          })
+          .from(ponderSyncSchema.logs)
+          .where(
+            and(
```
typedarray, line 670: Noting that we removed the `logFilter` where clauses here, not sure if intentional

packages/core/drizzle.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from "drizzle-kit";
+
+export default defineConfig({
```
typedarray, line 3: Worth making sure this file doesn't get published

packages/core/src/index.ts
```diff
@@ -24,6 +24,8 @@ export type NetworkConfig = Prettify<Config["networks"][string]>;
 export type BlockConfig = Prettify<Config["blocks"][string]>;
 export type DatabaseConfig = Prettify<Config["database"]>;
 
+export * as PONDER_SYNC_SCHEMA from "./sync-store/schema.js";
```
typedarray, line 27: I assume this is WIP?

packages/core/src/sync-store/index.ts
```diff
@@ -259,69 +252,82 @@ export const createSyncStore = ({
     await database.wrap(
       { method: "insertChildAddresses", includeTraceLogs: true },
       async () => {
-        const values: InsertObject<PonderSyncSchema, "factory_addresses">[] =
+        const values: (typeof ponderSyncSchema.factoryAddresses.$inferInsert)[] =
```
typedarray, line 255: I think we may have a merge conflict here with the batching we added on main

packages/core/src/sync-store/index.ts
```diff
@@ -536,136 +548,166 @@ export const createSyncStore = ({
...
-            sql<string>`state_root`.as("stateRoot"),
-            sql<string>`receipts_root`.as("receiptsRoot"),
-            sql<string>`transactions_root`.as("transactionsRoot"),
-            sql<string>`sha3_uncles`.as("sha3Uncles"),
-            "size",
-            "difficulty",
-            sql<string>`total_difficulty`.as("totalDifficulty"),
-            sql<string>`extra_data`.as("extraData"),
-          ])
+          .select({
```
typedarray, line 551: selectAll or perhaps spread operator?

packages/core/src/sync-store/index.ts
typedarray: Unless there's a strong reason to remove these, I think we should add them back.

PR #1649, opened 2025-04-08
improve user schema migrations

kyscott18:
Migrations of the "user" schema now validate the "version" of the `_ponder_meta` table.

https://github.com/ponder-sh/ponder/blob/5a3a2ddc80c550f2c2b6d630ddaf340718d40ccb/packages/core/src/database/index.ts#L104

New instances of Ponder will validate using this version or throw an error. Note that versions <=0.8 will throw the error because the `version` column was added in 0.9.

Also combined the `database.migrate()` and `database.recoverCheckpoint()` function.

### DIFF
--- a/packages/core/src/database/index.ts
+++ b/packages/core/src/database/index.ts
@@ -871,7 +871,7 @@ EXECUTE PROCEDURE "${namespace}".${notification};`),
                   `DROP INDEX IF EXISTS "${namespace}"."${indexStatement.data.name}"`,
                 ),
               );
-              common.logger.info({
+              common.logger.debug({
                 service: "database",
                 msg: `Dropped index '${indexStatement.data.name}' in schema '${namespace}'`,
               });
--- a/packages/core/src/database/index.ts
+++ b/packages/core/src/database/index.ts
@@ -1,6 +1,10 @@
 import { randomUUID } from "node:crypto";
 import { getPrimaryKeyColumns, getTableNames } from "@/drizzle/index.js";
-import { getColumnCasing, getReorgTable } from "@/drizzle/kit/index.js";
+import {
+  getColumnCasing,
+  getReorgTable,
+  sqlToReorgTableName,
+} from "@/drizzle/kit/index.js";
 import type { Common } from "@/internal/common.js";
 import { NonRetryableError, ShutdownError } from "@/internal/errors.js";
 import type {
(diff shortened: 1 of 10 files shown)


packages/core/src/database/index.ts
```diff
@@ -654,202 +649,8 @@ export const createDatabase = async ({
...
-            .dropTable("_ponder_status")
-            .ifExists()
-            .cascade()
-            .execute();
-        }
-      }
-
       await this.wrap(
-        { method: "migrate", includeTraceLogs: true },
+        { method: "createPonderSystemTables", includeTraceLogs: true },
```
typedarray, line 653: Should these statements use a transaction?

packages/core/src/database/index.ts
```diff
@@ -985,84 +811,88 @@ EXECUTE PROCEDURE "${namespace}".${notification};`),
...
-              // write metadata
+              await tx.update(PONDER_META).set({ value: metadata });
+              return {
+                status: "success",
+                crashRecoveryCheckpoint: undefined,
+              } as const;
+            }
+
+            if (
+              common.options.command === "dev" ||
```
typedarray, line 830: Should this be an and?

packages/core/src/database/index.ts
```diff
@@ -985,84 +811,88 @@ EXECUTE PROCEDURE "${namespace}".${notification};`),
...
+            for (const indexStatement of schemaBuild.statements.indexes.json) {
+              await tx.execute(
+                sql.raw(
+                  `DROP INDEX IF EXISTS "${namespace}"."${indexStatement.data.name}"`,
+                ),
+              );
               common.logger.info({
                 service: "database",
-                msg: `Detected crash recovery for build '${buildId}' in schema '${namespace}' last active ${formatEta(Date.now() - previousApp!.heartbeat_at)} ago`,
+                msg: `Dropped index '${indexStatement.data.name}' in schema '${namespace}'`,
```
typedarray, line 876: This should probably be debug level

packages/core/src/sync/index.ts
```diff
@@ -350,12 +350,13 @@ export const createSync = async (params: {
           for await (const { events, checkpoint } of eventGenerator) {
             // Sort out any events before the crash recovery checkpoint
             if (
+              params.crashRecoveryCheckpoint &&
```
typedarray, line 353: Before, we used the zero checkpoint string as the empty value here, correct? So, this diff should have no functional change?

packages/core/src/database/index.ts
kyscott18: No, this is logically what I intended. 

An error is thrown: 
1. if a `dev` deployment` is trying to replace a `start` deployment
2. if a `start` deployment is trying to replace a `start` deployment but `buildId` is different

packages/core/src/database/index.ts
kyscott18: The `IF NOT EXISTS` or `OR REPLACE` make all the statements in this block idempotent

packages/core/src/sync/index.ts
kyscott18: Correct. Now we use `undefined` to represent an app with no crash recovery instead of `zeroCheckpoint`.

PR #1680, opened 2025-04-16
improve build robustness

kyscott18:
- Fixed dependency rules between `ponder.config.ts`, `ponder.schema.ts`, and `src/**.ts`. Now all of these can import from each other without issue.
- Fixed dependency rules for `src/api/index.ts`. No imports are allowed from `src/api/index.ts`. 
Screenshot 2025-04-18 at 2 02 01 PM

### DIFF
--- a/packages/core/src/bin/commands/dev.ts
+++ b/packages/core/src/bin/commands/dev.ts
@@ -115,7 +115,7 @@ export async function dev({ cliOptions }: { cliOptions: CliOptions }) {
           return;
         }
 
-        const schemaResult = await build.executeSchema({ namespace });
+        const schemaResult = await build.executeSchema();
         if (schemaResult.status === "error") {
           buildQueue.add({
             status: "error",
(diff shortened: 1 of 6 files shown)


packages/core/src/build/plugin.ts
```diff
@@ -18,6 +18,10 @@ export default schema;
 
 const apiModule = () => `import { createPublicClient } from "viem";
 
+if (globalThis.PONDER_INDEXING_BUILD === undefined || globalThis.PONDER_DATABASE === undefined) {
+  throw new Error('Invalid dependency graph. Please ensure that nothing is imported from "src/api/index.ts"')
```
typedarray, line 22: Invalid dependency graph. Config, schema, and indexing function files cannot import objects from the API function file "src/api/index.ts".

packages/core/src/bin/commands/dev.ts
```diff
@@ -292,6 +292,8 @@ export async function dev({ cliOptions }: { cliOptions: CliOptions }) {
   const namespace =
     cliOptions.schema ?? process.env.DATABASE_SCHEMA ?? "public";
 
+  globalThis.PONDER_NAMESPACE_BUILD = namespace;
```
typedarray, line 295: Is there a reason we don't use namespaceCompile?

kyscott18: There are different rules for start vs dev, and namespaceCompile implements the start rules

typedarray (approved):

typedarray (approved):

PR #1704, opened 2025-04-28
New `_ponder_checkpoint` table with support for multiple chains

kyscott18:
## Overview
The primary feature this unlocks is multichain ordering during the historical backfill and more precise crash recoveries.
It also improves performance for "finalize" step during historical backfill (we noticed it was longer than expected).

## `_ponder_checkpoint`
- Replaced `_ponder_status` with `_ponder_checkpoint`
- removed `checkpoint` column from `_ponder_meta`
- `_ponder_checkpoint` has both "latest" and "safe" checkpoint for each chain
- Incremented  user schema `VERSION`  #1649 
https://github.com/ponder-sh/ponder/blob/42e9feec5f9b62fa4dce84f04e334eed30675d4f/packages/core/src/database/index.ts#L135

## multichain ordering
- Improved performance for "multichain" `getEvents` with better merging algorithm
- multichain crash recovery skips re-extracting events when possible

## other
- added `is_ready` column to `_ponder_meta`
- added index on `checkpoint` column for reorg tables
- removed `ready` property from `status` type
- increased max event batch size to 18000 https://github.com/ponder-sh/ponder/blob/80536244268af646b7941d9fe525133f277cd04a/packages/core/src/internal/options.ts#L121

### DIFF
--- a/packages/core/src/bin/utils/run.ts
+++ b/packages/core/src/bin/utils/run.ts
@@ -24,6 +24,7 @@ import { createMutex } from "@/utils/mutex.js";
 import { never } from "@/utils/never.js";
 import { createRequestQueue } from "@/utils/requestQueue.js";
 import { startClock } from "@/utils/timer.js";
+import { eq } from "drizzle-orm";
 
 /** Starts the sync and indexing services for the specified build. */
 export async function run({
@@ -141,7 +142,6 @@ export async function run({
 
   // If the initial checkpoint is zero, we need to run setup events.
   if (crashRecoveryCheckpoint === undefined) {
-    // TODO(kyle) how to mark setup as complete?
     await database.retry(async () => {
       await database.transaction(async (client, tx) => {
         const historicalIndexingStore = createHistoricalIndexingStore({
@@ -162,6 +162,15 @@ export async function run({
       });
     });
   }
+
+  await database.setCheckpoints({
+    checkpoints: indexingBuild.networks.map((network) => ({
+      chainId: network.chainId,
+      latestCheckpoint: sync.getStartCheckpoint(network),
+      safeCheckpoint: sync.getStartCheckpoint(network),
+    })),
+    db: database.qb.drizzle,
+  });
 
   // Run historical indexing until complete.
   for await (const events of recordAsyncGenerator(
@@ -292,7 +301,16 @@ export async function run({
             );
             endClock = startClock();
 
-            await database.setLatestCheckpoint(events.checkpoints);
+            await database.setCheckpoints({
+              checkpoints: events.checkpoints.map(
+                ({ chainId, checkpoint }) => ({
+                  chainId,
+                  latestCheckpoint: checkpoint,
+                  safeCheckpoint: checkpoint,
+                }),
+              ),
+              db: tx,
+            });
 
             common.metrics.ponder_historical_transform_duration.inc(
               { step: "finalize" },
@@ -344,6 +362,15 @@ export async function run({
   common.logger.info({
     service: "indexing",
     msg: "Completed historical indexing",
+  });
+
+  await database.setCheckpoints({
+    checkpoints: indexingBuild.networks.map((network) => ({
+      chainId: network.chainId,
+      latestCheckpoint: sync.getFinalizedCheckpoint(network),
+      safeCheckpoint: sync.getFinalizedCheckpoint(network),
+    })),
+    db: database.qb.drizzle,
   });
 
   await database.setReady();
@@ -408,14 +435,19 @@ export async function run({
           // }
         }
 
-        await database.setLatestCheckpoint([
-          { chainId: event.chainId, checkpoint: event.checkpoint },
-        ]);
+        await database.qb.drizzle
+          .update(database.PONDER_CHECKPOINT)
+          .set({
+            latestCheckpoint: event.checkpoint,
+          })
+          .where(eq(database.PONDER_CHECKPOINT.chainId, event.chainId));
 
         break;
       }
       case "reorg":
-        // TODO(kyle) database.setLatestCheckpoint
+        // Note: `setLatestCheckpoint` is not called here, instead it is called
+        // in the `block` case.
+
         await database.removeTriggers();
         await database.retry(async () => {
           await database.qb.drizzle.transaction(async (tx) => {
@@ -427,10 +459,10 @@ export async function run({
         break;
 
       case "finalize":
-        // TODO(kyle) database.setSafeCheckpoint
-        // await database.setSafeCheckpoint([
-        //   { chainId: event.chainId, checkpoint: event.checkpoint },
-        // ]);
+        await database.qb.drizzle.update(database.PONDER_CHECKPOINT).set({
+          safeCheckpoint: event.checkpoint,
+        });
+
         await database.finalize({
           checkpoint: event.checkpoint,
           db: database.qb.drizzle,
--- a/packages/core/src/bin/utils/run.ts
+++ b/packages/core/src/bin/utils/run.ts
@@ -256,7 +256,6 @@ export async function run({
                   Number(checkpoint.blockTimestamp),
                 );
               } else {
-                // TODO(kyle) does this handle networks with end blocks?
                 for (const network of indexingBuild.networks) {
                   common.metrics.ponder_historical_completed_indexing_seconds.set(
                     { network: network.name },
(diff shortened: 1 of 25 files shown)


packages/core/src/utils/generators.test.ts
```diff
@@ -74,7 +74,7 @@ test("mergeAsyncGenerators() yields all results", async () => {
 
   await new Promise((res) => setTimeout(res));
 
-  expect(results).toStrictEqual([1, 3, 2, 4]);
+  expect(results).toStrictEqual([1, 2, 3, 4]);
```
typedarray, line 77: What happened here?

packages/core/src/sync/index.test.ts
```diff
@@ -986,7 +991,16 @@ test("mergeAsyncGeneratorsWithEventOrder()", async () => {
   expect(results).toMatchInlineSnapshot(`
     [
       {
-        "checkpoint": "06",
+        "checkpoints": [
+          {
+            "chainId": undefined,
```
typedarray, line 996: This looks a bit strange

packages/core/src/graphql/middleware.ts
```diff
@@ -57,7 +57,7 @@ export const graphql = (
 
       return {
         getDataLoader,
-        getStatus: () => globalThis.PONDER_DATABASE.getStatus(),
+        getCheckpoints: () => globalThis.PONDER_DATABASE.getCheckpoints(),
```
typedarray, line 60: I wonder if it would ever make sense to do this using the available drizzle `db` object here rather than the global.

packages/core/src/drizzle/kit/index.ts
```diff
@@ -45,18 +46,26 @@ export const getReorgTable = <config extends TableConfig>(
...
   if (schema && schema !== "public") {
-    return pgSchema(schema).table(sqlToReorgTableName(getTableName(table)), {
+    return pgSchema(schema).table(
+      sqlToReorgTableName(getTableName(table)),
+      {
+        operation_id: serial().notNull().primaryKey(),
+        operation: integer().notNull().$type<0 | 1 | 2>(),
+        checkpoint: varchar({ length: 75 }).notNull(),
+      },
+      (table) => [index().on(table.checkpoint)],
```
typedarray, line 56: Makes sense, but I would not be surprised if this has a non-trivial perf and storage impact - checkpoint strings are long.

packages/core/src/database/index.ts
```diff
@@ -69,21 +72,51 @@ export type Database = {
...
+  setCheckpoints: ({
+    checkpoints,
+  }: {
+    checkpoints: {
+      chainName: string;
+      chainId: number;
+      safeCheckpoint: string;
+      latestCheckpoint: string;
+    }[];
+    db: Drizzle<Schema>;
```
typedarray, line 102: I understand why (sometimes in transaction, sometimes not) but noting that this looks a bit weird.

packages/core/src/database/index.ts
```diff
@@ -144,21 +177,21 @@ export const getPonderMeta = (namespace: NamespaceBuild) => {
   }));
 };
 
-export const getPonderStatus = (namespace: NamespaceBuild) => {
+export const getPonderCheckpoint = (namespace: NamespaceBuild) => {
```
typedarray, line 180: I'd probably call this `getPonderCheckpointTable`

packages/core/src/sync/index.ts
```diff
@@ -383,42 +372,68 @@ export const createSync = async (params: {
...
           ),
-          limit: Math.round(
-            params.common.options.syncEventsQuerySize /
-              (params.indexingBuild.networks.length + 1),
-          ),
+          limit:
+            Math.round(
+              params.common.options.syncEventsQuerySize /
+                (params.indexingBuild.networks.length + 1),
+            ) + 6,
```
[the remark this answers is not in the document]
kyscott18: lol, I thought it was really annoying that the batch size was consistently 1999 because of pagination, so I added some extra.

packages/core/src/sync/index.ts
```diff
@@ -341,29 +316,43 @@ export const createSync = async (params: {
           }
         }
 
-        async function* sortCompletedAndPendingEvents(
+        async function* sortCrashRecoveryEvents(
```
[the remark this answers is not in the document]
kyscott18: It's sorting the events into those that need to be processed now and those that need to be processed in realtime (there is also a function for crash recovery purposes). I guess sort commonly gets confused with ordering, which this function is not doing.

packages/core/src/database/index.ts
```diff
@@ -877,18 +920,19 @@ EXECUTE PROCEDURE "${namespace}".${notification};`),
...
+              ...checkpoints.map((c) => c.safeCheckpoint),
+            );
+
+            await this.revert({ checkpoint: revertCheckpoint, tx });
 
-            // TODO(kyle) eventually improve this using checkpoint for each chain
             await tx
-              .update(PONDER_STATUS)
-              .set({ block_number: null, block_timestamp: null, ready: false });
+              .update(PONDER_CHECKPOINT)
```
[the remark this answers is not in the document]
kyscott18: We could do that, the reason I didn't want to is that I want `latestCheckpoint` to be chain specific and `revertCheckpoint` would break this invariant. I guess `ZERO_CHECKPOINT` doesn't either, but it seems more clear.

This column gets used to server the status endpoint, so it would be incorrect to serve data that corresponds to a different chain.

packages/core/src/utils/generators.test.ts
kyscott18: I changed the `mergeAsyncGenerators` function to "look ahead" by one.

https://github.com/ponder-sh/ponder/blob/3ece6520bdf0f9dac8c4b68a0df45337d20d2abe/packages/core/src/utils/generators.ts#L26-L27

packages/core/src/drizzle/kit/index.ts
kyscott18: Hopefully this table is giant, the old values get cleared out after every finalization.

packages/core/src/database/index.ts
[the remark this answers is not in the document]
typedarray: Got it, I see. So in practice, the checkpoint will only be null/zero for a brief time (between when the process starts and when the first event batch gets indexed on this chain) ?

And, what do we return as the `block` value from the `/status` endpoint when the checkpoint is zero?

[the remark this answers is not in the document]
kyscott18: Correct on both.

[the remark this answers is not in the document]
kyscott18: Generally, I'm not completely sure on what the empty state should be for the status endpoint, once we figure that out I can work backwards and figure out the rest.

packages/core/src/sync/index.ts
```diff
@@ -383,42 +372,68 @@ export const createSync = async (params: {
...
+        let from: string;
+        if (
+          crashRecoveryCheckpoint &&
+          Number(decodeCheckpoint(crashRecoveryCheckpoint).chainId) ===
+            network.chainId
+        ) {
+          from = crashRecoveryCheckpoint;
+        } else {
+          from = getMultichainCheckpoint({ tag: "start", network })!;
+        }
```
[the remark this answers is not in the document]
kyscott18: Cross posting comment here:
```ts
// In order to speed up the "extract" phase when there is a crash recovery,
// the beginning cursor is moved forwards. This only works when `crashRecoveryCheckpoint`
// is defined and `crashRecoveryCheckpoint` refers to the same chain as `network`.
```
The reason for this is that our pagination logic uses "blockNumber`, so we need to check that the checkpoint refers to the correct chain before using it.

packages/core/src/database/index.ts
kyscott18: I have plans to fix, but out of the scope of this pr.

packages/core/src/bin/utils/run.ts
```diff
@@ -105,8 +107,6 @@ export async function run({
     eventCount,
   });
 
-  await database.setStatus(sync.getStatus());
```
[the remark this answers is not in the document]
kyscott18: The logic got moved here
https://github.com/ponder-sh/ponder/blob/6c0bfe1a69c54afd954b73faa5b0f5c4cb324566/packages/core/src/bin/utils/run.ts#L166-L174

PR #1706, opened 2025-04-29
add `chains` config

### LINKED ISSUE #1292: feat: update `networks` config
Similar to https://wagmi.sh/core/getting-started#create-config

```ts
import { createConfig } from "ponder";
import { mainnet, optimism } from "viem/chains";

export default createConfig({
  chains: [mainnet, optimism],
  rpcUrls: {
    [mainnet.id]: string | string[],
    [optimism.id]: string | string[],
  }, 
  // ... more config
});
```

### Related
#1291 #1290

kyscott18:
Introduced new `chains` config. Closes #1292, #1291 
```ts
import { createConfig } from "ponder";

export default createConfig({
  chains: {
    mainnet: {
      id: 1,
      rpc: process.env.PONDER_RPC_URL_1!,
    },
  },  
  // ... more config
});
```

Also updated `network` tag to `chain` for all metrics.


packages/core/src/internal/types.ts
```diff
@@ -281,26 +282,24 @@ export type ContractMetadata = {
...
-// Network
+// Chain
 
-export type Network = {
-  name: string;
-  chainId: number;
-  transport: ReturnType<Transport>;
-  chain: Chain;
+export type Chain = {
+  chain: ViemChain;
```
typedarray, line 301: Nit, but I'd probably include `name: string` and `id: number` here as the chain name and ID that were provided directly by the user, and consider renaming `chain: ViemChain` to `viemChain: ViemChain`. The Viem chain feels secondary / "best effort".

This would change `args.chain.chain.id` -> `args.chain.id` in downstream files which makes more sense to me.

packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -103,56 +104,111 @@ export async function buildConfigAndIndexingFunctions({
...
+      );
+
+      if (matchedChain === undefined) {
+        throw new Error(
+          `Chain "${chainName}" with id ${chain.id} not found in viem. Please update viem to the latest version.`,
+        );
+      }
+      matchedChain.name = chainName;
+
+      if (chain.rpcUrl === undefined) {
```
typedarray, line 130: Nit but would be more readable to coalesce these three paths to an array of strings first and check each entry each once

typedarray (commented):
Partial review - will follow up re: `rpcUrl` vs `transport` naming

packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -883,30 +933,35 @@ export async function buildConfigAndIndexingFunctions({
...
     logs,
   };
 }
 
 export async function safeBuildConfigAndIndexingFunctions({
+  common,
   config,
   rawIndexingFunctions,
 }: {
+  common: Common;
```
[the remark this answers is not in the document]
kyscott18: If we did that, the error case wouldn't be the same. Now, if there is an error, no warning logs get printed and if we did the logging inline, that wouldn't be true. Maybe that doesn't matter but just pointing it out.

PR #1721, opened 2025-05-02
improve logging for lagging chains

kyscott18:


kyscott18:
@typedarray I think this is logically correct.

The "latest" checkpoint always refers to the chain that has the minimum checkpoint.
https://github.com/ponder-sh/ponder/blob/9eabe29fb674abe91bf02bfc3a79bc37be8a657d/packages/core/src/sync/index.ts#L266

typedarray (approved):

PR #1738, opened 2025-05-14
allow `rpc` to be undefined

kyscott18:


packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -119,16 +119,17 @@ export async function buildConfigAndIndexingFunctions({
...
-      //   );
-      // }
       if (chain.rpc === undefined) {
-        throw new Error(
-          `Chain "${chainName}" with id ${chain.id} has no RPC URL.`,
-        );
-      } else if (typeof chain.rpc === "string" || Array.isArray(chain.rpc)) {
+        if (matchedChain === undefined) {
+          throw new Error(
+            `Chain "${chainName}" with id ${chain.id} has no RPC URL.`,
```
typedarray, line 125: We could probably make this more useful, eg "has no RPC defined and no default RPC URL was found in viem/chains"

PR #1751, opened 2025-05-17
fix: one -> many relations

### LINKED ISSUE #1745: v0.10.25 breaks one -> many relations between entities.
### Version

0.10.27

### Current behavior

When a user queries their owned "heroes" via the query 
```
user(address: <ADDRESS>) {
  heroes {
    items {
      id
    }
  }
}
```

the heroes returned is an empty array. i.e. one -> many relation does not work.

BUT 

We can query the heroes like
```
heroes(where: { owner: <ADDRESS> }) {
  items {
    id
  }
}
```

And the correct heroes will be returned. i.e. many -> one works. 

TL;DR 

one -> many relationship broken, but many -> one still works.

See steps to reproduce for schema and queries used.

### Expected behavior

The owned heroes are returned like the following image (Same code, but ran from version 0.10.24) 

Image

### Steps to reproduce

Schema 
```ts
export const user = onchainTable('user', t => ({
  address: t.hex().primaryKey(),
  ...rest of user fields
}));

export const hero = onchainTable('hero', t => ({
  id: t.bigint().primaryKey(),
  owner: t.hex().notNull(),
  ...rest of hero fields
}));

export const userRelations = relations(user, ({ many }) => ({
  heroes: many(hero),
  ... rest of relations
}));

export const heroRelations = relations(hero, ({ one, many }) => ({
  owner: one(user, { fields: [hero.owner], references: [user.address] }),
  ... rest of relations
}));
```

Query used: 
```
query GetUser {
  user(address: <ADDRESS>) {
    heroes {
      items {
        id
      }
    }
  }
  heros(where: {owner: <ADDRESS>}) {
    items {
      id
      owner {
        address
      }
    }
  }
}
```

### Link to repository

_No response_

### Anything else?

_No response_

kyscott18:
Closes #1745

### DIFF
--- a/packages/core/src/graphql/index.ts
+++ b/packages/core/src/graphql/index.ts
@@ -266,36 +266,46 @@ export function buildGraphQLSchema({
             };
           } else if (is(relation, Many)) {
             // Search the relations of the referenced table for the corresponding `one` relation.
-            // If "relationName" is not provided, use the first `one` relation that references this table.
+            // If `relation.relationName` is not provided, use the first `one` relation that references this table.
 
             let oneRelation: One | undefined;
 
-            for (const _relation of Object.values(referencedTable.relations)) {
-              if (
-                is(_relation, One) &&
-                relation.relationName === _relation.relationName
-              ) {
-                oneRelation = _relation;
-              }
-            }
+            // Note: can find the wrong relation if `relationName` is undefined.
+            // The first relation will be found.
 
-            if (oneRelation === undefined) {
-              for (const _relation of Object.values(
+            if (relation.relationName !== undefined) {
+              for (const referencedRelation of Object.values(
                 referencedTable.relations,
               )) {
                 if (
-                  is(_relation, One) &&
-                  table.dbName === _relation.referencedTableName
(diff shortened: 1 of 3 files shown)


packages/core/src/graphql/index.ts
```diff
@@ -266,36 +266,46 @@ export function buildGraphQLSchema({
...
-            for (const _relation of Object.values(referencedTable.relations)) {
-              if (
-                is(_relation, One) &&
-                relation.relationName === _relation.relationName
-              ) {
-                oneRelation = _relation;
+            // Note: can find the wrong relation if `relationName` is undefined.
+            // The first relation will be found.
+
+            if (relation.relationName !== undefined) {
```
kyscott18, line 276: @typedarray The root of this bug is using `relationName` to try and find the corresponding "one" relation without first checking if `relationName` is `undefined`.

If `relationName` is not used in the relations (commonly is not) then the first comparison will return true because both `relation.relationName` and `referencedRelation.relationName` will be `undefined`.

typedarray (approved):
LGTM

PR #1755, opened 2025-05-18
drop views before creating

kyscott18:


typedarray (approved):

packages/core/src/bin/commands/createViews.ts
```diff
@@ -104,9 +104,14 @@ export async function createViews({
...
   for (const table of meta[0]!.app.table_names) {
+    // Note: drop views before creating new ones to avoid enum errors.
+    await database.qb.drizzle.execute(
+      sql.raw(`DROP VIEW IF EXISTS "${cliOptions.viewsSchema}"."${table}"`),
+    );
+
     await database.qb.drizzle.execute(
       sql.raw(
-        `CREATE OR REPLACE VIEW "${cliOptions.viewsSchema}"."${table}" AS SELECT * FROM "${cliOptions.schema}"."${table}"`,
+        `CREATE "${cliOptions.viewsSchema}"."${table}" AS SELECT * FROM "${cliOptions.schema}"."${table}"`,
```
WuBruno, line 114: Would it be appropriate to bundle everything in a single transaction to avoid potential conflicting queries whilst this migration is happening?

razvanparaschiv (approved):

kyscott18:
> i was wondering about transaction as well

We've been in the habit of only using transactions when logically necessary. Have plans to change it wholesale but gonna stick to this for now.

PR #1762, opened 2025-05-20
integration testing

### LINKED ISSUE #1773: if start and end block are the same, nothing is indexed
### Version

0.10.26

### Current behavior

if the start and end blocks for a contract are the same value, nothing is indexed even though it feels like that exact block's transactions should be indexed

if this is expected behavior, that's totally fine, feel free to close

### Expected behavior

_No response_

### Steps to reproduce

_No response_

### Link to repository

_No response_

### Anything else?

_No response_

kyscott18:
e2e integration testing with rpc simulation.

## Bug Fixes

- `getLocalEventGenerator` pagination (closes #1773)
- `realtime_latency` metric
- `transaction.to` filtering
- reorg handling causing missed events x 2
-  checkpoint calculation x 2


packages/core/src/build/configAndIndexingFunctions.ts
```diff
@@ -199,6 +203,30 @@ export async function buildConfigAndIndexingFunctions({
     }),
   );
 
+  const finalizedBlocks = await Promise.all(
```
typedarray, line 206: Should `finalizedBlock` be part of our internal `Chain` type?

packages/core/src/sync-realtime/index.ts
```diff
@@ -243,7 +255,7 @@ export const createRealtimeSync = (
    */
   const fetchBlockEventData = async (
     block: SyncBlock,
-  ): Promise<BlockWithEventData> => {
+  ): Promise<Omit<BlockWithEventData, "callback">> => {
```
typedarray, line 258: Doesn't seem like `BlockWithEventData` has a `callback` property - maybe leftover?

packages/core/src/sync-realtime/index.ts
```diff
@@ -869,53 +865,57 @@ export const createRealtimeSync = (
...
 
-        const blockWithEventData = filterBlockEventData({ block, ...rest });
+        const blockWithEventDataAndFilters =
+          filterBlockEventData(blockWithEventData);
 
         if (
-          blockWithEventData.logs.length > 0 ||
-          blockWithEventData.traces.length > 0 ||
-          blockWithEventData.transactions.length > 0
+          blockWithEventDataAndFilters.logs.length > 0 ||
```
typedarray, line 896: `blockWithFilteredEventData` ?

packages/core/src/sync/index.ts
```diff
@@ -659,11 +679,7 @@ export const createSync = async (params: {
 
         // Raise event to parent function (runtime)
         if (to > from) {
-          params.onRealtimeEvent({
-            type: "finalize",
-            chain,
-            checkpoint: to,
-          });
+          params.onRealtimeEvent({ type: "finalize", chain, checkpoint: to });
```
typedarray, line 682: unawaited?

packages/core/src/sync/index.ts
```diff
@@ -1497,27 +1541,19 @@ export const getLocalSyncProgress = async ({
     cached === undefined
       ? [
           rpc.request({ method: "eth_chainId" }),
-          _eth_getBlockByNumber(rpc, { blockTag: "latest" }),
```
typedarray, line 1500: Just noting that this will probably change the experience slightly for folks who have a totally broken / misconfigured RPC, because the first request will now happen during build.

PR #1777, opened 2025-05-23
graphql-yoga != 200 response status

munikes:
Response status and statustext are read-only properties for the Response object. So when I Dockerize Ponder, it fails because instead of responding with a PonyfillResponse, it responds with a Response. That's why I added an if statement, but it's not a definitive solution. If Yoga, instead of responding with a PonyfillResponse, responds with a Response object in future versions and gives an error != 200, it will fail for everyone, both locally and in Docker.

It's curious that without Docker, it responds with a PonyfillResponse object, which can be written to, and when I Dockerize Ponder, it responds with a Response object. I don't know why the objects change.

Thanks to Yoga responding with a 200 in production, this workaround can be done. I've noticed that in testing it gives a 500.

I hope this helps.

Thank you so much for the work you do with Ponder; it's fantastic.


typedarray:
Thanks for opening, two questions:
1. It's not clear to me why running in Docker has any impact on Yoga returning Response vs PonyfillResponse. Do you know why?
2. I also don't know what you mean by this "Thanks to Yoga responding with a 200 in production" - could you elaborate?

PR #1779, opened 2025-05-27
improve rpc validations

kyscott18:
Improved rpc validations by
- moving all rpc validations to a single file (`utils/rpc.ts`)
- adding stricter validations to all response types
- sanitizing the responses, filling in default values for non-required properties

### DIFF
--- a/packages/core/src/sync-historical/index.ts
+++ b/packages/core/src/sync-historical/index.ts
@@ -526,7 +526,7 @@ export const createHistoricalSync = async (
         if (log.transactionHash === zeroHash) {
           args.common.logger.warn({
             service: "sync",
-            msg: `Detected '${args.chain.name}' log with empty transaction hash in block ${block.hash} at log index ${hexToNumber(log.logIndex)}. This is expected for some networks like ZKsync.`,
+            msg: `Detected '${args.chain.name}' log with empty transaction hash in block ${block.hash} at log index ${hexToNumber(log.logIndex)}. This is expected for some chains like ZKsync.`,
           });
         }
       }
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/rpc.ts
```diff
@@ -300,5 +338,538 @@ export const _debug_traceBlockByHash = (
...
+          `Detected inconsistent RPC responses. 'log.transactionIndex' ${log.transactionIndex} not found in 'block.transactions' for block ${block.hash}`,
+        );
+      } else if (transaction.hash !== log.transactionHash) {
+        throw new Error(
+          `Detected inconsistent RPC responses. 'log.transactionHash' ${log.transactionHash} does not match 'transaction.hash' ${transaction.hash} for block ${block.hash}`,
+        );
+      }
+    }
+
+    const id = `${log.blockNumber}-${log.logIndex}`;
```
typedarray, line 412: Could probably standardize on just using logIndex instead of this ID concept.

packages/core/src/sync-store/encode.ts
```diff
@@ -33,7 +33,7 @@ export const encodeBlock = ({
   receiptsRoot: block.receiptsRoot,
   transactionsRoot: block.transactionsRoot,
   sha3Uncles: block.sha3Uncles ?? null,
-  size: block.size ? hexToBigInt(block.size) : 0n,
+  size: hexToBigInt(block.size),
```
typedarray, line 36: Now that we're setting the default values in the "standardize" functions, there are probably more spots in this file we can simplify right?

Fine to do as a follow-up, just checking for understanding.

kyscott18: For all the previous cases of RPC not providing all properties before `block.size`, we would just mark the column as nullable.

https://github.com/ponder-sh/ponder/blob/acee34601710394e0e2094e35c9c0a4757042c2f/packages/core/src/sync-store/migrations.ts#L442-L457

PR #1782, opened 2025-05-28
improve raw sql speed with query parsing

kyscott18:


packages/core/src/client/validate.ts
```diff
@@ -103,6 +103,68 @@ export const validateQuery = async (
...
 };
 
+/**
+ * Find all table names in a SQL query.
+ *
+ * @param sql - SQL query
+ */
+export const findTableNames = async (sql: string) => {
+  // @ts-ignore
+  const Parser = await import(/* webpackIgnore: true */ "pg-query-emscripten");
```
typedarray, line 113: I assume the import will be cached, but may make sense to do so explicitly in the file scope?

packages/core/src/client/validate.ts
```diff
@@ -103,6 +103,68 @@ export const validateQuery = async (
...
+
+  if (sql.length > 5_000) {
+    throw new Error("Invalid query");
+  }
+
+  const { parse } = await Parser.default();
+  const parseResult = parse(sql) as {
```
typedarray, line 120: Perhaps this function should accept a `parseResult` - I think at the moment we'll parse the query twice needlessly

packages/core/src/client/validate.test.ts
```diff
@@ -1,7 +1,28 @@
 import { eq, onchainTable, relations } from "@/index.js";
 import { createClient } from "@ponder/client";
 import { expect, test } from "vitest";
-import { validateQuery } from "./validate.js";
+import { findTableNames, validateQuery } from "./validate.js";
+
+test("findTableNames", async () => {
```
typedarray, line 6: Would be good to see a test query that included qualified table names (with schema)

packages/core/src/client/validate.ts
kyscott18: Added a LRU cache for `findTableNames` that will give us the desired behavior.

packages/core/src/indexing-store/historical.ts
```diff
@@ -287,18 +287,25 @@ export const createHistoricalIndexingStore = ({
...
       async (_sql, params, method, typings) => {
-        await indexingCache.flush({ client });
+        // Note: Not all nodes are implemented in the parser,
+        // so we need to try/catch to avoid throwing an error.
+        let tableNames: Set<string> | undefined;
+        try {
+          tableNames = await findTableNames(_sql);
+        } catch {}
 
         let safeQuery = false;
```
kyscott18: `isSelectOnly`

PR #1789, opened 2025-05-30
fix: move 0.10 migration logs to debug level

typedarray:
Moved migration logs to debug level

Screenshot 2025-05-30 at 12 26 15 PM

### DIFF
--- a/packages/core/src/database/index.ts
+++ b/packages/core/src/database/index.ts
@@ -15,7 +15,7 @@ import type {
   Schema,
   SchemaBuild,
 } from "@/internal/types.js";
-import { migrationProvider } from "@/sync-store/migrations.js";
+import { buildMigrationProvider } from "@/sync-store/migrations.js";
 import * as ponderSyncSchema from "@/sync-store/schema.js";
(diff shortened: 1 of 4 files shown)


packages/core/src/sync-store/migrations.ts
```diff
@@ -1,7 +1,22 @@
+import { type Logger, createNoopLogger } from "@/internal/logger.js";
 import type { Kysely, Migration, MigrationProvider } from "kysely";
 import { sql } from "kysely";
 import { maxUint256 } from "viem";
 
+let logger = createNoopLogger();
```
kyscott18, line 6: Do you know what the noop logger would ever get used?

packages/core/src/database/index.ts
```diff
@@ -663,6 +663,7 @@ export const createDatabase = async ({
             plugins: [new WithSchemaPlugin("ponder_sync")],
           });
 
+          const migrationProvider = buildMigrationProvider(common.logger);
```
kyscott18, line 666: `globalThis.PONDER_COMMON` is also available.

PR #1790, opened 2025-05-30
Dynamic load balancing rpc transport

khaidarkairbek:
Handles rate-limiting by tracking requests per second metric and increasing/decreasing the cap according to the track of successful requests and 429 errors. Handles dynamic load balancing by tracking the latency of requests and prioritizing transports with lower latency. This transport aims to improve latency and throughput of rpc requests given a list of rpc urls of varying quality. 

Deprecated `maxRequestsPerSecond` config property.


packages/utils/src/dynamicLB.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    const queue: [EIP1193Parameters, PromiseWithResolvers<unknown>][] = [];
+
+    const buckets = _transports.map((transport, index) => ({
+      index,
+      retryDelay: 0,
+
+      activeConnections: 0,
+      isActive: true,
+      isJustActivated: false,
```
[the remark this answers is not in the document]
kyscott18: Realized you already have that information with `activeConnections`.

packages/utils/src/dynamicLB.ts
```diff
@@ -0,0 +1,299 @@
...
+    const INITIAL_MAX_RPS = 50;
+    const MIN_RPS = 1;
+    const MAX_RPS = 500;
+    const RPS_INCREASE_FACTOR = 1.2;
+    const RPS_DECREASE_FACTOR = 0.7;
+    const SUCCESS_WINDOW_SIZE = 100;
+
+    const queue: [EIP1193Parameters, PromiseWithResolvers<unknown>][] = [];
+
+    const buckets = _transports.map((transport, index) => ({
```
kyscott18, line 36: Can you define a type out of the function scope that defines a `Bucket`. This type should probably only have values, no functions. The functions should also be defined out of the scope of `dynamicLB` and take `Bucket` as a parameter.

packages/utils/src/dynamicLB.ts
```diff
@@ -0,0 +1,299 @@
...
+    setInterval(printRPCUsage, 5000);
+
+    return createTransport({
+      key: "dynamicLB",
+      name: "dynamic load balance",
+      request: async (body) => {
+        const p = promiseWithResolvers();
+        queue.push([body, p]);
+        await dispatch();
+        return p.promise;
```
kyscott18, line 292: If you inlined the dispatch function right here, would you still need `queue`?

packages/utils/src/dynamicLB.ts
```diff
@@ -0,0 +1,299 @@
...
+      index,
+      retryDelay: 0,
+
+      activeConnections: 0,
+      isActive: true,
+      isJustActivated: false,
+
+      latencies: [] as number[],
+      expectedLatency: 0,
+      requests: [] as number[],
```
kyscott18, line 46: It could be simpler to define an object that tracks all recent requests in a one object, and then compute expected latency just in time with a function. In a previous effort I tried this, where each element in the array represents a second.

```ts
  requests: {
    timestamp: number;
    response: number;
    count: number;
    errors: number;
  }[];
```

This has the added benefit of letting up figure out the historical requests per second for a bucket.

packages/core/src/rpc/index.ts
```diff
@@ -123,9 +345,18 @@ export const createRpc = ({
             stopClock(),
           );
 
+          addLatency(bucket, stopClock(), true);
```
typedarray: I'd move stopClock() out into a `duration` const and use the same value for the metric and the latency record.

packages/core/src/rpc/index.ts
```diff
@@ -45,73 +45,295 @@ export type Rpc = {
...
+        requestTimestamps: [],
+        consecutiveSuccessfulRequests: 0,
+        rpsLimit: INITIAL_MAX_RPS,
+
+        request,
+      }) satisfies Bucket,
+  );
+
+  const scheduleBucketActivation = (bucket: Bucket) => {
+    setTimeout(() => {
```
typedarray: May make sense to keep these timeout objects around and clear them on cleanup/shutdown.

packages/core/src/rpc/index.ts
```diff
@@ -45,73 +45,295 @@ export type Rpc = {
...
+    const availableBuckets = buckets.filter((b) => isAvailable(b));
+
+    if (availableBuckets.length === 0) {
+      const rpsSafePromise = new Promise<void>((resolve) => {
+        const checkRPSSafe = () => {
+          const availableBucket = buckets.find((b) => isAvailable(b));
+          if (availableBucket) {
+            resolve();
+          } else {
+            setTimeout(checkRPSSafe, 5);
```
typedarray: I wonder if there's a way to do this that doesn't create so many timeouts - IIRC they're somewhat CPU intensive, but I might be wrong.

khaidarkairbek: can't think of any to be honest. we can increase the timeout delay

kyscott18 (approved):

packages/core/src/internal/metrics.ts
```diff
@@ -85,7 +85,7 @@ export class MetricsService {
   >;
 
   ponder_rpc_request_duration: prometheus.Histogram<"chain" | "method">;
-  ponder_rpc_request_lag: prometheus.Histogram<"chain" | "method">;
+  ponder_rpc_request_error_total: prometheus.Counter<"chain" | "method">;
```
typedarray, line 88: Should include these metric changes in the changeset, and I believe we have the full metric list in the docs somewhere too.

packages/core/src/internal/metrics.ts
```diff
@@ -346,11 +346,11 @@ export class MetricsService {
       buckets: httpRequestDurationMs,
       registers: [this.registry],
     });
-    this.ponder_rpc_request_lag = new prometheus.Histogram({
-      name: "ponder_rpc_request_lag",
-      help: "Time RPC requests spend waiting in the request queue",
+
+    this.ponder_rpc_request_error_total = new prometheus.Counter({
```
typedarray, line 350: Should update `ponder_rpc_request_duration` help text to include the word "successful"

packages/core/src/rpc/index.ts
```diff
@@ -45,91 +45,327 @@ export type Rpc = {
...
-      request = http(chain.rpc)({
-        chain: chain.viemChain,
-        retryCount: 0,
-        timeout: 5_000,
-      }).request;
+      request = [
+        http(chain.rpc)({
+          chain: chain.viemChain,
+          retryCount: 0,
+          timeout: 5_000,
```
typedarray, line 189: Is this 5 second timeout different from Viem's default? It seems a bit low

packages/core/src/rpc/index.ts
```diff
@@ -163,12 +426,16 @@ export const createRpc = ({
           const duration = BASE_DURATION * 2 ** i;
           common.logger.debug({
             service: "rpc",
-            msg: `Failed ${body.method} request, retrying after ${duration} milliseconds`,
+            msg: `Failed ${body.method} request for chain '${chain.name}', retrying after ${duration} milliseconds`,
```
typedarray: These logs might look better as 
```
Failed '${chain.name}' ${body.method} request, ...
```

packages/core/src/rpc/index.ts
khaidarkairbek: Viem's default is 10_000

PR #1791, opened 2025-05-30
Add primary key update validation

kyscott18:
Add validations so that primary keys cannot be updated in `db.update().set` and `db.insert().values().onConflictDoUpdate()`


typedarray (approved):
This could be considering a breaking change, but I'm comfortable doing it in a patch - the new error message should be quite clear-cut.

PR #1794, opened 2025-06-02
improve database dx

kyscott18:
New `QB` type handles retries, logging, metrics, and transactions.

```ts
const qb = createQB(common, drizzle(pool, { casing: "snake_case" }));

const result = await qb("test").select().from(accounts);
```

- Created 4 postgres pools with a corresponding query builder (`sync`, `admin`, `user`, `readonly`)
- Added database error simulation to `integration-test`
- Added test query for testing database connection in `build` step
- Included #1804


packages/core/src/build/index.ts
```diff
@@ -370,6 +372,55 @@ export const createBuild = async ({
...
 
+      // diagnostic query
+      const dialect = preBuild.databaseConfig.kind;
+      if (dialect === "pglite") {
+        const driver = createPglite(preBuild.databaseConfig.options);
+        try {
+          await driver.query("SELECT version()");
+        } catch (e) {
+          const error = new BuildError(
+            `Failed to connect to database. Please check your database connection settings.\n\n${(e as any).message}`,
```
typedarray: `Failed to connect to PGlite database ...`

PR #1795, opened 2025-06-02
improve address checksum performance

kyscott18:


packages/core/src/utils/lazy.test.ts
```diff
@@ -0,0 +1,21 @@
...
+    {
+      address: "0x4838b106fce9647bdf1e7877bf73ce8b0bad5f97",
+    },
+    "address",
+  );
+
+  expect(lazyObj.address).toBe("0x4838B106FCe9647Bdf1E7877BF73cE8B0BAD5f97");
+});
+
+test("lazyChecksumAddress array", () => {
```
typedarray, line 15: Perhaps a test if the target property value is undefined?

PR #1801, opened 2025-06-04
feat: websocket for realtime

khaidarkairbek:
Added the feature to add websocket endpoint for realtime indexing, which will fallback to regular rpc polling in case of the fatal error.

### DIFF
--- a/packages/core/src/internal/types.ts
+++ b/packages/core/src/internal/types.ts
@@ -301,6 +301,7 @@ export type Chain = {
   name: string;
   id: number;
   rpc: string | string[] | Transport;
+  ws: string | undefined;
   pollingInterval: number;
   maxRequestsPerSecond: number;
   finalityBlockCount: number;
--- a/packages/core/src/rpc/index.ts
+++ b/packages/core/src/rpc/index.ts
@@ -3,7 +3,7 @@ import type { Common } from "@/internal/common.js";
 import type { Chain, SyncBlock } from "@/internal/types.js";
 import type { RealtimeSync } from "@/sync-realtime/index.js";
 import { createQueue } from "@/utils/queue.js";
-import { _eth_getBlockByNumber } from "@/utils/rpc.js";
+import { _eth_getBlockByHash, _eth_getBlockByNumber } from "@/utils/rpc.js";
 import { startClock } from "@/utils/timer.js";
 import { wait } from "@/utils/wait.js";
 import {
@@ -22,6 +22,7 @@ import {
   type PublicRpcSchema,
   type RpcError,
   TimeoutError,
+  type WebSocketTransport,
   isHex,
   webSocket,
 } from "viem";
@@ -234,6 +235,26 @@ export const createRpc = ({
         timeout: 5_000,
       }).request,
     ];
+  }
+
+  let ws_subscribe:
+    | NonNullable<ReturnType<WebSocketTransport>["value"]>["subscribe"]
+    | undefined = undefined;
+
+  if (typeof chain.ws === "string") {
+    const protocol = new url.URL(chain.rpc).protocol;
+
+    if (protocol === "wss:" || protocol === "ws:") {
+      ws_subscribe = webSocket(chain.ws)({
+        chain: chain.viemChain,
+        retryCount: 0,
+        timeout: 5_000,
+      }).value?.subscribe;
+    } else {
+      throw new Error(
+        `Inconsistent RPC URL protocol: ${protocol}. Expected wss or ws.`,
+      );
+    }
   }
 
   const buckets: Bucket[] = request.map((request, index) => ({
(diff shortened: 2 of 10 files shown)


packages/core/src/rpc/index.ts
```diff
@@ -432,23 +452,80 @@ export const createRpc = ({
...
+        }, chain.pollingInterval);
+        common.shutdown.add(() => {
+          clearInterval(interval);
+        });
+      } else {
+        wsTransport
+          .value!.subscribe({
+            params: ["newHeads"],
+            onData: (data) => {
+              _eth_getBlockByHash(rpc, { hash: data.result.hash })
```
kyscott18: This is because `data` is only a block header, and doesn't include full transactions? There's enough information `logsBloom` to make some progress, right? Could we change the `sync-realtime` so that the `eth_getBlock` requests can be skipped in the case when there are no matched logs?

packages/core/src/rpc/index.ts
```diff
@@ -432,23 +452,80 @@ export const createRpc = ({
...
+        interval = setInterval(() => {
+          _eth_getBlockByNumber(rpc, { blockTag: "latest" })
+            .then(onBlock)
+            .catch(_onError);
+        }, chain.pollingInterval);
+        common.shutdown.add(() => {
+          clearInterval(interval);
+        });
+      } else {
+        wsTransport
```
kyscott18: Is the `wsTransport.subscribe` return type useful at all?

packages/core/src/rpc/index.ts
khaidarkairbek: I think you are right, data has logsBloom which we can use to check if we need to request a block.

packages/core/src/rpc/index.ts
```diff
@@ -432,23 +452,80 @@ export const createRpc = ({
...
+          .value!.subscribe({
+            params: ["newHeads"],
+            onData: (data) => {
```
kyscott18: What happens if `data.error !== undefined`

packages/core/src/rpc/index.ts
```diff
@@ -432,23 +452,80 @@ export const createRpc = ({
   });
 
   let interval: NodeJS.Timeout | undefined;
+  let tryCount = 0;
```
kyscott18: Can you explain the difference between try count and retry count

kyscott18 (commented):
Great work!

packages/core/src/rpc/index.ts
khaidarkairbek: It returns subscriptionId and unsubscribe callback, we don't use it. On reconnect or unsubscribe, we close the socket, so no need for unsubscribe.

packages/core/src/rpc/index.ts
khaidarkairbek: do you mean reconnectCount? I was thinking about separating the two onError handlers, but on the second look we can reuse the same constant.

packages/core/src/rpc/index.ts
khaidarkairbek: Fixed this to behave similar to subscribe onError, but no reconnection is done for each error. Instead when maximum number of consecutive errors happen, fallback to polling right away since webSocket is faulty.

packages/core/src/internal/types.ts
```diff
@@ -400,6 +401,18 @@ export type LightBlock = Pick<
...
+  MakeOptional<RpcBlock<Exclude<BlockTag, "pending">, false>, "size">
+>;
+
+export const isSyncBlock = (
+  block: SyncBlock | SyncBlockHeader,
+): block is SyncBlock => {
+  return (
+    block.transactions.length === 0 || typeof block.transactions[0] !== "string"
+  );
+};
```
kyscott18: Nice type predicate!

packages/core/src/rpc/index.ts
```diff
@@ -432,23 +453,96 @@ export const createRpc = ({
...
+                const error = data.error as Error;
+                console.log(`Websocket returned an error: ${error}`);
+
+                if (retryCount++ === RETRY_COUNT) {
+                  common.logger.warn({
+                    service: "rpc",
+                    msg: `Failed "eth_subscribe" subscription after ${retryCount + 1} consecutive errors. Switching to polling.`,
+                    error,
+                  });
+                  wsTransport = undefined;
```
kyscott18: Should it unsubscribe before switching to polling? This also happens on line 505.

packages/core/src/rpc/index.ts
```diff
@@ -432,23 +453,96 @@ export const createRpc = ({
...
+                });
+                wsTransport = undefined;
+              } else {
+                await wsTransport!.value!.getRpcClient().then((r) => r.close());
+              }
+
+              rpc.subscribe({ onBlock, onError: _onError });
+            },
+          })
+          .then(() => {
```
kyscott18: I would find async await + try catch more readable for large blocks like this

kyscott18:
We might be able to simplify `sync-realtime` a little more once #1779 gets merged. I also think it would be good to add a feature to the integration tests that simulates blocks not having full transactions. Should be pretty easy to do this and I can show you more on Monday.

kyscott18 (commented):
Great work! Let's add some docs + before and after `ponder_realtime_arrival_latency`

kyscott18 (approved):

khaidarkairbek:
> Great work! Let's add some docs + before and after `ponder_realtime_arrival_latency`

The `ponder_realtime_arrival_latency` changes using webSocket in contrast to polling (average values using feature-multichain app):
optimism - 1243.97 -> 723.58 (-520.39 or -41.83%)
mainnet - 3346.39 -> 2566.78 (-779.61 or -23.29%)
base - 1959.11 -> 1417.25 (-541.86 or -27.66%)
polygon - 822.72 -> 477.51 (-345.21 or -42.96%)

docs/pages/docs/config/chains.mdx
```diff
@@ -128,6 +128,31 @@ Here are a few common transport options.
 * [`loadBalance`](/docs/api-reference/ponder-utils#loadbalance)
 * [`rateLimit`](/docs/api-reference/ponder-utils#ratelimit)
 
+### WebSocket
```
typedarray:
```
## WebSockets
```

docs/pages/docs/config/chains.mdx
```diff
@@ -128,6 +128,31 @@ Here are a few common transport options.
...
+
+Use the optional `ws` field to specify a WebSocket RPC endpoint for each chain.
+
+When provided, Ponder will use WebSocket connections for realtime block subscriptions instead of polling. Websocket connections offer lower latency and reduced RPC usage.
```
typedarray: Websocket connections typically offer

typedarray (approved):
LGTM

Issue #1798, opened 2025-06-02
Rewrite `decodeAbiParameters` to avoid converting string to bytes


kyscott18:
### Problem / use case

In `viem` the `decodeAbiParameters` function immediately converts string data to `UInt8Array`. For Ponder (which only uses string data) this introduces extra cpu overhead. 

https://github.com/wevm/viem/blob/38525bf1d55ec3fe0569e47700c7f9e70d3c971c/src/utils/abi/decodeAbiParameters.ts#L58

### Proposed solution

We should rewrite `decodeAbiParameters` to only accept string data, and never convert it to bytes. We should also use the lazy checksum proxy to delay checksumming addresses #1795. 

The solution should pass the entire test suite in viem https://github.com/wevm/viem/blob/38525bf1d55ec3fe0569e47700c7f9e70d3c971c/src/utils/abi/decodeAbiParameters.ts#L58.

PR #1807, opened 2025-06-06
Rewrite `decodeAbiParameters` from viem

khaidarkairbek:
Rewrote `decodeAbiParameters` for string input only removing unnecessary conversions. This rewrite has passed all `decodeAbiParameters` tests. Relates to #1798


kyscott18:
@khaidarkairbek Can you add the test file?

PR #1809, opened 2025-06-09
fix: handle nonconvetional schema names. e.g. - in schema name

rori4:
I found an issue with the prune command when I tried to clean the created schemas during ongoing deployments. It appears the schema names are not wrapped with quotes when doing the `DROP SCHEMA IF EXISTS` which causes errors like `syntax error at or near "-" at character 28`

this to happen if you are using dashes in the schema names for example


kyscott18:
I don't have permission to push to your remote, closing this and opening #1810

PR #1817, opened 2025-06-11
fix: crash recovery checkpoint

khaidarkairbek:
Crash recovery checkpoint is resolved per chain via db introspection.

### DIFF
--- a/packages/core/src/sync-store/index.ts
+++ b/packages/core/src/sync-store/index.ts
@@ -38,10 +38,12 @@ import {
   type SQL,
   and,
   asc,
+  desc,
   eq,
   gte,
   inArray,
   isNull,
+  lt,
   lte,
   or,
   sql,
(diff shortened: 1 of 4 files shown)


kyscott18 (approved):

packages/core/src/sync-store/index.ts
```diff
@@ -344,6 +350,29 @@ export const createSyncStore = ({
           });
       },
     ),
+
+  getSafeCrashRecoveryBlock: async ({ chainId, timestamp }) => {
+    return database.wrap({ method: "getSafeCrashRecoveryBlock" }, async () => {
+      const record = await database.qb.sync
```
typedarray: nit but I'd do this to avoid the assertion
```ts
const rows = ...
const row = rows[0]
if (row === undefined) return undefined // optional
return row
```

Issue #1802, opened 2025-06-05
feat: Allow buildId override


valamidev:
### Problem / use case

While working with huge datasets (500 GB+ schemas), it is not ideal that Ponder forces to recreate the schema without allowing the developer to decide whether the change actually affects the indexing.

Probably running Ponder in Docker (rebuild the image) even increase the probability that BuildId is changing. 

### Proposed solution

Introduce new Env variable e.g.:

STATIC_BUILD_ID=

kyscott18:
@valamidev `buildId` is meant to be a content hash for the app. I don't think allowing it to be overridden is a good idea. Going to close this issue. Happy to help if you can provide some more details about your specific problem.

PR #1826, opened 2025-06-14
Allow BuildID override

valamidev:
Add 'PONDER_BUILD_ID_OVERRIDE' to allow override BUILD_ID.

**Reason:**
- Having full control over unexpected schema invalidation. While using Ponder I did face several scenarios where out of context code changes triggered this defense mechanism.


kyscott18:
See #1802

PR #1841, opened 2025-06-20
feat: support Harmony, Moonriver, and Aurora retry ranges

ChiTimesChi:
Adds Harmony, Moonriver, and Aurora RPC support to `getLogsRetryHelper`.

  When these RPCs return block range errors, the helper now extracts the limit and
   automatically chunks requests accordingly:
  - **Harmony**: _"query must be smaller than size X"_ (limit: 1024)
  - **Moonriver**: _"block range is too wide (maximum X)"_ (limit: 1024)
  - **Aurora**: _"up to a X block range"_ (limit: 2000)

  Includes tests to verify the new functionality works as expected for all three
  chains.

### DIFF
--- a/packages/utils/src/_test/aurora.test.ts
+++ b/packages/utils/src/_test/aurora.test.ts
@@ -1,0 +1,68 @@
+import { LimitExceededRpcError, numberToHex } from "viem";
+import { aurora } from "viem/chains";
+import { expect, test } from "vitest";
+import { getLogsRetryHelper } from "../getLogsRetryHelper.js";
+import { type Params, getRequest } from "./utils.js";
+
+const request = getRequest("https://mainnet.aurora.dev", aurora);
+const fromBlock = 57_600_000n;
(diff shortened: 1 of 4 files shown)


ChiTimesChi:
One side note here is that it's potentially possible to merge these two:

https://github.com/ponder-sh/ponder/blob/f8ce33482aa171b3ff65c6b70c5ccf64a56bbacf/packages/utils/src/getLogsRetryHelper.ts#L156-L157

https://github.com/ponder-sh/ponder/blob/f8ce33482aa171b3ff65c6b70c5ccf64a56bbacf/packages/utils/src/getLogsRetryHelper.ts#L553-L554

into a common case, getting rid of the new aurora regexp.
```typescript
// quicknode, 1rpc, blast, aurora
match = sError.match(/(?:limited|up) to a ([\d,.]+)/);
```

But not sure is that's going to simplify things - let me know.

PR #1847, opened 2025-06-24
cache child addresses in hystorical sync

khaidarkairbek:
Resolves an issue of slow getChildAddresses method when factory is large during historical backfill via caching child addresses in memory


packages/core/src/sync-historical/index.ts
```diff
@@ -64,6 +64,7 @@ import {
 
 export type HistoricalSync = {
   intervalsCache: Map<Filter, { fragment: Fragment; intervals: Interval[] }[]>;
+  childAddressesCache: Map<Factory, Map<Address, number>>;
```
kyscott18: Rename to `childAddress` to be consistent with `sync-realtime`

packages/core/src/sync-historical/index.ts
```diff
@@ -426,12 +429,28 @@ export const createHistoricalSync = async (
...
-        if (childAddresses.has(address) === false) {
+        const prevBlockNumber = childAddresses.get(address);
+
+        if (
+          prevBlockNumber === undefined ||
+          prevBlockNumber > hexToNumber(log.blockNumber)
+        ) {
           childAddresses.set(address, hexToNumber(log.blockNumber));
         }
       }
```
kyscott18: Nice catch. Is there the same bug in `sync-realtime` as well?

https://github.com/ponder-sh/ponder/blob/fcb838296903b7caf6e84e2236c5a3d752aa15f1/packages/core/src/sync-realtime/index.ts#L548-L556

packages/core/src/sync/index.ts
```diff
@@ -1274,9 +1276,11 @@ export async function* getLocalEventGenerator(params: {
...
         case "log":
           if (isAddressFactory(filter.address)) {
-            const childAddresses = await params.syncStore.getChildAddresses({
-              factory: filter.address,
-            });
+            const childAddresses =
+              params.childAddressesCache.get(filter.address) ??
+              (await params.syncStore.getChildAddresses({
+                factory: filter.address,
+              }));
```
kyscott18: When would the first expression evaluate to `undefined`? 

`sync-realtime` uses a design that pre-populates the cache. That way everywhere the cache is used it is guaranteed that the child addresses are already fetched.

packages/core/src/sync-historical/index.ts
khaidarkairbek: No, it seems like the blocks are processed in order

packages/core/src/sync/index.ts
khaidarkairbek: I resolved this by prepopulating the cache before creating historical sync, so should not be a problem now.

kyscott18 (approved):

packages/core/src/sync-historical/index.ts
```diff
@@ -422,23 +425,34 @@ export const createHistoricalSync = async (
...
     for (const log of logs) {
       if (isLogFactoryMatched({ factory, log })) {
         const address = getChildAddress({ log, factory });
-        if (childAddresses.has(address) === false) {
+        const prevBlockNumber = childAddresses.get(address);
+
+        if (
+          prevBlockNumber === undefined ||
+          prevBlockNumber > hexToNumber(log.blockNumber)
+        ) {
```
typedarray: I'd find this easier to read as 
```ts
const existingBlockNumber = childAddresses.get(address);
const newBlockNumber = hexToNumber(log.blockNumber)

if (
  existingBlockNumber === undefined ||
  newBlockNumber < existingBlockNumber
) {
  childAddresses.set(address, newBlockNumber);
}
```

packages/core/src/sync-historical/index.ts
```diff
@@ -427,29 +427,24 @@ export const createHistoricalSync = async (
...
+        const existingBlockNumber = childAddressesRecord.get(address);
 
         if (
-          prevBlockNumber === undefined ||
-          prevBlockNumber > hexToNumber(log.blockNumber)
+          existingBlockNumber === undefined ||
+          existingBlockNumber > hexToNumber(log.blockNumber)
         ) {
           childAddresses.set(address, hexToNumber(log.blockNumber));
+          childAddressesRecord.set(address, hexToNumber(log.blockNumber));
```
kyscott18: It is an invariant that `childAddressRecord`  is a superset  `childAddresses`, right?

khaidarkairbek: Yes, you are right

PR #1860, opened 2025-06-30
chore: telemetry cleanup

typedarray:


docs/api/telemetry.ts
```diff
@@ -59,12 +27,8 @@ export default async function forwardTelemetry(
     return res.status(500).json({ error: "Server error" });
   };
 
-  // If the event has an distinctId, it's a new Posthog event from >= 0.4.3
+  // If the event has an distinctId, it's a new event from >= 0.4.3
   if (body.distinctId) {
```
typedarray, line 31: I'd consider an early return here with an explicit response code (probably a 5xx?)

Issue #1834, opened 2025-06-17
sync_store getEventBlockData() stuck processing


valamidev:
### Version

latest

### Current behavior

Running the `examples/reference-erc20` with the following config:

```
import { createConfig } from "ponder";
import { erc20ABI } from "./abis/erc20ABI";

export default createConfig({
  chains: {
    base: {
      id: 8543,
      rpc: process.env.PONDER_RPC_URL_8453,
    },
  },
  contracts: {
    ERC20: {
      chain: "base",
      abi: erc20ABI,
      startBlock: 31432324,
    },
  },
});

```

end up in a infinite loop of:

```
2:25:01 PM DEBUG sync       Extracted 0 'base' events for block range [31432325, 31432324]
2:25:01 PM DEBUG sync       Extracted 0 'base' events for block range [31432325, 31432324]
2:25:01 PM DEBUG sync       Extracted 0 'base' events for block range [31432325, 31432324]
2:25:01 PM DEBUG sync       Extracted 0 'base' events for block range [31432325, 31432324]
2:25:02 PM DEBUG sync       Extracted 0 'base' events for block range [31432325, 31432324]
2:25:02 PM DEBUG sync       Extracted 0 'base' events for block range [31432325, 31432324]
2:25:02 PM DEBUG sync       Extracted 0 'base' events for block range [31432325, 31432324]
2:25:02 PM DEBUG sync       Extracted 0 'base' events for block range [31432325, 31432324]
```

PR #1863, opened 2025-07-01
Fix getEventBlockData deadlock

khaidarkairbek:
Fixes the getEventBlockData pagination deadlock by resolving the last incomplete block with additional request instead of abandoning it. Thus, allowing to proceed even with large number of events per block. Relates to #1834.


packages/core/src/sync-store/index.ts
```diff
@@ -134,6 +135,285 @@ export type SyncStore = {
...
+    )
+    .limit(limit);
+};
+
+const getEventBlockData_traces = async (
+  args: {
+    traceFilters: TraceFilter[];
+    transferFilters: TransferFilter[];
+    fromBlock: number;
+    fromTraceIndex: number | undefined;
```
kyscott18: Traces should also be paginated on `transactionIndex`

packages/core/src/sync-store/index.ts
```diff
@@ -845,7 +967,144 @@ export const createSyncStore = ({
...
+            }
+          }
+        };
+
+        await Promise.all([
+          transactionsPromise,
+          transactionReceiptsPromise,
+          logsPromise,
+          tracesPromise,
+        ]);
```
kyscott18: So these basically pull all the extra data from end of the limit to the supremum block?

khaidarkairbek: yes

packages/core/src/sync-store/index.ts
khaidarkairbek: so the traceIndex is not unique per trace in block?

kyscott18: no it's unique per transaction

khaidarkairbek: ok, sounds good

packages/core/src/sync-store/index.ts
```diff
@@ -566,502 +566,513 @@ export const createSyncStore = ({
...
+        // all events for `supremum` block have been extracted
+        cursor = supremum;
+      } else {
+        // there may be events for `supremum` block that have not been extracted
+        blockData.pop();
+        cursor = supremum - 1;
+
+        if (cursor < fromBlock) {
+          limit = limit * 2;
+          return await fn();
```
kyscott18, line 1064: Should apply `wrap()` here

khaidarkairbek: Then it would be nested wrapped which would mess up the retrial, no?

kyscott18: Ah I see, this code is already inside `fn`

typedarray (approved):

PR #1867, opened 2025-07-02
add `setDatabaseSchema` to `@ponder/client`

kyscott18:


examples/with-offchain/offchain/src/schema.ts
```diff
@@ -1,27 +1,16 @@
...
+import { setDatabaseSchema } from "@ponder/client";
+import { relations } from "drizzle-orm";
+import * as ponderSchema from "../../ponder/ponder.schema";
 import * as offchainSchema from "./offchain";
 
 // Note: We need a separate file for merging the schemas because
 // "ponder.schema" can't be executed by drizzle-kit, and we also
 // don't want drizzle to generate migrations for onchain tables.
 
 // Note: `_ponderSchema` doesn't have information about which database schema
```
typedarray: No underscore

packages/client/src/index.ts
```diff
@@ -218,3 +223,18 @@ export {
   except,
   exceptAll,
 } from "drizzle-orm/pg-core";
+
+export const setDatabaseSchema = <T extends { [name: string]: unknown }>(
```
typedarray, line 227: Was it an explicit design decision to avoid copying the schema, or the only way to get it working?

Slightly worried about mutating - won't it affect any other consumers of the schema?

typedarray (approved):

PR #1868, opened 2025-07-02
fix: childAddresses by factory id

khaidarkairbek:


packages/core/src/sync-historical/index.ts
```diff
@@ -75,7 +76,7 @@ type CreateHistoricalSyncParameters = {
   common: Common;
   sources: Source[];
   syncStore: SyncStore;
-  childAddresses: Map<Factory, Map<Address, number>>;
+  childAddresses: Map<string, Map<Address, number>>;
```
typedarray: What do you think about adding a `id: string` property to this type, rather than recomputing the ID on the fly? https://github.com/ponder-sh/ponder/blob/main/packages/core/src/internal/types.ts#L172-L180

khaidarkairbek: Yeah, that makes sense. I am also thinking about the edge case where potentially given that we have two identical factories, we still be doing double the job and keeping both copies of childAddresses in database.

khaidarkairbek: But this seems to be out of scope for now.

packages/core/src/sync-historical/index.test.ts
```diff
@@ -46,22 +47,28 @@ beforeEach(setupCleanup);
 
 const setupChildAddresses = (
   sources: Source[],
-): Map<Factory, Map<Address, number>> => {
+): Map<string, Map<Address, number>> => {
```
typedarray: Another idea to consider - you could export a type
```
export type FactoryId = string;
```
from the `factory.ts` file, and use it instead of `string` here to make it easier to understand at a glance. Lightweight version of [branded types](https://egghead.io/blog/using-branded-types-in-typescript). Not a strong opinion though.

.changeset/fruity-mails-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Introduced factoryId which is used for child addresses tracking
```
kyscott18, line 5: The changeset should only include information relevant to the users. If doing a completely internal refactor, you don't have to include one. Or you could say something like "Improved factory reliability".

packages/core/src/build/factory.ts
```diff
@@ -34,6 +34,7 @@ export function buildLogFactory({
 
   if (indexedInputPosition > -1) {
     return {
+      id: `log_${Array.isArray(address) ? address.join("_") : address}_${chainId}_topic${(indexedInputPosition + 1) as 1 | 2 | 3}_${eventSelector}_${fromBlock ?? "undefined"}_${toBlock ?? "undefined"}`,
```
kyscott18, line 37: Another option is to just `JSON.stringify` the factory properties. Up to you

packages/core/src/sync-realtime/bloom.test.ts
```diff
@@ -139,6 +140,7 @@ test("isFilterInBloom returns true for factory without new child address", () =>
     type: "log",
     chainId: 1,
     address: {
+      id: `log_${"0xef2d6d194084c2de36e0dabfce45d046b37d1106"}_${1}_topic${1}_${"0x02c69be41d0b7e40352fc85be1cd65eb03d40ef8427a0ca4596b1ead9a00e9fc"}_${"undefined"}_${"undefined"}`,
```
kyscott18, line 143: If there is a bunch of text in your test, you could do `expect().toMatchInlineSnapshot` and vitest will fill it out for you after you run it the first time

.changeset/fruity-mails-visit.md
khaidarkairbek: Okay, sounds good.

packages/core/src/sync-realtime/bloom.test.ts
khaidarkairbek: noted

PR #1869, opened 2025-07-02
feat: include primary keys in update

khaidarkairbek:
Currently, we throw an error for an update args with primary keys. This solution tolerates including primary keys as log as they are identical to the previous row. Otherwise, it throws an error as before.

### DIFF
--- a/packages/core/src/indexing-store/index.ts
+++ b/packages/core/src/indexing-store/index.ts
@@ -14,6 +14,7 @@ import type { Schema } from "@/internal/types.js";
 import type { Db } from "@/types/db.js";
 import type { Table } from "drizzle-orm";
 import { getTableConfig } from "drizzle-orm/pg-core";
+import type { Row } from "./cache.js";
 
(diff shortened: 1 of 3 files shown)


packages/core/src/indexing-store/index.ts
```diff
@@ -38,17 +39,28 @@ export const parseSqlError = (e: any): Error => {
...
 
   for (const { js } of primaryKeys) {
     if (js in set) {
-      throw new NonRetryableError(
-        `Primary key column '${js}' cannot be updated`,
-      );
+      // Note: Strip off the primary keys if they are identical, otherwise throw an error.
+      if (prev !== null && newSet[js] === prev[js]) {
+        delete newSet[js];
+      } else {
```
kyscott18: Is it necessary to delete from `newSet` when the pks are equal? It would save memory if we didn't do this.

kyscott18 (approved):

PR #1881, opened 2025-07-09
Fix: reorg revert on multichain

khaidarkairbek:
Fixed the reorg behavior for multichain ordering both for databse and sync.

### DIFF
--- a/packages/core/src/database/index.ts
+++ b/packages/core/src/database/index.ts
@@ -78,7 +78,10 @@ export type Database = {
    */
   migrate({
     buildId,
-  }: Pick<IndexingBuild, "buildId">): Promise<CrashRecoveryCheckpoint>;
+    ordering,
+  }: Pick<IndexingBuild, "buildId"> & {
+    ordering: "omnichain" | "multichain";
+  }): Promise<CrashRecoveryCheckpoint>;
   createIndexes(): Promise<void>;
   createTriggers(): Promise<void>;
   removeTriggers(): Promise<void>;
@@ -112,6 +115,7 @@ export type Database = {
   getReady(): Promise<boolean>;
   revert(args: {
     checkpoint: string;
+    ordering: "multichain" | "omnichain";
     tx: PgTransaction<PgQueryResultHKT, Schema>;
   }): Promise<void>;
   finalize(args: { checkpoint: string; db: Drizzle<Schema> }): Promise<void>;
@@ -675,7 +679,7 @@ export const createDatabase = async ({
         },
       );
     },
-    async migrate({ buildId }) {
+    async migrate({ buildId, ordering }) {
       await this.wrap(
         { method: "createPonderSystemTables", includeTraceLogs: true },
         async () => {
@@ -930,7 +934,7 @@ EXECUTE PROCEDURE "${namespace.schema}".${notification};`),
               ...checkpoints.map((c) => c.safeCheckpoint),
             );
 
-            await this.revert({ checkpoint: revertCheckpoint, tx });
+            await this.revert({ checkpoint: revertCheckpoint, ordering, tx });
 
             // Note: We don't update the `_ponder_checkpoint` table here, instead we wait for it to be updated
             // in the runtime script.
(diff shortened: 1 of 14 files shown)


packages/core/src/database/index.ts
```diff
@@ -1080,21 +1109,65 @@ FOR EACH ROW EXECUTE FUNCTION "${namespace.schema}".${getTableNames(table).trigg
...
+  SELECT MIN(operation_id) AS min_op_id FROM "${namespace.schema}"."${getTableName(getReorgTable(table))}"
+  WHERE SUBSTRING(checkpoint, 11, 16)::numeric = ${String(decodeCheckpoint(checkpoint).chainId)}
+  AND checkpoint > '${checkpoint}'
+`,
+  )
+  .join(
+    `
+  UNION ALL
+`,
+  )}) AS all_mins
```
kyscott18: what does `all_mins` do?

packages/core/src/database/index.ts
```diff
@@ -1080,21 +1109,65 @@ FOR EACH ROW EXECUTE FUNCTION "${namespace.schema}".${getTableNames(table).trigg
...
-          tables.map(async (table) => {
-            const primaryKeyColumns = getPrimaryKeyColumns(table);
-
-            const result = await tx.execute(
-              sql.raw(`
+      await this.record(
+        { method: "revert", includeTraceLogs: true },
+        async () => {
+          const min_op_id: number | null =
+            this.ordering === "omnichain"
```
kyscott18: Could we use the same algorithm for both orderings?

packages/core/src/database/index.ts
```diff
@@ -1080,21 +1109,65 @@ FOR EACH ROW EXECUTE FUNCTION "${namespace.schema}".${getTableNames(table).trigg
...
+                  )
+                  .then((result) => {
+                    // @ts-ignore
+                    if (!result.rows) {
+                      return null;
+                    }
+
+                    // @ts-ignore
+                    return result.rows[0]!.global_min_op_id;
+                  });
```
kyscott18: Prefer `undefined` here then could do a one-liner `result.rows[0]?.global_min_op_id`

packages/core/src/database/index.ts
```diff
@@ -1080,21 +1109,65 @@ FOR EACH ROW EXECUTE FUNCTION "${namespace.schema}".${getTableNames(table).trigg
         return qb.drizzle
           .select()
           .from(PONDER_META)
-          .then((result) => result[0]?.value.is_ready === 1 ?? false);
+          .then((result) => result[0]?.value.is_ready === 1);
```
kyscott18, line 1096: nice

packages/core/src/drizzle/kit/index.ts
```diff
@@ -122,6 +128,10 @@ export const getSql = (schema: { [name: string]: unknown }): SqlStatements => {
       sql: fromJson(jsonCreateIndexesForCreatedTables),
       json: jsonCreateIndexesForCreatedTables,
     },
+    sequences: {
+      sql: [`CREATE SEQUENCE ${SHARED_OPERATION_ID_SEQUENCE};\n`],
+      json: [prepareCreateSequenceJson(SHARED_OPERATION_ID_SEQUENCE)],
+    },
```
kyscott18: I think you should remove all the sequence logic from this file except for setting the `operation_id` default. In `database/index.ts`, create the sequence with a raw sql statement the same way the meta and checkpoint tables are created.

packages/core/src/sync/index.ts
```diff
@@ -702,48 +702,84 @@ export const createSync = async (params: {
...
+        let _finalizedEvents = 0;
+
         // Remove all finalized data
+        let finalizeIndex: number | undefined = undefined;
+        for (const [index, event] of executedEvents.entries()) {
+          if (event.checkpoint > to) {
+            finalizeIndex = index;
+            break;
+          }
+        }
```
kyscott18: Kinda surprised that `chainId` isn't used in this algorithm at all.

kyscott18 (commented):
Few questions

packages/core/src/database/index.ts
khaidarkairbek: just an alias to keep sql happy

packages/core/src/database/index.ts
khaidarkairbek: No, the behavior can't be generalized. A bit tricky since the checkpoint passed here is the min of all checkpoints, and not the checkpoint caused by the reorg event on that chain. Thus, leaving any event after checkpoint might lead to breaking order in omnichain case.

packages/core/src/sync/index.ts
khaidarkairbek: No need for chainId since the finalize checkpoint is the min across all chains.

packages/core/src/database/index.ts
```diff
@@ -146,7 +150,7 @@ export type PonderApp = {
   heartbeat_at: number;
 };
 
-const VERSION = "2";
+const VERSION = "3";
```
kyscott18, line 153: @typedarray should we make this a patch or minor version

typedarray: patch

PR #1883, opened 2025-07-10
improve database error handling

kyscott18:


packages/core/src/bin/commands/createViews.ts
```diff
@@ -138,8 +140,9 @@ export async function createViews({
   const notification = "status_notify()";
   const channel = `${cliOptions.viewsSchema}_status_channel`;
 
-  await database.qb.drizzle.execute(
-    sql.raw(`
```
typedarray, line 152: Why can we get rid of sql.raw here?

packages/core/src/bin/utils/run.ts
```diff
@@ -1,16 +1,29 @@
...
+  getPonderMetaTable,
+} from "@/database/index.js";
+import {
+  commitBlock,
+  createIndexes,
+  createTrigger,
+  dropTrigger,
+  finalize,
+  revert,
+} from "@/database/utils.js";
```
typedarray: Maybe @/database/actions.js

packages/core/src/bin/utils/run.ts
```diff
@@ -129,45 +164,52 @@ export async function run({
...
+      indexingCache.qb = tx;
 
-        if (result.status === "error") {
+      const result = await indexing.processSetupEvents({
+        db: historicalIndexingStore,
+      });
+
+      if (result.status === "error") {
+        if (result.error instanceof RetryableError) {
+          throw result.error;
```
typedarray, line 177: So, this error will be caught and retried by  userQB.transaction?

packages/core/src/bin/utils/run.ts
```diff
@@ -129,45 +164,52 @@ export async function run({
...
+        } else {
           onReloadableError(result.error);
```
typedarray, line 179: Should we change this name?

packages/core/src/bin/utils/run.ts
```diff
@@ -129,45 +164,52 @@ export async function run({
...
-            onReloadableError(error as Error);
-            return;
-          }
-          throw error;
+      try {
+        await indexingCache.flush();
+      } catch (error) {
+        if (error instanceof DelayedInsertError) {
+          onReloadableError(error as Error);
+          return;
```
typedarray, line 189: This return seems a bit unsafe - there's no mechanism to ensure the current `run` function doesn't continue executing.

The call to `run` in `dev` looks like this:
```ts
run({
  // ...
  onFatalError: () => {
    exit({ reason: "Received fatal error", code: 1 });
  },
  onReloadableError: (error) => {
    buildQueue.clear();
    buildQueue.add({ status: "error", kind: "indexing", error });
  },
});
```

I wonder if we should replace these error callbacks with a `.catch` block that handles different error types:
```ts
run({ ... })
  .catch((error) => {
    if (error instanceof FatalError) {
      exit({ reason: "Received fatal error", code: 1 });
    } else if (error instanceof ReloadableError) {
      buildQueue.clear();
      buildQueue.add({ status: "error", kind: "indexing", error });
    } else {
      // internal error? exit? 
    }
  })
```
This way, we could just throw freely inside the run function.

packages/core/src/bin/utils/run.ts
```diff
@@ -129,45 +164,52 @@ export async function run({
...
-    // Note: `_ponder_checkpoint` must be updated after the setup events are processed.
-    await database.setCheckpoints({
-      checkpoints: indexingBuild.chains.map((chain) => ({
-        chainName: chain.name,
-        chainId: chain.id,
-        latestCheckpoint: sync.getStartCheckpoint(chain),
-        safeCheckpoint: sync.getStartCheckpoint(chain),
-      })),
-      db: database.qb.drizzle,
+      await tx.wrap({ label: "test" }, (tx) =>
```
typedarray: Looks like it needs a new label. Also, any reason not to make this a database util/action?

packages/core/src/bin/utils/run.ts
```diff
@@ -400,193 +456,263 @@ export async function run({
...
+                  });
+
+                  throw error;
+                }
+              });
             }
           }
+
+          if (event.checkpoints.length > 0) {
+            await database.userQB.wrap({ label: "update_checkpoints" }, (db) =>
```
typedarray, line 636: Should this be in the same transaction as the user queries and commitBlock?

packages/core/src/build/index.ts
```diff
@@ -377,6 +379,55 @@ export const createBuild = async ({
...
+            service: "build",
+            msg: "Failed build",
+            error,
+          });
+          return { status: "error", error };
+        } finally {
+          await driver.close();
+        }
+      } else if (dialect === "postgres") {
+        const pool = createPool(
```
typedarray, line 403: Could just create a single connection here instead of a pool

packages/core/src/database/utils.ts
```diff
@@ -0,0 +1,237 @@
...
+  for (const statement of statements.indexes.sql) {
+    await qb.transaction({ label: "create_indexes" }, async (tx) => {
+      // 60 minutes
+      await tx.wrap((tx) => tx.execute("SET statement_timeout = 3600000;"));
+      await tx.wrap((tx) => tx.execute(statement));
+    });
+  }
+};
+
+export const createTrigger = async (qb: QB, { table }: { table: PgTable }) => {
```
typedarray: Do we ever create/remove triggers for a subset of tables?

kyscott18: No

packages/core/src/bin/utils/run.ts
kyscott18: yes

packages/core/src/bin/commands/createViews.ts
kyscott18: drizzle implicitly converts strings to sql.raw

PR #1893, opened 2025-07-16
add retry logic to `context.client.getBlock()`

kyscott18:


.changeset/salty-pianos-sell.md
```diff
@@ -0,0 +1,7 @@
+---
+"ponder": patch
+---
+
+Improved retry behavior for `context.client.getBlock()`, `context.client.getTransaction()`, `context.client.getTransactionReceipt()`, and `context.client.getTransactionConfirmations()`. This fixes the error: `BlockNotFoundError: Block at number "[number]" could not be found`.
```
typedarray: Last sentence could be more specific e.g.

Now, the error `BlockNotFoundError: Block at number "[number]" could not be found` will be retried properly. This solves a common issue where RPCs return empty data for near-tip blocks.

typedarray (approved):

PR #1894, opened 2025-07-16
fix big chain ids

kyscott18:


.changeset/lucky-boxes-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Add support for chain IDs larger than Postgres max integer (2147483647).
```
typedarray: Added support for chain IDs between 2147483647 (Postgres max `integer`) and 9007199254740991 (JavaScript Number.MAX_SAFE_INTEGER).

typedarray: I think we might as well add a callout note to the Chains page on the docs about this.

typedarray (approved):

PR #1900, opened 2025-07-18
fix: ordering and chain aware app progress

khaidarkairbek:


packages/core/src/internal/metrics.ts
```diff
@@ -597,67 +575,183 @@ export async function getIndexingProgress(metrics: MetricsService) {
...
+              ? undefined
+              : Math.max(prev.eta, curr.eta),
+      }),
+      {
+        mode: "realtime",
+        progress: 1,
+        eta: 0,
+      },
+    ) as any;
+  } else {
```
typedarray, line 677: So, this `else` also handles the case where ordering is undefined?

packages/core/src/internal/metrics.ts
```diff
@@ -597,67 +575,183 @@ export async function getIndexingProgress(metrics: MetricsService) {
...
+      });
+    }
+
+    return perChainAppProgress.reduce(
+      (prev, curr) => ({
+        mode: curr.mode === "historical" ? curr.mode : prev.mode,
+        progress:
+          prev.progress === undefined || curr.progress === undefined
+            ? undefined
+            : Math.min(prev.progress, curr.progress),
```
typedarray, line 663: So, this is where we apply the "minimum progress % across all chains" logic?

packages/core/src/internal/metrics.ts
```diff
@@ -597,67 +575,183 @@ export async function getIndexingProgress(metrics: MetricsService) {
...
+        ? 0
+        : totalSeconds === 0
+          ? 1
+          : (completedSeconds + cachedSeconds) / totalSeconds;
+
+    return {
+      mode: progress === 1 ? "realtime" : "historical",
+      progress: progress,
+      eta: calculateEta(
+        metrics.progressMetadata.general!,
```
typedarray, line 702: What if `chainName === "general"` ? 😄

packages/core/src/internal/metrics.ts
khaidarkairbek: Added the handler.

packages/core/src/internal/metrics.ts
khaidarkairbek: yes, thats correct.

packages/core/src/internal/metrics.ts
khaidarkairbek: tbh, should not be a problem. since if its multichain, you wouldn't use general (its for omnichain use) and it is vacant for such chain. If its omnichain, you wouldn;t even care about chain name, because we are not tracking chain specific metrics.

kyscott18 (approved):

PR #1914, opened 2025-07-25
feat: isolated extension of db

khaidarkairbek:


packages/core/src/database/index.ts
```diff
@@ -921,12 +941,38 @@ EXECUTE PROCEDURE "${namespace.schema}".${notification};`),
...
+                  min(...checkpoints.map((c) => c.safeCheckpoint)),
+                );
+                for (const { chainId } of checkpoints) {
+                  // Replace chainId such that revert correctly cleans up reorg table
+                  const revertCheckpoint = encodeCheckpoint({
+                    ...revertCheckpoint_,
+                    chainId: BigInt(chainId),
+                  });
+                  await this.revert({ checkpoint: revertCheckpoint, tx });
+                }
```
kyscott18: This fixes the issue we talked about last week with crash recovery, right?

kyscott18 (approved):
Looks great. 

I think that `ordering` should be passed as a parameter to any relevant functions, rather than as a property of `database`.

packages/core/src/database/index.ts
khaidarkairbek: Yes, this aims to do that.

PR #1949, opened 2025-08-06
Crash recovery fix

khaidarkairbek:
Improve finalization logic to be more reliable on crash recoveries. 

@kyscott18 I am a bit unconfident about the way I handle checkpoint table state. (finalize assumes that checkpoint table is filled for every chain id)


packages/core/src/database/actions.ts
```diff
@@ -197,8 +198,25 @@ WHERE checkpoint > '${checkpoint}' RETURNING *
 export const finalize = async (
   qb: QB,
   { checkpoint, tables }: { checkpoint: string; tables: PgTable[] },
```
kyscott18: Do we not need a separate "omnichain" path?

packages/core/src/database/index.ts
```diff
@@ -713,11 +716,30 @@ EXECUTE PROCEDURE "${namespace.schema}".${notification};`,
...
-            ...checkpoints.map((c) => c.safeCheckpoint),
-          );
+          switch (ordering) {
+            case "omnichain": {
+              const revertCheckpoint = min(
+                ...checkpoints.map((c) => c.safeCheckpoint),
+              );
 
-          await revert(tx, { checkpoint: revertCheckpoint, tables, ordering });
+              await revert(tx, {
```
kyscott18: Would rather have a new `crashRecovery` action that just reverts every row from the reorg table. I could see the multiple reverts in multichain ordering causing us trouble.

packages/core/src/bin/utils/run.ts
```diff
@@ -176,12 +181,20 @@ export async function run({
...
               latestCheckpoint: sync.getStartCheckpoint(chain),
-              safeCheckpoint: sync.getStartCheckpoint(chain),
+              finalizedCheckpoint: encodeCheckpoint({
+                ...ZERO_CHECKPOINT,
+                chainId: BigInt(chain.id),
+              }),
+              safeCheckpoint: encodeCheckpoint({
+                ...ZERO_CHECKPOINT,
+                chainId: BigInt(chain.id),
+              }),
```
kyscott18: I think we should also set these to `startCheckpoint`

packages/core/src/graphql/index.test.ts
```diff
@@ -48,6 +48,14 @@ test("metadata", async (context) => {
...
         eventIndex: 0n,
       }),
+      finalizedCheckpoint: encodeCheckpoint({
+        blockNumber: 10n,
+        chainId: 1n,
+        blockTimestamp: 20n,
+        transactionIndex: 0n,
+        eventType: EVENT_TYPES.blocks,
+        eventIndex: 0n,
+      }),
```
kyscott18, line 58: What happened here?

kyscott18 (commented):
Looking great, I'm surprised there are not changes in `sync/index.ts` Thought we would need the same changes that we added to the database.

packages/core/src/graphql/index.test.ts
khaidarkairbek: I just needed finalizedCheckpoint field, decided to use same as safeCheckpoint

khaidarkairbek:
> Looking great, I'm surprised there are not changes in `sync/index.ts` Thought we would need the same changes that we added to the database.

You are rightfully surprised, I totally forgot about sync

packages/core/src/bin/utils/run.ts
khaidarkairbek: what if startCheckpoint > finalizedCheckpoint?

kyscott18 (approved):

PR #1954, opened 2025-08-07
[wip]: added multi-threading to isolated mode

khaidarkairbek:


packages/core/src/runtime/isolated.ts
```diff
@@ -47,28 +60,99 @@ import {
 } from "./index.js";
 import { getRealtimeEventsIsolated } from "./realtime.js";
 
-export async function runIsolated(
-  {
+const isWorker = isMainThread === false;
+
+if (isWorker) {
+  if (parentPort) {
```
kyscott18, line 66: Do you know when `parentPort` would be null?

kyscott18 (approved):

packages/core/src/runtime/isolated.ts
khaidarkairbek: parentPort is null if its main thread

khaidarkairbek: So I'm pretty sure its a big if true in this case. Can add logs later just in case.

PR #1968, opened 2025-08-13
fast catchup

kyscott18:


packages/core/src/runtime/historical.ts
```diff
@@ -60,147 +61,222 @@ export async function* getHistoricalEventsOmnichain(params: {
...
-        const fromBlock = await params.syncStore.getSafeCrashRecoveryBlock({
-          chainId: chain.id,
-          timestamp: Number(
-            decodeCheckpoint(crashRecoveryCheckpoint).blockTimestamp,
-          ),
-        });
+          yield {
+            events: pendingEvents.filter((event) => event.chainId === chain.id),
+            checkpoint: cursor,
+          };
```
khaidarkairbek: There might be an edge case here. There is theoretically some pending events that still cannot be processed. Should be something like this: 
```ts
const [left, right] = partition(
  pendingEvents,
  (event) => event.checkpoint <= omnichainTo && event.chainId === chain.id,
);

pendingEvents = right; 

yield {
  events: left,
  checkpoint: cursor,
};
```

khaidarkairbek (changes requested):
Looks good, just a small edge case to be aware of.

khaidarkairbek (approved):
looks good

PR #1969, opened 2025-08-13
[wip]: dev mode and ui for isolated ordering

khaidarkairbek:


packages/core/src/bin/commands/dev.ts
```diff
@@ -58,10 +58,19 @@ export async function dev({ cliOptions }: { cliOptions: CliOptions }) {
...
-  const common = { options, logger, metrics, telemetry };
+
+  const common = {
+    options,
+    logger,
+    metrics,
+    telemetry,
+    shutdown,
+    indexingShutdown,
+    apiShutdown,
```
kyscott18: What does this do?

packages/core/src/server/index.ts
```diff
@@ -124,8 +108,13 @@ export async function createServer({
       );
       const status: Status = {};
       for (const { chainName, chainId, latestCheckpoint } of checkpoints) {
+        const state =
+          common.metrics.app === undefined
+            ? undefined
+            : common.metrics.app[chainName]?.state;
         status[chainName] = {
           id: chainId,
+          state,
```
kyscott18, line 117: What happens in `ponder serve`?

kyscott18 (commented):
Looking good, one question I have is around the metrics. I understand that we now have a new concept: the main thread and worker thread relationship. When in isolated mode, the worker thread each get their own metrics instance and the main thread also get a metrics instance. The main thread is special because it doesn't have any state of its own, it just aggregates the metrics from the workers. 

If this description is correct, do you think we should create different classes or anything like that to make the distinction more clear?

packages/core/src/server/index.ts
khaidarkairbek: in ponder serve metrics is not aggregated so no state there

packages/core/src/bin/commands/dev.ts
khaidarkairbek: typo

khaidarkairbek:
> Looking good, one question I have is around the metrics. I understand that we now have a new concept: the main thread and worker thread relationship. When in isolated mode, the worker thread each get their own metrics instance and the main thread also get a metrics instance. The main thread is special because it doesn't have any state of its own, it just aggregates the metrics from the workers.
> 
> If this description is correct, do you think we should create different classes or anything like that to make the distinction more clear?

Tried couple approaches, seems like a lot of pain.

PR #1971, opened 2025-08-15
[wip]: isolated ordering

khaidarkairbek:
This pr introduces additional ordering called "isolated" that allows to run ponder instances on multiple chains with complete isolation and leveraging worker_threads for superior performance.

### DIFF
--- a/packages/core/src/bin/commands/isolatedController.ts
+++ b/packages/core/src/bin/commands/isolatedController.ts
@@ -117,6 +117,9 @@ export async function isolatedController({
           namespaceBuild,
           crashRecoveryCheckpoint,
         } satisfies Parameters<typeof isolatedWorker>[0],
+        resourceLimits: {
+          maxOldGenerationSizeMb: 1024,
+        },
       }),
       pwr,
     };


packages/core/src/database/index.ts
```diff
@@ -54,12 +55,14 @@ export type Database = {
...
-    IndexingBuild,
-    "buildId" | "chains" | "finalizedBlocks"
-  >): Promise<CrashRecoveryCheckpoint>;
+    ordering,
+  }: Pick<IndexingBuild, "buildId" | "chains" | "finalizedBlocks"> & {
+    ordering: Ordering;
+  }): Promise<CrashRecoveryCheckpoint>;
 };
 
+export type DatabaseInterface = Omit<Database, "migrateSync" | "migrate">;
```
kyscott18: A nice simplification would be to make the `Database` type not include `migrate` or `migrateSync` and instead implement those methods as actions.

packages/core/src/build/schema.ts
```diff
@@ -81,6 +86,27 @@ export const buildSchema = ({ schema }: { schema: Schema }) => {
             );
           }
         }
+
+        if (columnName === "chainId") {
```
kyscott18, line 90: How can we make this more robust, should we compare the js name or the sql name. Also consider different casings.

packages/core/src/drizzle/index.ts
```diff
@@ -11,8 +11,16 @@ export const getTableNames = (table: PgTable) => {
...
+    trigger(chainId?: number) {
+      return chainId === undefined
+        ? sqlToReorgTableName(sql)
+        : `_${chainId}_${sqlToReorgTableName(sql)}`;
+    },
+    triggerFn(chainId?: number) {
+      return chainId === undefined
+        ? `operation_reorg__${sql}()`
+        : `_${chainId}_operation_reorg__${sql}()`;
+    },
```
kyscott18: Do any of our `ponder db` commands get affected by this?

khaidarkairbek: I dont think so, mostly affect createTriggers/dropTriggers

packages/core/src/database/index.ts
khaidarkairbek: I decided to remove createDatabaseInterface since its basically doing same thing as createDatabase. Added minimal tweaks to createDatabase

packages/core/src/build/schema.ts
khaidarkairbek: I assumed getTableColumns returns js names so comparing to "chainId" makes sense here. For the casing, I could just add camel_case condition.

PR #1985, opened 2025-08-20
[feat] add manual string concat profiling

khaidarkairbek:


packages/core/src/indexing-store/profile.ts
```diff
@@ -19,6 +19,8 @@ const eq = (target: bigint | string | number | boolean, value: any) => {
   return false;
 };
 
+const delimiters = ["-", "_", "#", "$"];
```
goums: I remembered I used this pattern on an old indexer:
```
  const oracleId = `${chainId}:${address}`;
  let oracle = await Oracle.findUnique({ id: oracleId });
```

Might be worth adding `:` to the `delimiters`

kyscott18 (approved):

PR #1988, opened 2025-08-20
improve error message for unawaited promise

kyscott18:


packages/core/src/indexing-store/historical.ts
```diff
@@ -31,11 +32,24 @@ export const createHistoricalIndexingStore = ({
...
+  let isProcessingEvents = true;
 
-  const errorHandler = (fn: (...args: any[]) => Promise<any>) => {
+  const storeMethodWrapper = (fn: (...args: any[]) => Promise<any>) => {
     return async (...args: any[]) => {
       try {
-        return await fn(...args);
+        if (isProcessingEvents === false) {
+          throw new NonRetryableUserError(
+            "Indexing store query executed outside of the indexing function. Make sure all 'context.db' queries are properly awaited.",
```
typedarray: A store API method (find, update, insert, delete) was called after the indexing function returned. Hint: Did you forget to await the store API method call (an unawaited promise)?

Issue #1974, opened 2025-08-18
Add support for filtering timestamp fields as date types in GraphQL API


nboostani:
Currently, when filtering timestamp data types via the GraphQL API, the timestamp values are returned as strings. This makes it impossible to use filters such as 'is greater than' or 'is less than' on timestamp fields directly. This is especially problematic when trying to filter events that have expired, as I need to exclude those from my site.

Request:
- Enable filtering on timestamp fields so that 'is greater than' and 'is less than' can be used.
- This would make it much easier to filter out expired events and improve data handling via GraphQL.

Thank you for considering this feature!

typedarray:
The GraphQL type & available filter methods are derived from the Drizzle column type, e.g. `t.date()` or `t.integer()`.

Could you please clarify what column type you're using here? Ideally paste a the entire table definition from your `ponder.schema.ts`.

nboostani:
> The GraphQL type & available filter methods are derived from the Drizzle column type, e.g. `t.date()` or `t.integer()`.
> 
> Could you please clarify what column type you're using here? Ideally paste a the entire table definition from your `ponder.schema.ts`.

Thank you for your response!  
Here’s the entire table definition from my `ponder.schema.ts`:

```typescript
export const predictions = onchainTable("prediction", (t) => ({
  id: t.hex().primaryKey(),
  metadata: t.text(),
  name: t.text(),
  description: t.text(),
  outcome1: t.text(),
  outcome2: t.text(),
  optionalReward: t.bigint(),
  b: t.bigint(),
  url: t.text(),
  createdAt: t.timestamp(),
  status: t.text().default("created"),
  deadline: t.timestamp(),
  tags: t.text().array(),
  hidden: t.boolean().default(false),
}));
```

For the columns `createdAt` and `deadline`, I'm using the `timestamp` datatype (`t.timestamp()`).  
This should correspond to the appropriate GraphQL type and filter methods derived from Drizzle's `t.timestamp()` column type.

Let me know if you need any more details or if you have specific requirements for how timestamps should be handled!

PR #1993, opened 2025-08-24
Add Timestamp scalar type to GraphQL schema

nboostani:
Introduce a new Timestamp scalar type to the GraphQL schema, allowing for better handling of date and time values. Update relevant areas to utilize this new type. #1974


kyscott18:
We're not planning to add this, would recommend storing the unix timestamp as an integer instead.

Issue #1803, opened 2025-06-05
contract:setup event crash after reach idle-in-transaction timeout limit


valamidev:
### Version

'latest'

### Current behavior

While my `setup` event took more than 1 hour, the `idle-in-transaction timeout` crash the application.

### Expected behavior

Setup events either using much bigger limit or be customisable.

It is not uncommon, that bootstrapping take very long time(6-24hours), still much faster than sync from genesis, not because a huge amount of data which is bootstrapped just RPC calls are really slow event with multicall.

iHiteshAgrawal:
I’m running into the same problem while indexing Uniswap V3. Our setup events sometimes take over 1 hour, and the PostgreSQL connection gets terminated due to the idle-in-transaction timeout.

It would be great if this timeout were either configurable per session or automatically handled for long-running setup events.

typedarray:
Could you share more details about what logic you're running in the setup event? We can work on increasing the timeout, but we'd like to understand the scenario a little better.

iHiteshAgrawal:
> Could you share more details about what logic you're running in the setup event? We can work on increasing the timeout, but we'd like to understand the scenario a little better.

In our case where the Uniswap V3-like contract is nearly 2 years old, I think the issue is that Ponder wraps entire event batches in single database transactions. Looking at the code, [syncEventsQuerySize](https://github.com/ponder-sh/ponder/blob/35c165bd745be34ef0d2c59a047d9628f2040aec/packages/core/src/internal/options.ts#L121) is set to 12,000, so batches can contain ~10k+ events that all get processed in one transaction.

Even though events are chunked into groups of 93 for processing (see [omnichain.ts:336](https://github.com/ponder-sh/ponder/blob/35c165bd745be34ef0d2c59a047d9628f2040aec/packages/core/src/runtime/omnichain.ts#L336)), the entire batch stays in the same transaction from start to finish. 

For events doing any meaningful work (RPC calls, calculations, external api calls, etc.), it's easy to hit the [1-hour idle timeout](https://github.com/ponder-sh/ponder/blob/35c165bd745be34ef0d2c59a047d9628f2040aec/packages/core/src/utils/pg.ts#L59) and lose all progress. Same issue affects any slow historical indexing.

The chunking helps with memory but doesn't solve the timeout since it's all still one DB transaction. Maybe break these into smaller transactional chunks or make the timeout configurable as in [this PR](https://github.com/ponder-sh/ponder/pull/1994)

PR #1994, opened 2025-08-24
feat: add idleInTransactionSessionTimeout to postgres pool config

iHiteshAgrawal:
Addresses https://github.com/ponder-sh/ponder/issues/1803


typedarray:
Closing for now - let us know if you're still struggling with #1803 and we can revisit.

PR #2001, opened 2025-08-26
improve rpc error logs

kyscott18:


packages/core/src/utils/rpc.ts
```diff
@@ -376,11 +406,26 @@ export const validateTransactionsAndBlock = (block: SyncBlock) => {
...
-export const validateLogsAndBlock = (logs: SyncLog[], block: SyncBlock) => {
+export const validateLogsAndBlock = (
+  logs: SyncLog[],
+  block: SyncBlock,
+  blockIdentifier: "number" | "hash",
+) => {
   if (block.logsBloom !== zeroLogsBloom && logs.length === 0) {
-    throw new Error(
-      "Detected invalid eth_getLogs response. `block.logsBloom` is not empty but zero logs were returned.",
+    const error = new RpcProviderError(
```
typedarray, line 415: Probably missing something, but shouldn't this validation only run if the logs were collected using eth_getLogs with blockHash and no filters?

packages/core/src/utils/rpc.ts
```diff
@@ -433,31 +541,64 @@ export const validateLogsAndBlock = (logs: SyncLog[], block: SyncBlock) => {
...
+      ];
+      error.stack = undefined;
+      throw error;
     }
 
     const id = `${trace.transactionHash}-${trace.trace.index}`;
     if (traceIds.has(id)) {
-      throw new Error(
-        `Detected invalid debug_traceBlockByHash response. Duplicate trace index ${trace.trace.index} for transaction ${trace.transactionHash}.`,
+      const error = new RpcProviderError(
```
typedarray: This one is a bit strange - if this failed, wouldn't we have introduced it in our DFS post-processing step?

typedarray (approved):

PR #2006, opened 2025-08-27
custom websocket client

kyscott18:
Adds a websocket client built directly on `ws` rather than relying on `viem` websocket client.

### DIFF
--- a/packages/core/src/rpc/index.ts
+++ b/packages/core/src/rpc/index.ts
@@ -43,7 +43,6 @@ export type Rpc = {
   subscribe: (params: {
     onBlock: (block: SyncBlock | SyncBlockHeader) => Promise<boolean>;
     onError: (error: Error) => void;
-    polling?: boolean;
   }) => void;
   unsubscribe: () => Promise<void>;
 };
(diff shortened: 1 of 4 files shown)


packages/core/src/rpc/index.ts
```diff
@@ -463,146 +454,193 @@ export const createRpc = ({
...
+
+              ws?.send(JSON.stringify(subscriptionRequest));
+            });
 
-                if (webSocketErrorCount === RETRY_COUNT) {
+            ws.on("message", (data: Buffer) => {
+              try {
+                const msg = JSON.parse(data.toString());
+                if (msg.method === "eth_subscription") {
+                  onBlock(standardizeBlock(msg.params.result, true));
```
khaidarkairbek, line 528: Would be great to add validation that subscriptionId of this mesage corresponds to the current subscription.

packages/core/src/rpc/index.ts
```diff
@@ -463,146 +454,193 @@ export const createRpc = ({
...
                   error,
                 });
 
                 webSocketErrorCount += 1;
+              }
 
-                await disconnect();
-
-                rpc.subscribe({ onBlock, onError, polling: false });
+              if (ws && ws.readyState !== ws.OPEN) {
```
khaidarkairbek: Should this be `ws && ws.readyState === ws.CLOSED`?

PR #2012, opened 2025-08-29
[wip] new offchain example

typedarray:


hskang9:
Can you add a README on how to run both modules? Also explaining reason why the write access is not allowed in api due to unclear term called "non-deterministic" on both data/state.

hskang9:
From the failures of the integration tests, I think it is better to just use the same onchain table with write access on API. The code is not stable for other developers to use.

Here, this PR is just fully backward compatible. https://github.com/ponder-sh/ponder/pull/2013

PR #2021, opened 2025-09-04
column selection

khaidarkairbek:


packages/core/src/indexing/index.ts
```diff
@@ -344,19 +410,143 @@ export const createIndexing = ({
...
 
+const proxyHandler = ({
+  type,
+  profile,
+}: {
+  type: "log" | "block" | "trace" | "transaction" | "transactionReceipt";
+  profile: ColumnAccessProfile;
+}): ProxyHandler<any> => {
+  return {
+    get(obj, prop, receiver) {
```
kyscott18: We should add defined behavior for other object properties like `set`, `has`, `ownKeys`, `delete`.

packages/core/src/indexing/index.ts
```diff
@@ -344,19 +410,143 @@ export const createIndexing = ({
...
+          profile.resolved = false;
+          throw new InvalidEventAccessError(key);
+        }
+      }
+      return Reflect.get(obj, prop, receiver);
+    },
+  };
+};
+
+export const toProxy = ({
```
kyscott18: Is it possible to have one `eventProxy`? 

More likely, we will need to initiate five proxies up front (block, transaction, receipt, trace, log) but then we could swap out the underlying `event` without having to create any more proxies. This comment is more performance focused, I think what you have is functionally correct.

packages/core/src/runtime/omnichain.ts
```diff
@@ -591,6 +604,14 @@ export async function runOmnichain({
...
                 }
 
+                if (error instanceof InvalidEventAccessError) {
+                  common.logger.warn({
+                    service: "app",
+                    msg: `Refetching '${chain.name}' events for block ${Number(decodeCheckpoint(checkpoint).blockNumber)} due to restricted column selection`,
+                  });
+                  events = await syncStore.refetchEventData({ events });
+                }
+
```
kyscott18: Delete this, column selection only for historical backfill.

packages/core/src/sync-store/index.ts
```diff
@@ -576,28 +592,21 @@ export const createSyncStore = ({
         shouldGetTransactionReceipt,
       );
 
+      const blockSelect: Record<string, any> = {};
+      for (const columnKey of columnAccessProfile.blockInclude) {
+        const columnName = columnKey.replace("block.", "");
```
kyscott18: Should `blockInclude` be refactored so that the elements are actual column names?

packages/core/src/sync-store/index.ts
```diff
@@ -576,28 +592,21 @@ export const createSyncStore = ({
...
+        blockSelect[columnName] =
+          PONDER_SYNC.blocks[columnName as keyof typeof PONDER_SYNC.blocks];
+      }
+
+      blockSelect.timestamp = PONDER_SYNC.blocks.timestamp;
+      blockSelect.number = PONDER_SYNC.blocks.number;
+      blockSelect.hash = PONDER_SYNC.blocks.hash;
+      blockSelect.logsBloom = PONDER_SYNC.blocks.logsBloom;
+      blockSelect.parentHash = PONDER_SYNC.blocks.parentHash;
```
kyscott18: Is there a centralized list of which columns are required because they are used internally by Ponder? I think we should update the internal types (`InternalBlock`, etc) to reflect this so we can have some type safety around it.

packages/core/src/indexing/index.ts
```diff
@@ -57,12 +65,69 @@ export type Indexing = {
...
+};
+
+export const createColumnAccessProfile = ({
+  common,
+}: { common: Common }): ColumnAccessProfile => ({
+  blockInclude: [...defaultBlockInclude],
+  traceInclude: [...defaultTraceInclude],
+  transactionInclude: [...defaultTransactionInclude],
+  transactionReceiptInclude: [...defaultTransactionReceiptInclude],
+  accessed: new Set(),
```
kyscott18: Do you think we should have the column selection be filter-specific for other data sources?

packages/core/src/indexing/index.ts
```diff
@@ -57,12 +65,69 @@ export type Indexing = {
...
+  eventCount: 0,
```
kyscott18: I believe this is already tracked, see `createIndexing` params

packages/core/src/indexing/index.ts
khaidarkairbek: We can use five proxies and just use handlers to use the object that will be replaced on each execution.

packages/core/src/sync-store/index.ts
khaidarkairbek: I refactored `blockInclude`, `transactionInclude`, ... to basically be actual column names and then filter includes are `block.${column}`, ... .

packages/core/src/indexing/index.ts
khaidarkairbek: Added an interface for indexing function specific profiling, which could later be used if needed by other data sources and potentially replace filter.include

packages/core/src/indexing-store/profile.ts
```diff
@@ -103,7 +103,7 @@ const matchEventParameters = (
         };
       }
 
-      if (eq(event.event.block.miner, value)) {
+      if (event.event.block.miner && eq(event.event.block.miner, value)) {
```
typedarray, line 112: To recap, `block.miner` is the only property that we use in profiling that is *not* part of the default `include`, which is the reason it is the only property present in the diff in this file. IMO we should remove it from profiling until we figure out a way to avoid calling registering property accesses when calling from internal/framework code.

packages/core/src/indexing/index.ts
```diff
@@ -253,6 +313,10 @@ export const createIndexing = ({
         throw new ShutdownError();
       }
 
+      if (error instanceof InvalidEventAccessError) {
```
typedarray, line 326: Why doesn't this new error need to use the `indexingErrorHandler` system? E.g. to avoid being swallowed by userland try/catch?

packages/core/src/indexing/index.ts
```diff
@@ -343,6 +507,91 @@ export const createIndexing = ({
           msg: `Started indexing function (event="${event.name}", checkpoint=${event.checkpoint})`,
         });
 
+        // Note: Create a new event object instead of mutuating the original one because
+        // the event object could be reused across multiple indexing functions.
+        const proxyEvent: typeof event.event = { ...event.event };
```
typedarray, line 524: I wonder how much overhead this introduces - would it be better to run the copy only if we know it will be reused?

packages/core/src/runtime/historical.ts
```diff
@@ -494,6 +494,129 @@ export async function* getHistoricalEventsMultichain(params: {
...
+      ({ filter }) => filter.chainId === chain.id,
+    );
+
+    const chainEvents = params.events.filter(
+      (event) => event.chainId === chain.id,
+    );
+
+    if (chainEvents.length === 0) continue;
+
+    const refetchedEvents = await initRefectEvents({
```
typedarray: Small typo, looks like it should be `initRefetchEvents`

packages/core/src/runtime/multichain.ts
```diff
@@ -431,7 +439,19 @@ export async function runMultichain({
           indexingCache.invalidate();
           indexingCache.clear();
 
-          if (error instanceof NonRetryableUserError === false) {
+          if (error instanceof InvalidEventAccessError) {
+            common.logger.warn({
+              service: "app",
+              msg: `Refetching events due to restricted column selection. Missing: '${error.key}' field.`,
```
typedarray: Maybe `Retrying event batch due to unexpected column access. Missing: ...`

packages/core/src/sync-store/index.ts
```diff
@@ -573,31 +583,56 @@ export const createSyncStore = ({
...
+      };
+      // Note: `LogSelect` doesn't exist because all log columns are required.
+
+      const blockSelect: BlockSelect = {
+        number: PONDER_SYNC.blocks.number,
+        hash: PONDER_SYNC.blocks.hash,
+        timestamp: PONDER_SYNC.blocks.timestamp,
+      };
+
+      for (const column of unionFilterIncludeBlock(filters)) {
```
typedarray, line 629: If in practice we union these, might we run into an issue where if any events occur fewer than 100 times, no column selection will be applied?

packages/core/src/indexing/index.ts
kyscott18: Not sure what you mean, it does use the `indexingErrorHandler` in `createEventProxy`.

packages/core/src/indexing/index.ts
kyscott18: This only does a shallow copy of about 4 properties - it's showed no difference in my benchmarks

packages/core/src/indexing-store/profile.ts
goums: curious why do we check `block.miner` for profiling, do we have track record of this attribute being used as a key in some applications?

kyscott18: No, but the profiling step is very cheap and happens with a low sampling rate, so more is usually better.

PR #2027, opened 2025-09-08
fix: database connection error handling

kyscott18:


packages/core/src/client/index.ts
```diff
@@ -46,25 +48,56 @@ export const client = ({
       });
     });
   } else {
-    const pool = driver.admin;
+    (async () => {
+      let listen: pg.PoolClient | undefined;
```
typedarray: I'd probably name this `connection` or `client`

packages/core/src/client/index.ts
```diff
@@ -46,25 +48,56 @@ export const client = ({
...
+          try {
+            listen = await driver.admin.connect();
 
-      driver.listen.on("error", async () => {
-        driver.listen?.release();
-        await connectAndListen();
-      });
+            listen.on("notification", () => {
+              statusResolver.resolve();
+              statusResolver = promiseWithResolvers();
```
typedarray, line 82: Is this pattern of reassigning the `statusResolver` local working as expected with the `await statusResolver.promise;` line below?

packages/core/src/client/index.ts
```diff
@@ -46,25 +48,56 @@ export const client = ({
...
+              await wait(250);
+
+              resolve();
+            });
+
+            await listen.query(`LISTEN "${channel}"`);
+          } catch (error) {
+            globalThis.PONDER_COMMON.logger.warn({
+              service: "client",
+              msg: "Received error on listen connection, retrying after 250ms",
```
typedarray, line 103: It might also be nice to have a debug log for "Established listen connection for client middleware", so you can tell if the retry worked.

packages/core/src/client/index.ts
```diff
@@ -46,25 +48,56 @@ export const client = ({
...
-      driver.listen = await pool.connect();
+      globalThis.PONDER_COMMON.shutdown.add(() => {
+        listen?.release();
+        listen = undefined;
+      });
 
-      await driver.listen.query(`LISTEN "${channel}"`);
+      while (globalThis.PONDER_COMMON.shutdown.isKilled === false) {
+        // biome-ignore lint/suspicious/noAsyncPromiseExecutor: <explanation>
+        await new Promise<void>(async (resolve) => {
```
typedarray, line 71: You might be able to use `continue` for control flow here instead of the promise + `resolve()`

packages/core/src/client/index.ts
```diff
@@ -46,25 +48,56 @@ export const client = ({
       });
     });
   } else {
-    const pool = driver.admin;
+    (async () => {
+      let listen: pg.PoolClient | undefined;
 
-    const connectAndListen = async () => {
-      driver.listen = await pool.connect();
+      globalThis.PONDER_COMMON.shutdown.add(() => {
```
typedarray, line 64: We might want to explicitly validate that the necessary `globalThis` properties are present at the very beginning of the `client(...)` function, and if they aren't throw an error like "client() middleware cannot be initialized outside of a Ponder project"

packages/core/src/client/index.ts
kyscott18: yes

packages/core/src/client/index.ts
kyscott18: Wouldn't that cause the loop to loop infinitely without waiting?

PR #2028, opened 2025-09-09
feat(core): allow custom `graphql-yoga` plugins

normanzb:
could we expose an plugins option to allow using yoga specific plugins? 
one of such example is [sentry plugin](https://the-guild.dev/graphql/envelop/plugins/use-sentry) for system monitoring


typedarray:
Not currently planned, may revisit down the road.

PR #2031, opened 2025-09-11
fix: startup hang and compatibility issues

barajeel:


kyscott18:
The changed file is only part of the Ponder test suite, doesn't affect the `ponder` package.

Issue #1776, opened 2025-05-22
Configure database indexes before running historical set


murderteeth:
### Problem / use case

In some cases I want to compute values based on other values that have already been indexed. Example: a points system where rewards related to Transfer events from one contract are divided up based on data indexed from 4626 Deposit events on another contract. When I index the Transfer events, i want to do sth like this

```

const accounts = await context.db.sql.select()
.from(schema.Account)
.where(eq(schema.Account.vaultAddress, someErc4626.address))
for (account in accounts) {
  ...
}

```

This is ok if `schema.Account.erc4626Address` is the PK or first col of the PK. If it's not the PK we might try this in our schema:

```

export const Account = onchainTable("Account", (t) => ({
  address: t.hex().primaryKey(),
  vaultAddress: t.hex().notNull(),
  ...
}), (table) => ({
  vaultIdx: index().on(table.vaultAddress),
}));

```

but unlike with primary keys, ponder delays creating this index until after the historical set is complete, per https://ponder.sh/docs/schema/tables#indexes

I also tried manually adding the index with `CREATE INDEX`, but ponder deletes that index when syncing its schema.

Without the database index i need, my indexing code slows to a crawl.

### Proposed solution

Implement an override for the delayed database index behavior. When the override is used, the index is created when ponder starts and syncs schema.

Maybe could be implemented with a `createOnStart` builder method like this,

```

export const Account = onchainTable("Account", (t) => ({
  address: t.hex().primaryKey(),
  vaultAddress: t.hex().notNull(),
  ...
}), (table) => ({
  vaultIdx: index().on(table.vaultAddress).createOnStart(),
}));

```

typedarray:
Prerelease `0.0.0-kjs-early-indexes-20250522153448` (patch applied to 0.11.5) creates indexes before the backfill. If anyone runs it and can share before/after performance for the `ponder_indexing_store_raw_sql_duration` metric, that would help build our case to include this on main.

<img width="812" alt="Image" src="https://github.com/user-attachments/assets/063a64b6-cffd-45bb-9647-04b3044211fe" />

Jojoooo1:
Might be some issue with postgres planner after large copy ? Observed similar issue during historical sync usually some of my index are not used.

PR #2032, opened 2025-09-11
allow for creating database indexes before indexing

erikolsson:
Hi!

During historical sync, I need to perform many lookups. These lookups would benefit significantly from custom indices. However, since Ponder creates indexes after the historical sync, my syncs take much longer than necessary.

I understand the reasoning behind adding indexes after sync, but for my use case where I frequently query non-indexed columns on tables with more than 500k rows, I think the benefits of having indexes available during sync outweigh the downsides.

I believe this should be available as a configurable option.

I've done my best to provide a minimal implementation + docs, let me know what you think and if you need anything else!

### DIFF
--- a/.changeset/wild-symbols-unite.md
+++ b/.changeset/wild-symbols-unite.md
@@ -2,4 +2,4 @@
 "ponder": patch
 ---
 
-add PONDER_DATABASE_CREATE_INDEXES_EARLY option
+Added `DATABASE_INDEXES_EARLY` environment variable and `--indexes-early` CLI flag for creating database indexes immediately rather than after the completion of the historical backfill.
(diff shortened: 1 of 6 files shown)


kyscott18:
@erikolsson we're interested in adding this - could you give me write access to your fork of Ponder so I could make some tweaks before merging?

erikolsson:
> @erikolsson we're interested in adding this - could you give me write access to your fork of Ponder so I could make some tweaks before merging?

Done, thank you!

typedarray:
Thanks for opening this - we discussed internally and have decided on a different API that makes it possible to mark specific indexes (one by one) to be created before the backfill. This will make it possible to create indexes that target specific raw SQL queries that are slowing down the backfill, vs. a wholesale option.

This will require a bit more work (with Drizzle internals) and will look rather different from this PR, so closing for now.

Related issue: https://github.com/ponder-sh/ponder/issues/1776

PR #2035, opened 2025-09-15
fix: handle missing address and empty results safely

famouswizard:
added a couple of safety checks to avoid runtime crashes:

* return early if `address` param is missing.
* use optional chaining + default for `result[0]?.count`.

just makes things safer and stops random errors from popping up.


kyscott18:
This functionally doesn't change the behavior at all.

PR #2051, opened 2025-09-23
improve `getRpcRequestResults` performance

kyscott18:


packages/core/src/sync-store/index.ts
```diff
@@ -1089,13 +1090,102 @@ export const createSyncStore = ({
...
+                    eq(PONDER_SYNC.rpcRequestResults.chainId, BigInt(chainId)),
+                    inArray(
+                      PONDER_SYNC.rpcRequestResults.requestHash,
+                      nonBlockNumberRequests.map((request) =>
+                        crypto
+                          .createHash("md5")
+                          .update(
+                            toLowerCase(JSON.stringify(orderObject(request))),
+                          )
+                          .digest("hex"),
```
goums: I think we could avoid computing again the requests hashes for those request as it was computed just before already for all requests.
Would save some CPU.

packages/core/src/sync-store/index.ts
```diff
@@ -1089,13 +1090,102 @@ export const createSyncStore = ({
...
+        nonBlockNumberRequests.push(request);
+      }
+    }
+
     const requestHashes = requests.map((request) =>
       crypto
         .createHash("md5")
         .update(toLowerCase(JSON.stringify(orderObject(request))))
         .digest("hex"),
     );
```
goums: Maybe something like this would avoid an extra loop and recomputation of hashes:

```suggestion
    const blockNumberRequests: (EIP1193Parameters & { blockNumber: number })[] =
      [];
    const nonBlockNumberRequests: (EIP1193Parameters & { hash: string })[] = [];
    const requestHashes: string[] = [];
    for (const request of requests) {
    
      const hash =  crypto
        .createHash("md5")
        .update(toLowerCase(JSON.stringify(orderObject(request))))
        .digest("hex");
        
      const blockNumber = extractBlockNumberParam(request);

      // Note: "latest" is sorted into `nonBlockNumberRequests`
      if (isHex(blockNumber)) {
        blockNumberRequests.push({
          ...request,
          blockNumber: hexToNumber(blockNumber),
        });
      } else {
        nonBlockNumberRequests.push({...request, hash});
      }
    }

   requestHashes.push(hash);
```

PR #2064, opened 2025-09-24
Improve realtime sync error handling

kyscott18:


packages/core/src/sync-realtime/index.ts
```diff
@@ -806,214 +796,178 @@ export const createRealtimeSync = (
...
-          );
-        }
-
-        const text = _text.filter((t) => t !== undefined).join(" and ");
-        args.common.logger.info({
-          service: "realtime",
-          msg: `Synced ${text} from '${args.chain.name}' block ${hexToNumber(block.number)}`,
-        });
+        blockCallback?.(false);
       } else {
```
typedarray, line 847: What's the logical change here?

packages/core/src/sync-realtime/index.ts
```diff
@@ -1070,15 +1025,17 @@ export const createRealtimeSync = (
...
         // Note: `reconcileBlock` must be called serially.
 
         await realtimeSyncLock.lock();
 
-        yield* reconcileBlock(blockWithEventData, blockCallback);
+        try {
+          yield* reconcileBlock(blockWithEventData, blockCallback);
+        } finally {
+          realtimeSyncLock.unlock();
+        }
```
typedarray, line 1036: I wonder if failing to unlock the mutex here could be leading to the rare state where realtime blocks stop being processed (IIRC when we've seen this, it's after lots of RPC instability).

packages/core/src/sync-realtime/index.ts
```diff
@@ -92,7 +91,7 @@ type CreateRealtimeSyncParameters = {
   childAddresses: Map<FactoryId, Map<Address, number>>;
 };
 
-const MAX_LATEST_BLOCK_ATTEMPT_MS = 3 * 60 * 1000; // 3 minutes
+const MAX_LATEST_BLOCK_ATTEMPT_MS = 10 * 60 * 1000; // 3 minutes
```
typedarray: Update comment

packages/core/src/sync-realtime/index.ts
kyscott18: removed try / catch. If you hide the whitespace it's a small diff

packages/core/src/sync-realtime/index.ts
kyscott18: Possibly, at least one of the cases that I've seen is from a Ponder version before this code was introduced.

PR #2066, opened 2025-09-24
fix: graphiql fetcher path

typedarray:


packages/core/src/graphql/graphiql.html.ts
```diff
@@ -44,7 +44,7 @@ export const graphiQLHtml = (path: string) => `<!--
     <script src="https://cdn.jsdelivr.net/npm/graphiql@3.7.2/graphiql.min.js" crossorigin="anonymous"></script>
     <script src="https://cdn.jsdelivr.net/npm/@graphiql/plugin-explorer@3.2.3/dist/index.umd.js" crossorigin="anonymous"></script>
     <script>
-      const fetcher = GraphiQL.createFetcher({ url: "${path}" });
+      const fetcher = GraphiQL.createFetcher({ url: "" });
```
typedarray, line 47: Turns out passing `""` leads to the desired behavior (requests use the current URL).

PR #2073, opened 2025-09-26
improve eventloop and metrics updates

kyscott18:


packages/core/src/runtime/multichain.ts
```diff
@@ -333,46 +332,49 @@ export async function runMultichain({
...
-          historicalIndexingStore.isProcessingEvents = false;
+          // Note: at this point, the next events can be preloaded, as long as the are not indexed until
+          // the "flush" + "finalize" is complete.
 
-          await new Promise(setImmediate);
+          await indexingCache.flush();
+
+          common.metrics.ponder_historical_transform_duration.inc(
+            { step: "load" },
+            endClock(),
```
typedarray, line 376: Looks like you may have eliminated one clock reset - intentional?

kyscott18: hmm still looks correct to me

PR #2079, opened 2025-09-29
Feat/optional finality block count config

ssups:
Hi!
I think it would be helpful to support a configurable finality block count in the chain config, so we can better accommodate different EVM-based chains.
What do you think?
Thank you.

### DIFF
--- a/.changeset/clean-shirts-bake.md
+++ b/.changeset/clean-shirts-bake.md
@@ -1,0 +1,5 @@
+---
+"ponder": minor
+---
+
+add finalityBlockCount config option


typedarray:
Hi! Definitely agree that the finality block count selection is a bit unscientific at the moment.

Have you encountered specific problems associated with the finality block count? We're not currently aware of concrete issues around this, and we generally don't add configuration options unless they solve a known problem.

typedarray:
Closing for now - please let us know if you're aware of issues caused by the finality block count.

PR #2080, opened 2025-09-29
fix: `client.db` handling many connections

kyscott18:


packages/core/src/client/index.test.ts
```diff
@@ -200,3 +201,46 @@ FROM infinite_cte;`,
...
+  );
+
+  const query = {
+    sql: "SELECT * FROM account",
+    params: [],
+  };
+
+  // Note: This fixes a race condition where the listen shutdown
+  // gets called before the listen connection is established.
+  await wait(50);
```
typedarray: Could we move the shutdown function to a `then` on the listen connection promise?

PR #2117, opened 2025-10-14
fix interval cache for factories

kyscott18:
Fixes https://t.me/pondersh/15618

### DIFF
--- a/.changeset/proud-moons-do.md
+++ b/.changeset/proud-moons-do.md
@@ -1,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Fixed a bug that caused factories to miss child addresses after updating the factory start block . Please note that this did not affect the rpc cache, users do not have to refetch block data.


packages/core/src/runtime/filter.ts
```diff
@@ -571,3 +572,87 @@ export const unionFilterIncludeTrace = (filters: Filter[]): (keyof Trace)[] => {
...
+
+      if (isAddressFactory(filter.toAddress)) {
+        intervals.push([
+          filter.toAddress.fromBlock ?? 0,
+          filter.toAddress.toBlock ?? Number.POSITIVE_INFINITY,
+        ]);
+      }
+  }
+
+  return sortIntervals(intervals);
```
typedarray, line 657: Would you also want to merge/union these before sorting and returning?

packages/core/src/runtime/filter.ts
```diff
@@ -571,3 +572,87 @@ export const unionFilterIncludeTrace = (filters: Filter[]): (keyof Trace)[] => {
   }
   return Array.from(includeTrace);
 };
+
+export const filterFromBlock = (filter: Filter): number => {
```
typedarray, line 576: Would name these `getFilterFromBlock` and `getFilterIntervals`, starting with "filter" makes it sound like a verb/action but filter is being used as a noun here

packages/core/src/runtime/index.ts
```diff
@@ -242,35 +225,28 @@ export const getCachedBlock = ({
...
 
-    const earliestCompletedInterval = completedIntervals[0]!;
-    if (earliestCompletedInterval[0] !== (filter.fromBlock ?? 0)) {
-      // Use `fromBlock` - 1 as completed block if the earliest
-      // completed interval does not start at `fromBlock`.
-      if ((filter.fromBlock ?? 0) === 0) return undefined;
-      return filter.fromBlock! - 1;
-    }
-    return earliestCompletedInterval[1];
+    if (missingIntervals[0]![0] === 0) return undefined;
```
typedarray, line 248: Might be better to pull this out into a const and name it, not sure what this represents

kyscott18:
See #2164

PR #2119, opened 2025-10-14
fix `usePonderQuery` types to support `select` transformer

noahsaso:
This fixes the types of `usePonderQuery` to support the `select` function of `useQuery` (https://tanstack.com/query/latest/docs/framework/react/reference/useQuery), which conveniently allows transforming the results of a query. It was just missing some generic type arguments.

Because Ponder wraps `useQuery` and expects `queryFn` to return generated SQL in order to automatically compute the `queryKey` for caching, the `select` function is the only way to transform results of queries to conform data to local frontend types.

### DIFF
--- a/.changeset/shaggy-buckets-visit.md
+++ b/.changeset/shaggy-buckets-visit.md
@@ -1,0 +1,5 @@
+---
+"@ponder/react": patch
+---
+
+Fixed `select` property types in `usePonderQuery`.
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -534,7 +534,7 @@ importers:
     dependencies:
       forge-std:
         specifier: github:foundry-rs/forge-std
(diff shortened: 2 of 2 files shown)


typedarray:
Thanks for opening! I wonder if the useQuery `select` option would work here instead?
Screenshot 2025-10-16 at 3 54 21 PM

Docs: https://tanstack.com/query/latest/docs/framework/react/reference/useQuery

noahsaso:
> Thanks for opening! I wonder if the useQuery `select` option would work here instead?
> 
> Screenshot 2025-10-16 at 3 54 21 PM
> 
> 
> 
> Docs: https://tanstack.com/query/latest/docs/framework/react/reference/useQuery

oh lol yeah that looks like exactly what i want. i wish i didn't have to memoize the select callback but that's fine. i'll try it and report back.

noahsaso:
alright @typedarray, `select` is great but `usePonderQuery` needed types adjusted to support the `queryFn` and result having different types based on the `select` function. this PR should be good to go now!

PR #2122, opened 2025-10-15
docs: fix broken link to Call Traces

hexcow:
the link to the Call Traces documentation was returning a 404.
updated it to point to the correct guide URL so users can access the page without errors.


kyscott18:
Not going to update the changelog from last year

PR #2124, opened 2025-10-15
`ordering: "experimental_isolated"`

kyscott18:


packages/core/src/build/schema.ts
```diff
@@ -91,6 +96,32 @@ export const buildSchema = ({ schema }: { schema: Schema }) => {
...
+        if (
+          getPrimaryKeyColumns(s).some(
+            ({ js }) =>
+              js.toLowerCase() === "chainid" || js.toLowerCase() === "chain_id",
+          ) === false
+        ) {
+          throw new Error(
+            `Schema validation failed: '${name}.chain_id' column is required to be in the primary key when ordering is 'isolated'.`,
+          );
+        }
```
typedarray, line 129: Should probably also validate the column type here?

packages/core/src/database/queryBuilder.ts
```diff
@@ -1,3 +1,4 @@
+import crypto from "node:crypto";
```
typedarray, line 1: rm?

packages/core/src/bin/commands/start.ts
```diff
@@ -292,11 +299,19 @@ export async function start({
...
-    runMultichain(app);
+  switch (preCompileResult.result.ordering) {
+    case "omnichain":
+      runOmnichain(app);
+      break;
+    case "multichain":
+      runMultichain(app);
+      break;
+    case "isolated": {
+      isolatedController(app);
```
typedarray, line 310: FWIW, I'd expect this to be named runIsolated and for the file(s) to be in the runtime dir

packages/core/src/bin/isolatedController.ts
```diff
@@ -0,0 +1,263 @@
...
+      count: perThreadWorkers.length,
+      duration: backfillEndClock(),
+    });
+
+    common.metrics = new AggregateMetricsService(
+      common.metrics,
+      perThreadWorkers,
+    );
+
+    common.shutdown.add(async () => {
```
typedarray: async?

packages/core/src/bin/isolatedWorker.ts
```diff
@@ -0,0 +1,183 @@
...
+import { IsolatedMetricsService } from "@/internal/metrics.js";
+import { createShutdown } from "@/internal/shutdown.js";
+import { createTelemetry } from "@/internal/telemetry.js";
+import type {
+  CrashRecoveryCheckpoint,
+  NamespaceBuild,
+} from "@/internal/types.js";
+import { runIsolated } from "@/runtime/isolated.js";
+
+if (isMainThread === false && parentPort) {
```
typedarray: Is it an invariant that this will never run on the main thread? Might be worth validating for completeness

packages/core/src/database/queryBuilder.ts
kyscott18: This fixed a bug with Railway, we use `crypto` to generate a query ID

PR #2126, opened 2025-10-16
improve realtime latency

kyscott18:
This includes two changes aimed at reducing realtime latency, improving Ponder's ability to track chain tips.

## Background
The realtime indexing pipeline uses several `AsyncGenerator`s 
1. fetch block data, reconcile into local chain
2. extract extract from events, order blocks/events across all chains
3. index block events with user indexing functions

An `AsyncGenerator` yields a value, and doesn't continue execution until the value is consumed.

## Implementation
In order to minimize the amount of time spent waiting for a value from an `AsyncGenerator` to be consumed, I added a buffer between each of the `AsyncGenerator`s .

I also removed the `isFetching` deduplication logic, superseded by the `latestFetchedBlock` introduced in #2093.

### DIFF
--- a/packages/core/src/runtime/realtime.ts
+++ b/packages/core/src/runtime/realtime.ts
@@ -111,17 +111,6 @@ export async function* getRealtimeEventsOmnichain(params: {
           syncStore: params.syncStore,
         }),
         100,
-        (bufferSize) => {
-          if (bufferSize < 100) return;
-
-          params.common.logger.warn({
-            msg: "Live indexing reached buffer limit, applying backpressure",
-            chain: chain.name,
-            chain_id: chain.id,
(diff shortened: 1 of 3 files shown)


packages/core/src/runtime/realtime.ts
```diff
@@ -100,15 +100,29 @@ export async function* getRealtimeEventsOmnichain(params: {
...
+          syncProgress,
+          childAddresses,
+          syncStore: params.syncStore,
+        }),
+        100,
+        (bufferSize) => {
+          if (bufferSize < 100) return;
+
+          params.common.logger.warn({
+            msg: "Detected live indexing backpressure",
```
typedarray: Live block buffer reached limit, applying backpressure

PR #2130, opened 2025-10-17
fix: "unable to decode" log volume

typedarray:
Screenshot 2025-10-17 at 11 38 47 AM

### DIFF
--- a/packages/core/src/runtime/events.ts
+++ b/packages/core/src/runtime/events.ts
@@ -752,7 +752,7 @@ export const decodeEvents = (
 
   if (logDecodeFailureCount > 0) {
     common.logger.debug({
-      msg: "Failed to decode matched event logs",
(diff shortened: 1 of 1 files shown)


packages/core/src/runtime/events.ts
```diff
@@ -530,135 +530,143 @@ export const decodeEvents = (
...
-            try {
-              if (
-                event.log!.topics[0] === undefined ||
-                source.abiEvents.bySelector[event.log!.topics[0]] === undefined
-              ) {
-                throw new Error();
-              }
+            const selector = event.log!.topics[0];
+            if (selector === undefined) {
+              break;
```
kyscott18, line 550: For future reference, it's a de facto invariant that this should never happen, and future type changes, like `EventCallback`, should make it impossible.

kyscott18 (approved):

Issue #1135, opened 2024-10-04
RangeError: Too many elements passed to Promise.all


pyk:
Hey, I tried building an NFT indexer using Ponder 0.6.5 and encountered the following error:

```
2:17:18 AM ERROR process    Caught unhandledRejection event
RangeError: Too many elements passed to Promise.all
    at Function.all (<anonymous>)
    at syncLogFilter (file:///app/node_modules/@ponder/core/src/sync-historical/index.ts:225:34)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at file:///app/node_modules/@ponder/core/src/sync-historical/index.ts:463:21
    at async Promise.all (index 0)
    at file:///app/node_modules/@ponder/core/src/sync-historical/index.ts:457:11
    at async Promise.all (index 0)
    at Object.sync (file:///app/node_modules/@ponder/core/src/sync-historical/index.ts:424:7)
    at localHistoricalSyncGenerator (file:///app/node_modules/@ponder/core/src/sync/index.ts:1035:23)
    at file:///app/node_modules/@ponder/core/src/utils/generators.ts:11:22
2:17:18 AM WARN  process    Received unhandledRejection, starting shutdown sequence
2:17:18 AM FATAL process    Finished shutdown sequence, terminating (exit code 1)
```

I'm using the following Ponder version, config, and event listeners, which might help reproduce the error:

**Ponder version:**

```
        "@ponder/core": "0.6.5",
        "hono": "4.6.3",
        "viem": "2.21.16"
```

**Ponder config:**

```ts
import { createConfig } from "@ponder/core"
import { http } from "viem"

import { ERC721Abi } from "./abis/ERC721Abi"
import { ERC1155Abi } from "./abis/ERC1155Abi"

export default createConfig({
    networks: {
        mainnet: {
            chainId: Number(process.env.CHAIN_ID),
            transport: http(process.env.PONDER_RPC_URL),
            maxRequestsPerSecond: 15,
        },
    },
    contracts: {
        ERC721: {
            network: "mainnet",
            abi: ERC721Abi,
            startBlock: 0,
        },
        ERC1155: {
            network: "mainnet",
            abi: ERC1155Abi,
            startBlock: 0,
        },
    },
    database: {
        kind: "postgres",
        schema: `nft_ponder_${process.env.DEPLOYMENT_ID?.slice(0, 8)}`,
    },
})
```

**Event listeners:**

```ts
[...]

PR #2149, opened 2025-10-23
Improve historical sync memory usage

### LINKED ISSUE #1287: bug: Unbounded `historicalSync.sync()`
## Description
`historicalSync.sync()` takes `interval` as a parameter and syncs all events in the given interval. If this interval is event dense, or very large, the function invocation can attempt to sync too many events, causing an OOM or other errors. Ideally, there would be an escape hatch when too many objects are held at once. This algorithm should optimize for correctness and not performance, as this rarely happens.

Related #1135

kyscott18:
This improves the `sync-historical` memory usage and performance.

### Background
The `sync-historical` module fetches JSON-RPC data and inserts it into the `ponder_sync` database tables. The module takes a list of `Filter`s to be synced, decides the overall `Interval` to sync, and returns an `AsyncGenerator` that yields the closest-to-tip synced block number for all `Filter`s.

Our current design breaks the overall `Interval` into chunks that it estimates can complete in 1 second. Each estimated `Interval` is processed sequentially, and all relevant block, transactions, receipts, traces, and logs for the estimated `Interval` are held in-memory. When the estimate is too large, the block data objects can consume all available memory and cause an OOM.

### Implementation
The new algorithm still uses the same estimated `Interval` design, but breaks the process of syncing an estimated `Interval` into two different stages. The first stage is the "range" stage. It fetches all data that can be fetched in a range (with "from" + "to" parameters). Currently that is only logs, including logs used for factory addresses. The second stage is the "block" stage. This stage fetches all data that must be fetched block-by-block. 

The "block" stage currently processes up to a few hundred blocks concurrently (very similar to design before #1014). Block data is inserted into the database after each individual block is fetched, rather than at the end of the estimated `Interval`. Because of this, the overall memory usage is capped, regardless of how large the estimated `Interval` is. The new memory behavior also allows us to use a more aggressive estimate `Interval` algorithm.

This implementation is pipelined, the first stage runs at the same time as the second stage of the previous interval. At a given time, only one "range" stage and one "block" stage is executing.
image

Also, this pr uses a database transaction for inserting both live and backfill block data.

Closes #1287 #1135 
Related #1973

### DIFF
--- a/packages/core/src/sync-historical/index.ts
+++ b/packages/core/src/sync-historical/index.ts
@@ -427,8 +427,6 @@ export const createHistoricalSync = (
         }),
       );
 
-      // TODO(kyle) exit early if the number of logs is greater than 50k
-
       await Promise.all(
         requiredIntervals
           .filter(({ filter }) => filter.type === "log")


packages/core/src/utils/promiseAllSettledWithThrow.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Like `Promise.allSettled` but throws if any of the promises reject.
+ *
+ * @dev This is very useful when dealing with multiple concurrent promises
+ * in a database transaction.
+ */
+export function promiseAllSettledWithThrow<T>(
```
typedarray, line 7: Isn't this how Promise.all works?

packages/core/src/runtime/realtime.ts
```diff
@@ -761,102 +763,90 @@ export async function handleRealtimeSyncEvent(
...
-            chainId: params.chain.id,
-          },
-          context,
-        ),
-        ...Array.from(childAddresses.entries()).map(
-          ([factory, childAddresses]) =>
-            params.syncStore.insertChildAddresses(
+          ]);
+
+          if (params.chain.disableCache === false) {
```
typedarray, line 816: Should we just skip the entire transaction here if disableCache is true (avoid writing data entirely) ?

packages/core/src/runtime/historical.ts
```diff
@@ -426,7 +437,10 @@ export async function* getHistoricalEventsMultichain(params: {
           ) {
             from = crashRecoveryCheckpoint;
           } else {
-            const fromBlock = await params.syncStore.getSafeCrashRecoveryBlock({
+            const fromBlock = await createSyncStore({
```
typedarray, line 440: Does it make more sense to set up the sync store once on line ~82 ?

packages/core/src/sync-historical/index.ts
```diff
@@ -464,613 +344,595 @@ export const createHistoricalSync = (
...
-    // Note: `factory` must refer to the same original `factory` in `filter`
-    // and not be a recovered factory from `recoverFilter`.
-    childAddressesCache.set(factory, childAddresses);
+    return childAddresses;
   };
 
   /**
-   * Return all addresses that match `filter` after extracting addresses
-   * that match `filter` and `interval`.
+   * Fetch child addresses for `factory` within `interval`
```
typedarray, line 351: This comment is the same as the comment for `syncLogAddressFactory` - would be helpful to clarify the difference

packages/core/src/sync-historical/index.ts
```diff
@@ -464,613 +344,595 @@ export const createHistoricalSync = (
...
+        initialStart: true,
+        concurrency: MAX_BLOCKS_IN_MEM,
+        worker: syncBlockData,
+      });
 
-      let childAddressCount = 0;
-      for (const childAddresses of childAddressesCache.values()) {
-        childAddressCount += childAddresses.size;
-      }
+      await Promise.all(intervalRange(interval).map(queue.add));
```
typedarray, line 916: This could in practice enqueue a large number of no-op tasks, correct? Perhaps we should have a hard cap on the `interval` size.

packages/core/src/utils/promiseAllSettledWithThrow.ts
kyscott18: No, when a promise that is an argument to `Promise.all` rejects, the overall promise immediately rejects

packages/core/src/runtime/realtime.ts
kyscott18: We need to insert the data so that it can be queried

packages/core/src/runtime/historical.ts
kyscott18: I think it's better this way because some `createSyncStore` take query builders backed by transactions and some don't

packages/core/src/sync-historical/index.ts
kyscott18: Yes, the no-op tasks should resolve immediately. We already have a limit on the max `interval` size, it's currently 100,000.

PR #2152, opened 2025-10-24
improve `sql-over-http`

kyscott18:
This adds improved performance to `sql-over-http` and an invariant that live queries will update **only when the query result changes**.

### Background
Ponder has an HTTP server capable with `/db` and `/live` endpoints. Both endpoints take advantage of intrinsic properties of Ponder. Specifically, tables are only updated in discrete transactions for each block during live indexing.

Currently, there is a trigger on the `_ponder_checkpoint` table. It calls a procedure that sends a notification on every update. The `client` middleware listens for notifications on the channel. Every time the `_ponder_checkpoint` table is updated, the cache backing the `/db` endpoint is cleared and all live queries are sent an update message with SSE.

### Implementation
- all queries parsed to find referenced tables, with support for views
- live queries are saved on the server with the previous result hash
- live queries updated and cached queries deleted when a table they reference is updated
- live queries send full query result instead of update message

Ponder creates a temporary table for each "live" indexing transaction. Every tables has a trigger to write to the temporary table when it is updated. Before committing the transaction, the rows of the temporary table are used to populate a Postgres notification. When a query, live or standard, is received, it is parsed to find which tables are referenced. This step also supports views: if a query references a view that references a table, the underlying table will be detected. The referenced tables for each query are stored in memory. Every time a block is indexed, the updated tables are used to decide which queries to invalidate. Invalidate queries from the in-memory cache are deleted.

Live queries use the same query invalidation logic as mentioned above to intelligently update. When a query is invalidated, it is re-executed and the new result is compared to the previous. If the result is updated, it is pushed to the client with SSE. This is net new behavior and saves an HTTP roundtrip for live queries.

Also, updated `client.db` to use "GET" requests, live queries only available during live indexing.

### Related
- #2133

### DIFF
--- a/docs/pages/docs/query/sql-over-http.mdx
+++ b/docs/pages/docs/query/sql-over-http.mdx
@@ -181,7 +181,7 @@ The `@ponder/client` package re-exports all Drizzle utility functions (like `des
 
 ### Live queries
 
-Live queries are automatically updated when the underlying data changes. Ponder uses Server-Sent Events to stream updates to the client every time a new block is processed.
+Live queries are automatically updated when the underlying data changes. Ponder uses Server-Sent Events to stream updates to the **only when the query result changes**.
 
 :::code-group
 
@@ -467,4 +467,3 @@ Here are the measures taken by the `client` middleware to prevent malicious quer
   SET statement_timeout = '500ms';
   SET lock_timeout = '500ms';
   ```
-


packages/core/src/drizzle/onchain.ts
```diff
@@ -38,6 +40,16 @@ import {
 } from "./json.js";
 import { PgTextBuilder, type PgTextBuilderInitial } from "./text.js";
 
+export const getLiveQueryTriggerName = (table: PgTable) => {
+  return `live_query_trigger_${getTableName(table)}`;
+};
+export const getLiveQueryProcedureName = (table: PgTable) => {
+  return `live_query_procedure_${getTableName(table)}()`;
```
typedarray: We might need to use a shorter prefix string here. If the user provides two table names that are quite long and only differ in the last X characters, we could get a name collision.

packages/client/src/index.ts
```diff
@@ -138,31 +141,77 @@ export const createClient = <schema extends Schema>(
...
+            // @ts-ignore
+            await drizzleShim._.session
+              .prepareQuery(
+                query,
+                // @ts-ignore
+                orderedFields,
+                undefined,
+                false,
+              )
+              .execute(),
```
typedarray: Can this throw? If so we'd probably want it to call `onError` ?

kyscott18: It already does

PR #2153, opened 2025-10-24
add JSON-RPC warn log for requests longer than timeout

kyscott18:


packages/core/src/rpc/index.ts
```diff
@@ -420,6 +420,18 @@ export const createRpc = ({
         const endClock = startClock();
         const id = crypto.randomUUID().slice(0, 8);
 
+        const surpassTimeout = setTimeout(() => {
+          logger.warn({
+            msg: "JSON-RPC request unexpectedly surpassed timeout",
```
kyscott18, line 439: @typedarray Can you think of a better message?

PR #2158, opened 2025-10-29
fix: eth_getLogs range for merkle

typedarray:


packages/utils/src/_test/merkle.test.ts
```diff
@@ -4,29 +4,30 @@ import { getLogsRetryHelper } from "../getLogsRetryHelper.js";
 import { type Params, UNI, WETH, fromBlock, getRequest } from "./utils.js";
 
 const request = getRequest("https://eth.merkle.io");
-const maxBlockRange = 10_000n;
+const maxBlockRange10k = 10_000n;
+const maxBlockRange1k = 1_000n;
 
-test("merkle success", async () => {
+test("merkle success 10k", async () => {
```
kyscott18, line 10: How does this succeed? Isn't this the same as the 1k error case?

typedarray: It fails. But, I don't see why we'd want to remove the 10k handler (they could change their limit back to 10k)

kyscott18: You could make the logic use the value in the message

PR #2169, opened 2025-11-03
use custom HTTP client

kyscott18:
This pr adds support for throwing timeout errors when `await response.json()` takes longer than ten seconds.

### DIFF
--- a/packages/core/src/rpc/http.ts
+++ b/packages/core/src/rpc/http.ts
@@ -1,3 +1,4 @@
+import { URL } from "node:url";
 import type { Common } from "@/internal/common.js";
 import type { Chain } from "@/internal/types.js";
 import { HttpRequestError, RpcRequestError, TimeoutError } from "viem";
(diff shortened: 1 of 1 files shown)


packages/core/src/rpc/http.ts
```diff
@@ -0,0 +1,159 @@
...
+
+          reject(new TimeoutError({ body, url }));
+
+          if (reader) {
+            common.logger.warn({
+              msg: "JSON-RPC request timed out reading response body",
+              chain: chain.name,
+              chain_id: chain.id,
+              // @ts-ignore
+              request_id: headers ? headers["X-request-id"] : undefined,
```
jaylmiller: this should be all lowercase `"x-request-id"`, right?

kyscott18: I'm pretty sure the convention is upper case X

typedarray: I think HTTP clients/servers must treat headers as case insensitive to be compliant - but convention for this one seems to be `X-Request-ID` https://http.dev/x-request-id

PR #2173, opened 2025-11-06
fix: interval cache for factories

kyscott18:
This pr fixes a bug with how Ponder's RPC cache handles filters containing factories with different start blocks than the filter.

### Implementation
A new `factory_log` fragment type was added to the `intervals` table. It represents the completed block ranges for a `Factory`. 

There are three functions where interval math happens:
- `getCachedBlock`: Used to determine where to start `syncHistorical` from
- `getRequiredIntervals`: Used to determine the cache % (non-critical)
- `getRequiredIntervalsWithFilters`: Used to determine which filter + intervals should be fetched and what to insert into the `intervals` table.

Each function has been updated to account for this axiom: **When a filter with a factory is missing blocks, all blocks after the first missing block are also missing.** 

No user should have to drop their `ponder_sync` data, even if they are currently experiencing the bug.

Related #2117 
Related #2164

### DIFF
--- a/integration-test/src/index.ts
+++ b/integration-test/src/index.ts
@@ -858,6 +858,7 @@ const onBuild = async (app: PonderApp) => {
 
   if (SIM_PARAMS.MAX_UNCACHED_BLOCKS > 0) {
     for (const interval of await APP_DB.select().from(PONDER_SYNC.intervals)) {
+      if (interval.fragmentId.startsWith("factory_")) continue;
       const intervals: [number, number][] = JSON.parse(
         `[${interval.blocks.slice(1, -1)}]`,
       );
--- a/packages/core/src/runtime/historical.ts
+++ b/packages/core/src/runtime/historical.ts
@@ -1252,7 +1252,7 @@ export async function* getLocalSyncGenerator(params: {
     });
 
     let closestToTipBlock: SyncBlock | undefined;
-    if (requiredIntervals.length > 0) {
+    if (requiredIntervals.length > 0 || requiredFactoryIntervals.length > 0) {
       const pwr = promiseWithResolvers<void>();
 
       const durationTimer = setTimeout(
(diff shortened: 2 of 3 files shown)


packages/core/src/runtime/index.ts
```diff
@@ -257,127 +245,244 @@ export async function getCachedIntervals(params: {
...
-            Math.max(filter.fromAddress.fromBlock ?? 0, params.interval[0]),
-            Math.min(
-              filter.fromAddress.toBlock ?? Number.POSITIVE_INFINITY,
-              params.interval[1],
-            ),
-          ]);
-        }
+    const filterTotalIntervals = intervalIntersection(
+      [params.interval],
+      [[filter.fromBlock ?? 0, filter.toBlock ?? Number.POSITIVE_INFINITY]],
```
typedarray, line 353: Should this use getFilterFromBlock / getFilterToBlock ?

packages/core/src/sync-store/index.ts
```diff
@@ -305,22 +372,22 @@ export const createSyncStore = ({
...
         const fragments = getFragments(filter);
         result.set(filter, []);
-        for (let j = 0; j < fragments.length; j++) {
-          const fragment = fragments[j]!;
+
+        for (const fragment of fragments) {
           const intervals = rows
-            .filter((row) => row.filter === `${i}`)
-            .filter((row) => row.fragment === `${j}`)
+            .filter((row) => row.fragment === `${index}`)
```
typedarray, line 390: In the query, it seems like we increment the index on the first pass (such that the first value will be `'1'`, but here we are incrementing after the first pass.

packages/core/src/runtime/index.ts
kyscott18: No, in this function the factory and filter intervals are handled separately. `factoryTotalIntervals` is defined a few lines below.

packages/core/src/sync-store/index.ts
kyscott18: `index++` is the post-increment operator. It returns the current value, then increments.

PR #2177, opened 2025-11-12
add `ethGetLogsBlockRange` config

kyscott18:


packages/core/src/rpc/index.ts
```diff
@@ -556,6 +556,7 @@ export const createRpc = ({
 
           if (
             body.method === "eth_getLogs" &&
+            chain.eth_getLogsRangeLimit !== undefined &&
```
typedarray: Shouldn't this be `=== undefined` ? Not clear what the desired log behavior is here

.changeset/plain-animals-change.md
```diff
@@ -0,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Introduced `eth_getLogsRangeLimit` to chains config. [Read more](http://localhost:5173/docs/config/chains#eth_getlogs-range-limit).
```
typedarray: Maybe `ethGetLogsBlockRange` ? It's not really a "range limit", it's the value that will be used for every request.

PR #2184, opened 2025-11-14
fix failing simulation tests

kyscott18:
This fixes all known failing integration tests. It also improves the integration tests themselves, specifically the GitHub CI. 

There were three bugs, one in each of the ordering modes, that were causing failing tests.


.github/workflows/integration-test.yml
```diff
@@ -2,12 +2,63 @@ name: Integration Test
...
+    if: github.event_name == 'workflow_dispatch'
+    name: Integration Test Failure-Inducing
+    runs-on: self-hosted
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - app_id: assessment
+            seed: 3938fff044d4feb11b33dbdb57bae142ec6d73029611ddff14d9d4b11ada2fe6
```
typedarray: If it's possible, I think it would be helpful to add comments explaining why each of these was included, ideally with a link to an issue/PR

.github/workflows/integration-test.yml
```diff
@@ -2,12 +2,63 @@ name: Integration Test
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: false
+
 jobs:
-  integration-test:
-    name: Integration Test
+  integration-test-failure-inducing:
+    if: github.event_name == 'workflow_dispatch'
+    name: Integration Test Failure-Inducing
```
typedarray: FWIW, I think "Simulation Test" is probably a better name for what this is (from "Deterministic Simulation Testing" term). And I might call this "Simulation Test Notable Seeds" or "Simulation Test Known Failures"

PR #2210, opened 2025-12-02
fix `disable_cache: true`

### LINKED ISSUE #2209: disableCache in localhost config breaks ponder
### Version

0.15.12

### Current behavior

Upgrading from ponder 0.13.14 to 0.15.12 and ran into this error: 
```
ERROR unhandledRejection
TypeError: Cannot read properties of undefined (reading 'map')
    at file:///<~~project~~>/node_modules/ponder/src/runtime/index.ts:507:42
    at Array.map (<anonymous>)
    at getCachedBlock 
// etc
```
chains config:
```js
    chains: {
        localhost: {
            disableCache: true,
            id: 31337,
            rpc: http(process.env.PONDER_RPC_URL_31337),
        },
    },
```

Removing `disableCache` and ponder appears to run without issue.

### Expected behavior

I would expect to be able to `disableCache` on my local development node.

### Steps to reproduce

I haven't put together a simple case to repro. 

### Link to repository

_No response_

### Anything else?

_No response_

kyscott18:
Fixes a crash when using factories and `disable_cache: true`

Closes #2209

### DIFF
--- a/.changeset/floppy-goats-raise.md
+++ b/.changeset/floppy-goats-raise.md
@@ -1,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Fixed a bug that caused Ponder to crash when `disable_cache: true`.


kyscott18:
https://github.com/ponder-sh/ponder/actions/runs/19865947120

PR #2219, opened 2025-12-05
in-memory database for live indexing

kyscott18:
This pr uses the in-memory database currently used for backfill indexing, for live indexing. This is meant to be a performance improvement without any behavior changes.

Instead of processing `context.db` and `context.client` calls one by one, directly converting them into their respective database queries and RPC calls, this uses the results of previous runs to improve performance.

This makes it so that there is no difference between live and backfill indexing. Previously delayed errors were only present in backfill indexing.


kyscott18:
https://github.com/ponder-sh/ponder/actions/runs/19976082640

PR #2224, opened 2025-12-08
feat: bun support

jaylmiller:
## User changes
- Update docs
- Needed to disable the default nodeJS metrics when running bun
- Update create-ponder to set the `package.json` scripts to use `--bun` (when the user ran create-ponder with bun runtime).

## Dev changes
Most changes are to the tests to get them to work with the bun test runner. 

- Remove some uses of `toMatchInlineSnapshot`.
    - Depending on the object being asserted on, the expected inline snapshots are different, so any case where that happens we change those asserts to for example `.toMatchObject` or `.toThrowError`, depending on what's being asserted on.
 - Remove usage of `context` argument to test cases.
   - Bun test runner does not pass a mutable context to each test case, so we change the `context` to a singleton that can be imported by each test.
 - Modify `globalSetup` to work with bun too. 
    - In vite the default export of the globalSetup file is automatically run, in bun its not, so we need to call the setup function in the globalSetup.
 - Reset values set in `globalThis` between test cases.
     - the `PONDER_` values in  `globalThis` were persisting across test cases, causing some subtle bugs. For the bun tests we register a `beforeEach` and `afterEach` that resets them
 - Modify `beforeEach` hooks. In vite, when `beforeEach` functions return another function, those functions are automatically registered as cleanup funcs. In bun they are not so we need to register those to cleanup using `afterEach`
- Modify usage of `expect` when using with promises. 
    - Bun runner does not like when you pass a function that returns a promise into an `expect`, it expects just the promise.


docs/pages/docs/production/railway.mdx
```diff
@@ -35,6 +35,10 @@ yarn start --schema $RAILWAY_DEPLOYMENT_ID
 npm run start -- --schema $RAILWAY_DEPLOYMENT_ID
 ```
 
+```bash [bun]
+bun start -- --schema $RAILWAY_DEPLOYMENT_ID
```
typedarray: Worth confirming this is correct, the extra `--` is not required for pnpm or yarn, only npm

docs/pages/docs/guides/bun.mdx
```diff
@@ -0,0 +1,59 @@
...
+[Bun](https://bun.sh) is a toolkit for modern JavaScript and TypeScript applications. Bun includes a **runtime** (alternative to Node.js), a **package manager** (alternative to `npm`, `pnpm`, and `yarn`), and more.
+
+Ponder supports Bun as both a runtime and package manager.
+
+## New project
+
+To create a new project using Bun, just use `bun --bun` to run the `create-ponder` CLI tool. [Read more](/docs/api-reference/ponder/cli). This will create a new project directoty using the Bun package manager and runtime.
+
+```bash [shell]
+bun --bun create ponder {...options}
```
typedarray, line 12: Worth double checking that this correctly uses Bun as the package manager in the created project (e.g. creates a `bun.lock` file)

.changeset/tiny-turtles-post.md
```diff
@@ -0,0 +1,6 @@
+---
+"create-ponder": patch
+"ponder": patch
+---
+
+Support bun
```
typedarray: Let's do this in the 0.16 minor

docs/pages/docs/guides/bun.mdx
jaylmiller: yup it works
Screenshot 2025-12-09 at 4 07 53 PM

docs/pages/docs/production/railway.mdx
jaylmiller: oh oops your right it works like pnpm does

.changeset/tiny-turtles-post.md
jaylmiller: So just change both lines to `minor` instead of `patch`?

typedarray: Yes

packages/core/src/_test/globalSetup.ts
```diff
@@ -5,6 +5,7 @@ import { startProxy } from "@viem/anvil";
 import dotenv from "dotenv";
 import { execa } from "execa";
 import { Pool } from "pg";
+import { IS_BUN_TEST } from "./utils.js";
```
jaylmiller: i think you can't import anything that imports vitest in globalSetup

kyscott18 (approved):

PR #2228, opened 2025-12-11
improve schema validation and database object names

kyscott18:
This adds more validations to `ponder.schema.ts` and `DATABASE_SCHEMA`. It changes the names of several database objects, including the live query channel, in order to avoid names that are longer than the max length in postgres (63 characters).

### DIFF
--- a/packages/core/src/build/schema.ts
+++ b/packages/core/src/build/schema.ts
@@ -176,8 +176,6 @@ export const buildSchema = ({
         }
       }
 
-      // Note: Ponder lets postgres handle the table name length limit and truncation.
-
       if (tableNames.has(getTableName(s))) {
         throw new Error(
           `Schema validation failed: table name '${getTableName(s)}' is used multiple times.`,


packages/core/src/build/schema.ts
```diff
@@ -129,6 +159,8 @@ export const buildSchema = ({
         }
       }
 
+      // Note: Ponder lets postgres handle the table name length limit and truncation.
```
typedarray: What about when we append the `_reorg...` prefix - are we still vulnerable to conflicts after truncation?

I figured that we would need to manually enforce an identifier limit that's less than 63 characters, to leave room for prefixes/suffixes.

packages/core/src/database/actions.ts
```diff
@@ -233,7 +236,7 @@ CREATE OR REPLACE FUNCTION "${schema}".${procedure}
 RETURNS TRIGGER LANGUAGE plpgsql
 AS $$
 BEGIN
-  INSERT INTO live_query_tables (table_name)
+  INSERT INTO ${getLiveQueryTempTableName()} (table_name)
```
typedarray, line 239: Does "live_query_tables" also need to be a reserved table name?

kyscott18: No - it's a temp table and shadows other tables with the same name.

packages/core/src/build/schema.ts
kyscott18: When we append the `_reorg` prefix it's still possible to have a table name collision. If this happened there would be a non-retryable error emitted in the `migrate` function. 

We could enforce a limit that is less than 63 characters but I thought it might cause more harm than good when interfacing with external systems.

typedarray: IMO better to validate explicitly / fail fast. How short would we need to make it to leave room for all our possible prefixes and suffixes?

PR #2231, opened 2025-12-11
fix: execute indexing files in series during build to avoid hang

jaylmiller:


packages/core/src/build/index.ts
```diff
@@ -248,10 +249,19 @@ export const createBuild = async ({
...
-          file,
-        })),
+        files.map((file) =>
+          Promise.race([
+            executeFile({ file }).then((res) => ({
+              ...res,
+              file,
+            })),
+            setTimeout(10_000).then(() => ({
+              error: new Error("File execution did not complete (waited 10s)"),
```
kyscott18: Do you think this should be a retryable or non-retryable error?

packages/core/src/build/index.ts
```diff
@@ -248,10 +249,19 @@ export const createBuild = async ({
...
-        files.map(async (file) => ({
-          ...(await executeFile({ file })),
-          file,
-        })),
+        files.map((file) =>
+          Promise.race([
+            executeFile({ file }).then((res) => ({
+              ...res,
+              file,
+            })),
```
kyscott18: Just noting that the execute file doesn't actually stop executing when it "loses" the race. I don't think it matters in this case because we immediately crash

packages/core/src/build/index.ts
jaylmiller: Non-retry

PR #2237, opened 2025-12-29
docs: clarify invariant comment in indexing-store utils

Olexandr88:
Clarify an invariant comment by replacing a misleading TODO with a NOTE


kyscott18:
I don't think this is providing much value, going to close.

PR #2244, opened 2026-01-07
fix: numeric type graphql

jean-neiverth:
### Issue

This PR is related to columns created with `numeric` type, like in this example:

```
export const table = onchainTable(
  "table",
  (t) => ({
    //...
    numericColumn: t.numeric().notNull(),
    //...
```

On the GraphQL interface, the numeric filters don't exist for `numericColumn` (like `numericColumn_gt`, `numericColumn_lt`, etc.), so it's not possible to query the column properly

### Solution

This PR fix this issue, creating a new `GraphQLNumeric` type structure, similar to how it's done with BigInt

### DIFF
--- a/.changeset/brown-pants-care.md
+++ b/.changeset/brown-pants-care.md
@@ -2,4 +2,4 @@
 "ponder": patch
 ---
 
-Fixed GraphQL filtering for `int8()` and `numeric()` column types.
+Fixed GraphQL filtering for `int8()` and `numeric()` column types. Previously these column types were using string filter conditions (contains, starts_with, ends_with), now they use numeric filter conditions (gt, gte, lt, lte).


packages/core/src/graphql/index.ts
```diff
@@ -677,6 +678,22 @@ const GraphQLBigInt = new GraphQLScalarType({
   },
 });
 
+const GraphQLNumeric = new GraphQLScalarType({
+  name: "Numeric",
+  serialize: (value) => String(value),
+  parseValue: (value) => {
+    return BigInt(value as any);
```
kyscott18: @typedarray This is the part I'm not sure of, what does changing the return type of this function do?

kyscott18:
I also updated the graphql types for `int8()` columns, which were also incorrect.

kyscott18 (approved):

packages/core/src/graphql/index.ts
```diff
@@ -686,7 +687,10 @@ const columnToGraphQLCore = (
   column: Column,
   enumTypes: Record<string, GraphQLEnumType>,
 ): GraphQLOutputType => {
-  if (column.columnType === "PgEvmBigint") {
+  if (
+    column.columnType === "PgEvmBigint" ||
+    column.columnType === "PgNumeric"
+  ) {
     return GraphQLBigInt;
```
jean-neiverth, line 694: I think it maybe doesn't make sense to attach GraphQLBigInt to "PgNumeric", once "PgNumeric" values can have decimals (i.e. "1.234"). So conversion to BigInt may cause a high information loss or even fail. Wdyt?

PR #2250, opened 2026-01-12
chore: add package links in README.md

gap-editor:
## Add package links to README

Added npm links for all packages in the Packages section to make it easier for users to learn about and navigate to each package.

### Changes
- Added npm package links for all 6 packages

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,12 +165,12 @@ If you're interested in contributing to Ponder, please read the [contribution gu
 
 ## Packages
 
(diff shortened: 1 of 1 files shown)


gap-editor:
maybe we need to add short descriptions to every package? let me know

kyscott18:
Thanks for adding this. I've created #2280 which extends this change to update all 4 README files because I wasn't able to make changes to your branch.

PR #2269, opened 2026-02-16
feat(graphql): add support for custom plugins

jrwbabylonlab:


kyscott18:
This is outside the scope of what we're working on for Ponder at the moment, but feel free to build it out in a fork

PR #2277, opened 2026-03-09
Add O(1) detection of new indexed data for downstream systems  

areinel:
## Reasoning                                                                                                                                                                         
                                                            
  Downstream systems need a way to detect when new indexed data is available without diffing checkpoints. A monotonically increasing integer on `_ponder_checkpoint` lets them do this 
  with a single O(1) query — just compare the stored value against the last seen value. It also bumps on reorgs so that even if a reorg replays zero events, downstream knows the data
  changed.                                                                                                                                                                             
                                                            
  ## Changes

  - Added `data_checkpoint` integer column to the `_ponder_checkpoint` schema
  - Historical indexing: accumulates `events.length` into `data_checkpoint` via upsert
  - Realtime indexing: increments `data_checkpoint` by `events.length` on each block update
  - Reorg handlers: bumps `data_checkpoint` by +1 in multichain, isolated, and omnichain runtimes
  - Changeset added for patch release

  ## Docs

  - Expanded the Checkpoint table section in the database reference with full column schema
  - Added usage example for polling `data_checkpoint` from downstream systems

  ## Tests

  - Column exists after migration
  - Defaults to 0 when omitted
  - Accumulates correctly on upsert conflict
  - Increments correctly on update
  - Independent per chain
  - Reorg bump pattern works end-to-end

### DIFF
--- a/docs/pages/docs/api-reference/ponder/database.mdx
+++ b/docs/pages/docs/api-reference/ponder/database.mdx
@@ -67,9 +67,30 @@ style Z2 fill:#f00,color:#fff
 
 The `_ponder_meta` table tracks the state of an instance running in a specific database schema. It stores the `build_id`, indexing checkpoint, and table names of the current instance, and uses a heartbeat locking mechanism to support the lifecycle logic described above.
 
-### Checkpoint Table
+### Checkpoint table
 
-The `_ponder_checkpoint` table tracks the indexing progress of each chain.
+The `_ponder_checkpoint` table tracks the indexing progress of each chain. It contains one row per chain.
+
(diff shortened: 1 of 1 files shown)


kyscott18:
We are not interested in adding this feature to ponder at the moment. Feel free to create a fork and maintain your feature there.

brandonaaskov:
@kyscott18 did you even review this PR? This is something that would be nice to have. As it stands, there's no way to "watch" for new data. This is a very small change to your codebase, but it's a very impactful one. I think you should take another look at it. Personally, I want this as well, and I don't want to have to chase down a random fork for a feature I would have expected to be in Ponder already.

PR #2282, opened 2026-03-11
fix: add Filecoin support, including handling of null rounds

rvagg:
Filecoin has intentional occasional null epochs where there is no tipset (~block) produced, but the chain isn't interrupted. Because ponder assumes contiguous block numbers, it fails to index Filecoin when it encounters one of these null rounds. The fixes in here add Filecoin chain support and also deal with all of the places where we need special handling of non-contiguous block numbers.


kyscott18:
This isn't something we're planning to add to Ponder right now, but you're welcome to implement it in a fork

PR #2283, opened 2026-03-11
add low-level `location` field to factory config

### LINKED ISSUE #2275: Feature Request: Optional custom byte offset override for `factory()` config
### Problem / use case

We are building an indexer for a protocol that uses generic cross-chain messaging (like LayerZero). Our dynamic child contracts are tracked via a single event that looks like this: `MessageReceived(bytes32 guid, bytes message)`.

The actual child address we need to track is packed deep inside that dynamic `bytes message` payload. Because Ponder's engine relies on strictly calculated 32-byte ABI offsets to slice strings during the bulk historical sync phase, it is currently impossible to point the `factory()` config at our nested address.

For example, our target address sits exactly at byte 100 in the unindexed log data. If we try to spoof the ABI by passing a fake parameter list, Ponder's `buildLogFactory` strictly pads everything to 32 bytes, meaning it will calculate an offset of 96 or 128, completely missing our address.

Currently, the only ways to work around this are to either deploy an on-chain "translator" contract just for the indexer, or to physically monkey-patch `@ponder/core` (specifically `buildLogFactory`) to hardcode our magic number into the offset calculation.

### Proposed solution

It would be incredibly helpful to have an optional `offset` property in the `factory()` configuration. If provided, Ponder would skip the standard ABI offset calculation and just use this literal byte offset to slice the raw JSON-RPC log data.

Here is an example of how the API could look:

```typescript
contracts: {
  TargetContract: {
    abi: TargetAbi,
    address: factory({
      address: "0xYourFactory...",
      // The actual event emitted by the contract
      event: parseAbiItem("event MessageReceived(bytes32 guid, bytes message)"),
      parameter: "message",
      
      // Proposed API: Override the strict 32-byte calculation
      // Force Ponder to slice exactly at byte 100 in the unindexed data
      offset: 100 
    }),
  }
}

```

---
The two alternatives we are using or considering are highly friction-heavy:

1. Using `pnpm patch` to physically modify `node_modules/@ponder/core/dist/build/factory.js` so it forces `offset = 100` when it sees our specific event selector.
2. Deploying a low-gas translator contract on-chain that listens for these LayerZero messages, decodes the bytes, and emits a standard, cleanly padded event.

I know Ponder deliberately bypasses standard JS/viem decoding during the factory sync to maximize bulk sync speed using native string slicing (e.g., `offset100`). Allowing us to optionally pass our own integer directly into that string-slicing engine would solve this problem perfectly without requiring any structural changes or performance hits to the sync architecture!

kyscott18:
This adds support for `location` as a low-level alternative to `parameter` in factory config.

Closes #2275 

`sync-historical` and `sync-realtime` already handle invalid child address decoding.
https://github.com/ponder-sh/ponder/blob/main/packages/core/src/sync-realtime/index.ts#L479-L482
https://github.com/ponder-sh/ponder/blob/main/packages/core/src/sync-historical/index.ts#L385-L388

### DIFF
--- a/docs/pages/docs/api-reference/ponder/config.mdx
+++ b/docs/pages/docs/api-reference/ponder/config.mdx
@@ -283,7 +283,7 @@ Specifies a list of addresses collected from decoded event logs. Both [`contract
 | **address**   |     `0x{string} \| 0x{string}[] \| undefined`        | **Optional**. Address of the factory contract that creates instances of this contract.     |
 | **event**     | [`AbiEvent`](https://abitype.dev/api/types#abievent) | **Required**. ABI item of the event that announces the creation of a new child contract.   |
 | **parameter** |                       `string`                       | **Optional**. Name of the parameter within `event` that contains child contract addresses. |
-| **location**  |       `topic${1 | 2 | 3}` | `offset${number}`        | **Optional**. Low-level alternative to `parameter`. Specifies the exact location (e.g. topic or offset) of the child contract address in `event`. Most users should prefer `parameter` instead. |
+| **location**  |       `topic{1\|2\|3} \| offset{number}`        | **Optional**. Low-level alternative to `parameter`. Specifies the exact location (e.g. topic or offset) of the child contract address in `event`. Most users should prefer `parameter` instead. |
 
 ```ts [ponder.config.ts] {8-14}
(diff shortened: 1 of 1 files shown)


typedarray (approved):

james-a-morris:
Thanks gents 👌

PR #2292, opened 2026-03-18
feat: add context.isRealtime to indexing handlers

filo87:
## feat: add `context.isRealtime` to indexing handlers

Indexing functions now receive **`context.isRealtime: boolean`**:

- **`false`** — event is handled during **historical backfill** or in a **`setup`** handler  
- **`true`** — event is handled by the **live** indexer  

Useful when behavior should differ between backfill and live (e.g. webhooks only on new blocks). Documented that **`true` does not mean “caught up to chain tip”** (live can still lag).

**Implementation:** set on the shared handler context from `processHistoricalEvents` / `processRealtimeEvents` (and `false` for setup). Types updated in `Virtual.Context`; API docs, indexing overview, migration guide, and changelog updated; tests assert historical vs realtime.

**Other:** typing/build fixes for the historical path (`Event` assertion for proxied `event`, safe `eventCount` updates under `noUncheckedIndexedAccess`).


filo87:
@kyscott18 proposal for a feature that would be super helpful for our use case!

PR #2293, opened 2026-03-25
fix: Make pglite in memory work

krzkaczor:
`createPglite()` was mutating options.dataDir to undefined for memory://, which broke downstream code that still referenced the original object. Basically Pglite API expects passing "memory://" string.


kyscott18:
Replaced by #2336

PR #2298, opened 2026-04-27
chore: upgrade to ENSv2 (Universal Resolver)

Dhaiwat10:
Hey! I'm Dhaiwat from DevRel at ENS Labs — getting libraries across the ecosystem ready for ENSv2.

This PR tightens the viem peer dep in `packages/core` and `packages/utils` from `>=2` to `>=2.35` (first viem with the Universal Resolver). Without this, downstream indexers can install old viem and silently use legacy ENS resolution — Ponder exposes viem's ENS getters to user-written indexing code.

Bumps pinned viem in examples, templates, and root from `^2.21.3` (and `2.30.1` at root) to `2.35.0` to match the new floor. **Note:** newer 2.x (tested 2.48.4) breaks Ponder's typecheck.

Verification: `ur.integration-tests.eth` → `0x2222222222222222222222222222222222222222` (legacy resolution incorrectly returns `0x1111111111111111111111111111111111111111`).

More: https://docs.ens.domains/web/ensv2-readiness

### DIFF
--- a/examples/feature-accounts/package.json
+++ b/examples/feature-accounts/package.json
@@ -14,7 +14,7 @@
   "dependencies": {
     "ponder": "workspace:*",
     "hono": "^4.5.0",
-    "viem": "^2.21.3"
+    "viem": "^2.35.0"
   },
   "devDependencies": {
     "@types/node": "^20.10.0",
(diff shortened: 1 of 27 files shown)


kyscott18:
Since `viem` is a peer dependency, it's actually up to the user to install the correct version. Ponder uses the version that the user has installed.

We can make this change but will wait for a minor version of Ponder

kyscott18:
Replaced by #2333

PR #2299, opened 2026-04-28
fix: include public schema in PGLite search_path for raw SQL queries

### LINKED ISSUE #2235: PGLite search_path doesn't include public schema by default
# PGLite `search_path` Issue in Ponder Dev Mode

## Summary

Raw SQL queries return 0 rows in ponder dev mode (PGLite) while Drizzle ORM queries work correctly. This affects `@ponder/client` SQL HTTP endpoints.

## Environment

- ponder: 0.15.17
- Mode: dev (PGLite)

## What works

- Drizzle ORM queries via `db.select().from(schema.table)`
- Production with Postgres

## What fails

- Raw SQL via `db.execute(sql`SELECT * FROM "TableName"`)` returns 0 rows
- `@ponder/client` SQL HTTP endpoints return 0 rows

## Root cause

PGLite's default `search_path` doesn't resolve unqualified table names to the `public` schema.

Drizzle ORM internally uses schema-qualified names, so it works. Raw SQL doesn't, so it fails silently.

## Reproduction

```ts
// src/api/index.ts
import { db } from 'ponder:api'
import schema from 'ponder:schema'
import { count, sql } from 'ponder'

// This works - returns correct count
const drizzleResult = await db.select({ count: count() }).from(schema.myTable)

// This fails - returns 0
const rawResult = await db.execute(sql`SELECT COUNT(*) FROM "MyTable"`)

// This works - explicit schema
const fixedResult = await db.execute(sql`SELECT COUNT(*) FROM public."MyTable"`)
```

## Workaround

Use schema-qualified table names in raw SQL:

```sql
SELECT * FROM public."TableName"
```

## Why it's confusing

1. Drizzle queries work fine
2. Raw SQL fails silently (empty results, no error)
3. Same code works in production with Postgres
4. `@ponder/client` uses raw SQL internally, so all client queries returns empty

## Suggested fix

Set `search_path = public` when initializing the PGLite connection in ponder dev mode.

tenderdeve:
## Summary
- Raw SQL queries via `db.execute()` return 0 rows in dev mode (PGLite) while Drizzle ORM queries work correctly
- **Root cause:** The `SET search_path` only included the namespace schema, excluding `public`. Drizzle ORM uses schema-qualified names internally so it worked, but raw SQL with unqualified table names failed silently.
- **Fix (two parts):**
  1. Include `public` alongside the namespace schema in the in-migration `SET search_path` when the namespace is not already `public`
  2. Set `search_path` persistently after migration via PGLite's `exec()` method (analogous to how the Postgres driver sets it per-connection via `createReadonlyPool`)

**File:** `packages/core/src/database/index.ts`

Fixes #2235

## Test plan
- [ ] Run ponder in dev mode (PGLite), execute raw SQL via `db.execute(sql\`SELECT * FROM "TableName"\`)` — should return rows
- [ ] Verify Drizzle ORM queries still work
- [ ] Verify production Postgres mode is unaffected

### DIFF
--- a/packages/core/src/database/index.ts
+++ b/packages/core/src/database/index.ts
@@ -646,7 +646,12 @@ CREATE TABLE IF NOT EXISTS "${namespace.schema}"."${PONDER_CHECKPOINT_TABLE_NAME
 
           if (dialect === "pglite" || dialect === "pglite_test") {
             await tx.wrap(
-              (tx) => tx.execute(`SET search_path TO "${namespace.schema}"`),
+              (tx) =>
+                tx.execute(
+                  namespace.schema === "public"
+                    ? `SET search_path TO "public"`
+                    : `SET search_path TO "${namespace.schema}", "public"`,
+                ),
               context,
             );
           }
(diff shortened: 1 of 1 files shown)


kyscott18:
I was not able to reproduced the described bug. If you are able to add a failing test case along with this PR, please re-open.

PR #2305, opened 2026-06-18
Add experimental range-scan realtime sync to reduce RPC usage on L2 chains

damianmarti:
## Summary

  Adds an opt-in `experimentalRangeScan` chain config option that lets a larger
  `pollingInterval` actually reduce realtime RPC usage.

  Today, setting `pollingInterval` higher than a chain's block time does not reduce
  RPC usage because the realtime engine fetches **every** block to check for reorgs
  (documented limitation in `config/chains`). On L2s with sub-second block times this
  makes realtime indexing very RPC-expensive.

  ## Motivation

  On high-throughput L2s, a `pollingInterval` of e.g. 4s still triggers ~16
  `eth_getBlockByNumber` calls (plus per-block `eth_getLogs`) per poll, so raising it
  buys nothing. This change makes a larger `pollingInterval` translate into
  proportionally lower RPC usage.

  ## How it works

  When `experimentalRangeScan` is enabled, on each poll the realtime engine:

  1. Fetches the head **header only** (`eth_getBlockByNumber(["latest", false])`).
  2. Scans the entire unfinalized window with a single ranged `eth_getLogs([from, to])`
     instead of fetching every block.
  3. Fetches full block data only for blocks that contain matching events
     (`eth_getBlockByHash`), and emits an empty head event so the realtime checkpoint
     still advances to the tip.

  This makes per-poll RPC cost roughly constant (~2 calls + one per event-bearing
  block) regardless of how many blocks elapsed.

  Reorgs are detected by diffing the block hashes of previously-emitted event-bearing
  blocks against the rescan (a changed or disappeared hash signals a reorg). The
  existing `block` / `reorg` / `finalize` event shapes are reused, so the downstream
  indexing pipeline is unchanged.

  ## Scope

  To keep this a safe, incremental rollout, the fast path is gated:

  - It is **opt-in** via the `experimentalRangeScan` flag (default `false`); existing
    behavior is untouched.
  - It only applies to chains whose indexed sources are **exclusively non-factory
    `log` events**. Chains using factory, block, transaction, trace, or transfer
    sources log a warning and fall back to the default per-block realtime sync.

  ## Tests

  Added realtime sync tests:

  - RPC efficiency: asserts a single ranged `eth_getLogs`, zero per-block
    `eth_getBlockByNumber`, and only event-bearing blocks fetched by hash.
  - Reorg detection of an event-bearing block (via Anvil snapshot/revert).
  - Fallback to the per-block path when an unsupported (block) filter is present.

  Full realtime suite passes (19/19), plus the config-build and rpc suites.

  ## Docs

  Updated `config/chains` and the config API reference to document the new option and
  its limitations, and added a changeset.

### DIFF
--- a/.changeset/experimental-range-scan.md
+++ b/.changeset/experimental-range-scan.md
@@ -1,0 +1,5 @@
+---
+"ponder": minor
+---
+
+Added an experimental `experimentalRangeScan` chain config option. When enabled, the realtime indexing engine scans each polling interval with a single ranged `eth_getLogs` request instead of fetching every block, so a `pollingInterval` larger than the chain's block time reduces RPC usage. Useful for high-throughput L2 chains. Only applies to chains whose indexed sources are exclusively non-factory `log` events.
--- a/docs/pages/docs/api-reference/ponder/config.mdx
+++ b/docs/pages/docs/api-reference/ponder/config.mdx
@@ -158,6 +158,7 @@ export default createConfig({
 | **pollingInterval**       |  `number`                          | **Default: `1_000`**. Frequency (in ms) used when polling for new events on this chain.                                                                                         |
 | **disableCache**          |  `boolean`                         | **Default: `false`**. Disables the RPC request cache. Use when indexing a [local node](/docs/guides/foundry) like Anvil.                                                        |
 | **ethGetLogsBlockRange** |  `number`                          | **Default: `undefined`**. Maximum block range for `eth_getLogs` requests. If undefined, Ponder will attempt to determine the block range automatically based on error messages. |
+| **experimentalRangeScan** |  `boolean`                         | **Default: `false`**. Experimental. Reduce realtime RPC usage by scanning each polling interval with a single ranged `eth_getLogs` request. Only applies to chains with exclusively non-factory `log` sources. |
 
 ### `contracts`
 
--- a/docs/pages/docs/config/chains.mdx
+++ b/docs/pages/docs/config/chains.mdx
@@ -201,6 +201,8 @@ export default createConfig({
 
 This option only applies to chains whose indexed sources are exclusively `log` events with a static `address`. Chains that use factory, block, transaction, trace, or transfer sources — or a `log` source that matches every address — fall back to the default per-block realtime sync.
 
+Each scan costs about two requests no matter how many blocks elapsed, so this option only reduces RPC usage when `pollingInterval` is several times the chain's block time — roughly 4x or more. Below that it can cost *more* than the default per-block sync, which is why it is not useful on chains with slow blocks. Ponder logs a warning if it observes too few blocks per poll.
+
 This option also ignores the [`ws`](#websocket) option. A `newHeads` subscription delivers every block, which would run one scan per block and cost more than the default per-block sync. Polling bounds the scan rate to `pollingInterval`.
 
 :::warning
--- a/docs/pages/docs/config/chains.mdx
+++ b/docs/pages/docs/config/chains.mdx
@@ -161,7 +161,9 @@ export default createConfig({
 
 The `pollingInterval` option controls how frequently (in milliseconds) the indexing engine checks for a new block in realtime. The default is `1000` (1 second).
 
-If you set `pollingInterval` greater than the chain's block time, it **does not reduce RPC usage**. The indexing engine still fetches every block to check for reorgs. The default is suitable for most chains.
+By default, setting `pollingInterval` greater than the chain's block time **does not reduce RPC usage**. The indexing engine still fetches every block to check for reorgs. The default is suitable for most chains.
+
+To reduce RPC usage on high-throughput chains (e.g. L2s with sub-second block times), enable [`experimentalRangeScan`](#range-scan-experimental) alongside a larger `pollingInterval`.
 
 ```ts [ponder.config.ts]
 import { createConfig } from "ponder";
@@ -176,6 +178,32 @@ export default createConfig({
   },
 });
 ```
+
+## Range scan (experimental)
+
+The `experimentalRangeScan` option reduces realtime RPC usage by scanning each polling interval with a single ranged `eth_getLogs` request instead of fetching every block. When enabled, a `pollingInterval` larger than the chain's block time reduces RPC usage proportionally — useful for high-throughput L2 chains.
+
+```ts [ponder.config.ts]
+import { createConfig } from "ponder";
+
+export default createConfig({
+  chains: {
+    base: {
+      id: 8453,
+      rpc: process.env.PONDER_RPC_URL_8453,
+      pollingInterval: 5_000, // 5 seconds [!code focus]
+      experimentalRangeScan: true, // [!code focus]
+    },
+  },
+  contracts: { /* ... */ },
+});
+```
+
+This option only applies to chains whose indexed sources are exclusively non-factory `log` events. Chains that use factory, block, transaction, trace, or transfer sources fall back to the default per-block realtime sync.
+
+:::warning
+This option is experimental. It changes how reorgs are detected in realtime (by diffing the block hashes of event-bearing blocks rather than tracing a contiguous chain of block hashes). The default is `false`.
+:::
 
 ## Disable caching
 
--- a/docs/pages/docs/config/chains.mdx
+++ b/docs/pages/docs/config/chains.mdx
@@ -201,8 +201,10 @@ export default createConfig({
 
 This option only applies to chains whose indexed sources are exclusively non-factory `log` events. Chains that use factory, block, transaction, trace, or transfer sources fall back to the default per-block realtime sync.
 
+This option also ignores the [`ws`](#websocket) option. A `newHeads` subscription delivers every block, which would run one scan per block and cost more than the default per-block sync. Polling bounds the scan rate to `pollingInterval`.
+
 :::warning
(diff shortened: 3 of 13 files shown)


kyscott18:
I agree that Ponder's current stance (fetch every block) doesn't cover all use cases. It's expensive in terms of RPC credits. A much earlier version of Ponder had implemented an `eth_getLogs` first version of live indexing similar to what you're proposing, and used heuristics to switch between that version and the baseline `eth_getBlockByNumber` version. It led to many production reorg detection bugs + missed events and we ended up removing it.

However, since then we've greatly expanded our testing ability (primarily through the simulation tests) so I'm open to revisiting. Can you say more about why the first request is `eth_getBlockByNumber`? Does this method rely on the fact that the entire finalized range can be fetched in one `eth_getLogs`? What if it can't?

Can you use the Alchemy compute unit costs per method to do some numerical analysis comparing the cost of the two algorithms accounting for:
- polling interval
- blocks per second
- p(log per block), probability of a follow up `eth_getBlockByNumber`

Are you running this fork in production?

damianmarti:
You were right to be skeptical. I wired this path into the simulation harness, and it found two bugs that my unit tests — including ones written specifically for missed events — passed straight through. Both are fixed. Details below, along with the numbers you asked for.

## Simulation results

Wired `EXPERIMENTAL_RANGE_SCAN` into `SIM_PARAMS` and read it in the two log-only apps (`reference-erc20`, `basepaint`). `create-app.ts` never sets it, so the expected tables always come from the default per-block sync — the comparison is meaningful. A `FORCE_RANGE_SCAN` override pins the parameter so a failing seed can be replayed against both strategies, which is how I attributed each failure.

Adding a parameter doesn't disturb existing seeds: `pick` keys its generator on the parameter name, so every other draw is unchanged and the pinned failure seeds in `simulation-test.yml` still reproduce.

**12/12 seeds pass** on `reference-erc20`, all pre-filtered to actually exercise the realtime path:

| | range |
|---|---|
| unfinalized window | 50 – 300 blocks |
| reorg rate | 0.02 – 0.05 |
| deep reorgs | enabled on 8 of 12 |
| uncached blocks | 0 – 100 |
| reorgs actually injected | 2 – 5 per run |

One seed timed out on its first run and passes on retry; its signature is the rpc-layer livelock described at the bottom, which I also hit on the default per-block path.

Two seeds reproduce specific bugs against the pre-fix code, if you want to see them fail:

- `SEED=877d0dc2dca1eb8d9f3ae1ef9869688f6ddf843cde2c6c8f24ca5cb2531dc644` — drops the final block's events
- `SEED=7d4d91b0768232efb392699a8336de38c077e26f5486aefa26ab1e842aa22201` — poisons the block cache from the finalize path

## What it found

**1. `blockCallback` fired before events were indexed.** `reconcileBlock` attaches `blockCallback` to the block event it emits, and the runtime calls it only after that block's events are indexed. `reconcileRange` called it directly at the end of reconciliation, so the head was reported as processed while its events were still pending. When the chain reached its end block, the runtime completed the chain and dropped them. In the failing seed, the final block's two events were written to `ponder_sync` but never indexed — two accounts missing, a balance short by 20000e18. The same seed with the range scan disabled passes.

**2. Finalization fetched the boundary block.** The per-block path takes the finalized block from its contiguous `unfinalizedBlocks`; mine is sparse, so I fetched the boundary block directly — as a header. Now it finalizes to the highest local block at or below the boundary instead. Finalizing short is always safe, and a head block is appended every poll, so the window stays bounded. This also removes a request per finalization.

Two more found by review before that: a reorg that *introduces* logs into a previously event-free block went undetected (the local tip advances past those blocks, so the new logs were never ingested and `finalize` then cached the interval as synced), and `ws` made the scan run per-block rather than per-poll — strictly worse than the default.

Four defects total; two of them only visible under simulation. That's a much stronger argument for your position than anything I'd have written otherwise.

## Why the first request is `eth_getBlockByNumber`

It's now the same request the default path already makes, `["latest", true]`. The head gives a stable `toBlock` for the scan — `toBlock: "latest"` races across chunked requests and leaves no hash to anchor to — plus the hash/parentHash/timestamp needed to advance the realtime checkpoint.

I originally optimized this to a header-only request and have removed that. It saved no compute units (Alchemy prices `eth_getBlockByNumber` identically either way), it needed a `LightBlock` → `SyncBlockHeader` cast, and it broke the simulation harness. The range-scan path only reads four fields from the head, so a full block serves it.

## Does it rely on one `eth_getLogs` for the whole range?

It did, and that was the weakest part. The window has to be the full unfinalized range (`[finalized + 1, head]`) for correctness, since a reorg can introduce logs anywhere in it — and that window grows to `2 * finalityBlockCount` before finalization, so 480 blocks on Arbitrum. The original version chunked only by a statically configured `ethGetLogsBlockRange` and retried the same failing request every poll.

Now, borrowing from `syncLogsDynamic`:

- rejected requests are split into the ranges the error suggests via `getLogsRetryHelper`, recursively, and the narrower range is remembered for later polls (and grown back by 5% per clean scan when it came from an error that didn't state a limit)
- addresses batch in groups of 50
- the request is constrained to the union of every filter's `address` **and** `topic0`; the union is a superset of each filter, so client-side filtering stays exact
- logs are matched against the filters *before* fetching blocks by hash, so a block that only matches the address union costs nothing
- a log source matching every address falls back to the per-block sync rather than scanning the whole chain each poll

## Cost analysis

Alchemy's current table (`eth_getBlockByNumber` 20, `eth_getBlockByHash` 20, `eth_getLogs` 60). Let `b` = blocks/sec, `P` = polling interval, `p` = P(block has a matching log), `q` = P(bloom matches a filter; `q ≥ p`), `R` = getLogs requests per scan.

```
per-block CU/s  = 20b + 60bq            (independent of P — every block is fetched)
range scan CU/s = (20 + 60R)/P + 20bp
```

The per-block path's second term is driven by `q`, not `p`, because of the `logsBloom` check in `fetchBlockEventData` — that skip is load-bearing for this comparison, and without it the default path would cost two requests per block.

Arbitrum One (b = 4, R = 1), CU/hour:

| p = q | per-block | P=2s | P=4s | P=12s |
|---|---|---|---|---|
| 0.001 | 288,900 | 144,300 (2.0×) | 72,300 (4.0×) | 24,300 (11.9×) |
| 0.05 | 331,200 | 158,400 (2.1×) | 86,400 (3.8×) | 38,400 (8.6×) |
| 0.5 | 720,000 | 288,000 (2.5×) | 216,000 (3.3×) | 168,000 (4.3×) |
| 1.0 | 1,152,000 | 432,000 (2.7×) | 360,000 (3.2×) | 312,000 (3.7×) |

Break-even is **P ≈ 4 × blockTime** at low density and R = 1 — 1s on Arbitrum, 2s on Base, ~48s on mainnet. Savings survive even at p = 1, because the per-block `eth_getLogs` disappears; the floor is `20bp`.

The counter-case matters: on mainnet at P = 12s the range scan is ~4× *worse* (24,000 vs ~6,200 CU/hr). Break-even also moves with `R` — at three getLogs requests per scan it's closer to ten blocks per poll. Rather than assume either, Ponder now measures blocks, requests and matched blocks over the first ten polls and warns once if the scan is costing more than the default would.

Caveat: Alchemy's 60 CU for `eth_getLogs` is flat regardless of range or response size. Providers that price by bandwidth, and Alchemy's own throughput limits, will penalize a wide window scan — which is why the address/topic constraints above matter.

## Production

No. I've been running it against a local Arbitrum indexer (12s polling, a handful of static-address contracts) and held off on production pending review. Given that the simulation then found two event-dropping bugs, that was the right call — and I'd not recommend anyone run this until you're satisfied with the testing.

Happy to run it in production for a week and report before/after credits from the Alchemy dashboard if that's useful.

## Two things worth knowing independently of this PR

**The harness's block cache doesn't distinguish header from full responses.** It keys `eth_getBlock` by number and hash, so a `[n, false]` response is later served to a `[n, true]` request, and validation rejects it with `'transaction.hash' is a required property`. Ponder requests headers legitimately at startup (`runtime/index.ts:135-172`), so the hazard exists today; my path just made it likely by requesting those blocks in full later. I purge those rows between local runs. Happy to send a small harness fix separately — keying on `includeTransactions`, or not caching header responses.

**A livelock when all providers go inactive.** With a single provider and sustained errors (real 429s or the harness's simulated ones), the process stops making progress and spins at ~100% CPU without recovering. I hit this on the **default** per-block path while building the ground-truth tables, so it's not specific to this change. Happy to open an issue with the repro.

## What else would you want to see?

The obvious gaps: I've only run `reference-erc20`, and `basepaint` is wired but unrun. More seeds, a longer sweep, or specific configurations you'd trust more — tell me what would move this and I'll run it.

kyscott18:
Thanks for the descriptive response. 

> Why the first request is eth_getBlockByNumber

Makes sense

> Does it rely on one eth_getLogs for the whole range?

I think there's still an issue if the chain reorgs between the first and second request of a split `eth_getLogs` request. Or would this eventually be resolved when the next polling interval runs and finds mismatched logs?

> Cost analysis

This is great and demonstrates the value of this change.

> What else would you want to see?

We'd need to update the simulation test to make sure it's able to run this branch for all apps. We may also need to introduce new forms of reorgs (in `eth_getLogs` responses) to make sure the new codepaths are robust.

A change this big to essentially the most important and historically difficult part of Ponder is going to require a lot of thinking and will take more time. However, you can run this with your Ponder apps by patching in this change (supported by most package managers https://pnpm.io/cli/patch). It's how we test changes to Ponder like this internally.

PR #2306, opened 2026-06-23
[ClickHouse sink 1/6] Add durable finalized sink delivery

neuroborus:
Stack: Part 1/6. See the stack [discussion](https://github.com/ponder-sh/ponder/discussions/2307). Base of the ladder.

**What**

Add a generic, finalized-only analytics sink contract and a durable Postgres outbox.

**How**

- Add optional sink configuration with source-compatible defaults.
- Persist each finalized batch in the same transaction as canonical data and checkpoint advancement.
- Drain after commit, acknowledge only after sink success, and reject PGlite.
- Cover historical delivery, realtime finalization, restart replay, failures, and all runtime ordering modes.

**Notes**

- No ClickHouse dependency or SQL enters core.
- Existing projects remain behaviorally and source compatible.
- This is the foundational review: it must land before any vendor package.

### DIFF
--- a/.changeset/silver-sinks.md
+++ b/.changeset/silver-sinks.md
@@ -1,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Added durable finalized analytics sink delivery.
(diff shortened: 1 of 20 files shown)


kyscott18:
We're working on a lower-level approach to fix the same problem your PR addresses.

Issue #2301, opened 2026-05-14
PostgreSQL drops all same-block events for factory-discovered contract addresses (regression in v0.16.0)


samclassix:
## Summary

When using a PostgreSQL database, **all events on a factory-discovered contract address that occur in the same block as the factory discovery event are silently dropped**. This affects both events that appear before *and* after the factory event in log order. The same configuration works correctly with PGlite (no `DATABASE_URL`).

This is a regression introduced in **v0.16.0** — it worked correctly on ≤ 0.15.x.

---

## Reproduction

**Setup:**
- Factory contract (e.g. `MintingHub`) emits a `PositionOpened` event to register a child address
- Child contract (e.g. `Position`) emits `OwnershipTransferred` and `MintingUpdate` in the **same transaction** as `PositionOpened`
- PostgreSQL backend (`DATABASE_URL` set)

**Example log order in the creation transaction:**
```
[1070]  Position    : OwnershipTransferred   ← before PositionOpened
[1072]  MintingHub  : PositionOpened          ← factory discovery
[1076]  Position    : MintingUpdate           ← after PositionOpened, same block
[1077]  Position    : OwnershipTransferred    ← after PositionOpened, same block
[1078]  Position    : MintingUpdate           ← after PositionOpened, same block
[1080]  Position    : OwnershipTransferred    ← after PositionOpened, same block
```

**Result with PostgreSQL:** all 5 child events (logs 1070, 1076, 1077, 1078, 1080) are never passed to their indexing functions. Tables remain empty.

**Result with PGlite:** all 5 events are processed correctly.

---

## Root cause (from reading the source)

In `sync-historical/index.js`, the `args.childAddresses` map used by `isAddressMatched` is built from the **database** before a sync batch runs:

```js
// runtime/filter.js
export const isAddressMatched = ({ address, blockNumber, childAddresses }) => {
  // childAddresses is pre-built from DB before this batch started
  childAddresses.get(toLowerCase(address)) <= blockNumber
  ...
}
```

Addresses discovered by a factory event within the **current batch** are inserted into the database during that batch (`insertChildAddresses`), but the `args.childAddresses` map is already frozen. Every event on the newly discovered address in that same batch fails the filter and is dropped.

**Why PGlite is unaffected:** `isolatedController.js` gates on `database.driver.dialect === "pglite"` (alongside `command === "dev"`) and takes a different code path that processes events sequentially, allowing same-block factory events to be handled correctly.

```js
// isolatedController.js
[...]

PR #2308, opened 2026-06-24
fix(sync-historical): fetch child events when child filter appears cached during same-batch factory creation

samclassix:
## Problem

When a factory creates a child contract **and** the child emits an event in the same `ethGetLogsBlockRange` batch window, the child's events are silently dropped on a cold re-index.

**Root cause:** When `syncBlockRangeData` processes batch `[A, B]`, it calls `syncAddressFactory` to discover new child addresses. If a child is created at block X within `[A, B]` and also emits events at block X, those events are never fetched — because the child's filter was not in `requiredIntervals` at the start of the batch (the child didn't exist yet), and the initial `eth_getLogs` did not include it. After the batch completes, `[A, B]` is marked as processed for the child filter, so the events from the creation block are permanently missed on any subsequent restart.

**Observed symptom:** A position contract created and denied in the same eth_getLogs batch shows `denied: false` after a full re-index — the `PositionDenied` event was never stored.

## Fix (3 parts)

1. **Pass event filters into `createHistoricalSync`** via a new optional `filters` parameter so the sync instance knows which child topics to look for.

2. **Supplementary `eth_getLogs` fetch in `syncBlockRangeData`:** after `syncAddressFactory` discovers newly created child addresses, immediately issue additional `eth_getLogs` calls for those addresses using their matching filter `topic0`. The existing dedupe step eliminates duplicates.

3. **Relax the `syncBlockData` processing guard** from `requiredIntervals.length > 0` to `requiredIntervals.length > 0 || logs.length > 0` so supplementary logs are not silently discarded when the child filter has no required intervals. Also adds a fallback log-matching path in `syncBlockData` that checks `args.filters` when `logFilters` is empty.

## Test

Added a regression test in `sync-historical/index.test.ts` that simulates the full bug scenario: empty `requiredIntervals` for both `syncBlockRangeData` and `syncBlockData` while the factory IS in `requiredFactoryIntervals`. The test verifies the child's Swap log is stored.

## Verification

End-to-end verified against a production Frankencoin ponder indexer: position `0xf35378277191c1f0d90869426f7177aa0393f77d` (created and denied in the same batch) now correctly returns `denied: true` after a full re-index with this patch applied -- and many more "dropped" events are found, comparing the patch against the production instance and on-chain data.

### DIFF
--- a/packages/core/src/runtime/historical.ts
+++ b/packages/core/src/runtime/historical.ts
@@ -1221,7 +1221,10 @@ export async function* getLocalSyncGenerator(params: {
     });
   }
 
-  const historicalSync = createHistoricalSync(params);
+  const historicalSync = createHistoricalSync({
+    ...params,
+    filters: params.eventCallbacks.map(({ filter }) => filter),
+  });
 
   const { callback: intervalCallback, generator: intervalGenerator } =
     createCallbackGenerator<{
--- a/packages/core/src/sync-historical/index.test.ts
+++ b/packages/core/src/sync-historical/index.test.ts
@@ -1142,3 +1142,86 @@ test("syncAddress() handles many addresses", async () => {
   expect(dbLogs).toHaveLength(1);
   expect(factories).toHaveLength(11);
 });
+
+test("sync() with log factory fetches child events even when child filter appears cached", async () => {
+  // Regression test for the "same-batch factory child address miss" bug.
+  //
+  // Scenario: a factory creates a child contract AND the child emits an event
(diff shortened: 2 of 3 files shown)


samclassix:
`0.16.6` — changelog notes factory fixes in 0.15.0 and 0.15.12 but none specifically address same-batch child events.

Issue awareness found here:
https://github.com/Frankencoin-ZCHF/ponder/blob/main/docs/ponder-factory-same-batch-miss.md

kyscott18:
See #2301

PR #2310, opened 2026-06-30
feat(core): pluggable historical sync provider

mo4islona:
## What this adds

A chain can back its historical backfill with a custom data source instead of JSON-RPC. You set one field in `ponder.config.ts`:

```ts
import { myProvider } from "my-sync-provider";

export default createConfig({
  chains: {
    mainnet: {
      id: 1,
      rpc: process.env.PONDER_RPC_URL_1,
      sync: { historical: myProvider({ /* ... */ }) },
    },
  },
  // ...
});
```

When `chain.sync.historical` is set, Ponder streams that chain's historical data from your provider. When it's unset, you get the built-in JSON-RPC sync, unchanged.

## Why

The historical backfill only runs over JSON-RPC. `createHistoricalSync` is wired straight into `runtime/historical.ts`, so there's no seam to swap it. For a large range that's the bottleneck: `eth_getLogs` splitting on size errors, one block at a time, rate limits. Bulk sources serve the same data faster — columnar exports, data lakes, archive dumps — but there's no supported way to feed one in.

A range query is the natural unit, so the seam is a range query. A provider implements two methods — `syncBlockRangeData` (fetch a range, write rows, return matched logs) and `syncBlockData` (finalize per-block data, return the tip block) — and writes `Sync*` rows to the `SyncStore`.

The runtime change is one line at each call site:

```ts
const createSync = params.chain.sync?.historical ?? createHistoricalSync;
const historicalSync = createSync(params);
```

The rest of the diff makes the contract reachable from outside core.

## The `ponder/sync` entrypoint

A provider lives in its own package and can't reach `@/...` internals. This PR adds a `ponder/sync` export with the contract a provider needs: `HistoricalSyncFactory`, the `Sync*` data types it produces, the `SyncStore` it writes to, the per-interval `Filter`/`Factory` inputs, and the filter-matching and block-interval helpers. So a provider needs no deep imports and doesn't re-declare core's shapes.

It's lower-level than the main `ponder` entrypoint, versioned with `ponder`, and may change in minor releases — pin accordingly. The doc comment on `sync.ts` notes this.

## Scope

- Default path (no `chain.sync`) is unchanged.
- `rpc` stays required. Realtime sync and indexing-function reads (`eth_call` and friends) still use it. This only swaps the historical backfill.
- `chain.sync.realtime` is reserved. The seam and `RealtimeSyncFactory` type exist and resolve the same way, but there's no realtime provider yet — unset uses the built-in RPC realtime sync.

## Tests

- `sync.test-d.ts` — type-level tests pinning the `ponder/sync` surface and the factory contract.
- The default path stays covered by the existing `sync-historical` / `sync-realtime` suites, since `chain.sync` defaults to the built-in factories.

Related discussion #2302

### DIFF
--- a/.changeset/pluggable-sync-providers.md
+++ b/.changeset/pluggable-sync-providers.md
@@ -1,0 +1,5 @@
+---
+"ponder": minor
+---
+
+Added `chain.sync` to plug in custom historical sync providers, and exposed the `ponder/sync` extension API. Setting `chain.sync.historical` streams a chain's historical backfill from a custom data source instead of JSON-RPC, without forking core. `ponder/sync` exports the full provider contract (`HistoricalSyncFactory`, the `Sync*` data types, `SyncStore`, filter/interval helpers) so a provider needs no deep imports. The default (no `chain.sync`) is unchanged; realtime and indexing-function reads still use `rpc`. `chain.sync.realtime` is reserved for a future realtime provider seam.
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -33,6 +33,10 @@
       "types": "./dist/types/index.d.ts",
       "import": "./dist/esm/index.js"
     },
+    "./sync": {
+      "types": "./dist/types/sync.d.ts",
+      "import": "./dist/esm/sync.js"
+    },
     "./virtual": {
       "types": "./src/types.d.ts"
     }
--- a/packages/core/src/_test/utils.ts
+++ b/packages/core/src/_test/utils.ts
@@ -668,6 +668,7 @@ export const getChain = (params?: {
     name: "mainnet",
     id: 1,
     rpc: `http://127.0.0.1:8545/${TEST_POOL_ID}`,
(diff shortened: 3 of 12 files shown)


mo4islona:
I’m more than happy to implement some kind of internal test suite for external sync providers, if maintainers approve the approach

kyscott18:
We're working on a lower-level approach to fix the same problem your PR addresses.

dokugo:
@kyscott18
Is there an ETA for this? This feature would be greatly appreciated. I noticed there is a fork with this already implemented (with participation from this PR's author): https://github.com/subsquid-labs/portal-ponder
But I'm wondering about the plans for historical sync providers in Ponder itself.

hskang9:
I want to know if there is an ETA for this too.

hskang9:
Is this applied now?

hskang9:
never mind, just use self-hosted erpc.cloud. this will solve everything.

PR #2315, opened 2026-07-06
Fix live query notify table indexes

kyscott18:
## Summary
- encode live query notifications as table indexes instead of table names
- decode indexes back to schema table names in the client
- preserve isolated ordering behavior by mapping partition names to base table names in SQL
- remove the schema-level 100 table limit
- add coverage for building schemas with more than 100 tables

## Notes
- This PR intentionally does not include notification batching, so it can be compared independently with the batching-only approach.

## Testing
- Not run: node/npm/pnpm/corepack are not available on PATH in this environment
- git diff --check


kyscott18:
Closing in favor of the batching approach in #2314.

PR #2334, opened 2026-07-15
chore: remove telemetry

kyscott18:
## Summary
- remove telemetry collection and lifecycle reporting from `ponder`
- remove the current telemetry docs page, endpoint, navigation, and rewrites
- retain telemetry pages for archived documentation versions
- remove the unused `@clickhouse/client` dependency

## Validation
- `pnpm --filter ponder typecheck`
- `pnpm --filter ponder-docs build`
- `pnpm exec biome check ...`


docs/pages/docs/0.10/advanced/telemetry.mdx
```diff
@@ -1,18 +0,0 @@
-# Telemetry [How Ponder collects and uses telemetry]
```
typedarray, line 1: I think we should leave these telemetry pages in the older versions of the docs.

typedarray (approved):

PR #2338, opened 2026-07-20
fix: simplify event query block bounds

### LINKED ISSUE #2208: Sync bottleneck by logsQuery
### Version

latest

### Current behavior

We experiencing serious performance issues with the following query:

```
const logsQuery = qb.raw
        .select({
          blockNumber: PONDER_SYNC.logs.blockNumber,
          logIndex: PONDER_SYNC.logs.logIndex,
          transactionIndex: PONDER_SYNC.logs.transactionIndex,
          address: PONDER_SYNC.logs.address,
          topic0: PONDER_SYNC.logs.topic0,
          topic1: PONDER_SYNC.logs.topic1,
          topic2: PONDER_SYNC.logs.topic2,
          topic3: PONDER_SYNC.logs.topic3,
          data: PONDER_SYNC.logs.data,
        })
        .from(PONDER_SYNC.logs)
        .where(
          and(
            eq(PONDER_SYNC.logs.chainId, BigInt(chainId)),
            gte(PONDER_SYNC.logs.blockNumber, BigInt(fromBlock)),
            lte(PONDER_SYNC.logs.blockNumber, BigInt(toBlock)),
            or(...logFilters.map((filter) => logFilter(filter))),
          ),
        )
        .orderBy(
          asc(PONDER_SYNC.logs.blockNumber),
          asc(PONDER_SYNC.logs.logIndex),
        )
        .limit(limit);
```

This ORM generate the following abomination even with only 2 ABI in an indexer:

```
select "block_number", "log_index", "transaction_index", "address", "topic0", "topic1", "topic2", "topic3", "data" from "ponder_sync"."logs" where ("ponder_sync"."logs"."chain_id" = $1 and "ponder_sync"."logs"."block_number" >= $2 and "ponder_sync"."logs"."block_number" <= $3 and (("ponder_sync"."logs"."topic0" = $4 and "ponder_sync"."logs"."address" in ($5) and "ponder_sync"."logs"."block_number" >= $6) or ("ponder_sync"."logs"."topic0" = $7 and "ponder_sync"."logs"."topic1" = $8 and true and "ponder_sync"."logs"."block_number" >= $9))) order by "ponder_sync"."logs"."block_number" asc, "ponder_sync"."logs"."log_index" asc limit $10
```

As time fly and the ponder_sync database grow, executing this queries became more and more expensive, at some point it they are became slower than the `statement_timeout` then indexing became unavailable or speed like 1000 blocks / hour or slower.

### Expected behavior

_No response_

### Steps to reproduce

_No response_

### Link to repository

_No response_

### Anything else?

_No response_

kyscott18:
## Summary
- merge filter block ranges with pagination so logs and traces queries contain exactly one lower and upper block bound
- skip filters that do not overlap the current page and preserve exact per-filter range checks in memory
- leave event-specific `OR` predicates in place; a follow-up can evaluate merging them with `IN` expressions or dropping very large predicate sets in favor of in-memory filtering

Closes #2208.

### DIFF
--- a/packages/core/src/sync-store/index.test.ts
+++ b/packages/core/src/sync-store/index.test.ts
@@ -1066,21 +1066,26 @@ test("getEventData() applies one block range to logs and traces queries", async 
     limit: 10,
   });
 
-  const logsQuery = querySpy.mock.calls.find(([query]) =>
-    query.includes('from "ponder_sync"."logs"'),
+  const queries = querySpy.mock.calls.map(([query, params]) =>
+    typeof query === "string"
+      ? { text: query, values: params }
+      : (query as { text: string; values: unknown[] }),
   );
-  const tracesQuery = querySpy.mock.calls.find(([query]) =>
-    query.includes('from "ponder_sync"."traces"'),
+  const logsQuery = queries.find(({ text }) =>
+    text.includes('from "ponder_sync"."logs"'),
+  );
+  const tracesQuery = queries.find(({ text }) =>
+    text.includes('from "ponder_sync"."traces"'),
   );
 
   expect(logsQuery).toBeDefined();
-  expect(logsQuery![0].match(/"block_number" >=/g)).toHaveLength(1);
-  expect(logsQuery![0].match(/"block_number" <=/g)).toHaveLength(1);
-  expect(logsQuery![1]).toEqual(expect.arrayContaining([10n, 90n]));
+  expect(logsQuery!.text.match(/"block_number" >=/g)).toHaveLength(1);
+  expect(logsQuery!.text.match(/"block_number" <=/g)).toHaveLength(1);
+  expect(logsQuery!.values).toEqual(expect.arrayContaining([10n, 90n]));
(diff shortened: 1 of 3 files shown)


packages/core/src/sync-store/index.test.ts
```diff
@@ -998,6 +1000,94 @@ test("getEventBlockData() returns events", async () => {
...
+  expect(
+    getFilterBlockRange({
+      filters: [
+        { fromBlock: 0, toBlock: 10 },
+        { fromBlock: 100, toBlock: 110 },
+      ],
+      fromBlock: 20,
+      toBlock: 90,
+    }),
+  ).toEqual({ fromBlock: 91, toBlock: 90 });
```
typedarray: Is this correct?

kyscott18: Nice catch, this codepath is unreachable in practice. I updated the tests accordingly and added a comment to document the implicit assumption.

typedarray (approved):

PR #2343, opened 2026-07-25
perf(graphql): batch many relation queries

hellowodl:
## Summary

- batch compatible GraphQL `many` relation resolutions in a request-scoped DataLoader
- use a partitioned window query to preserve per-parent ordering, limits, cursors, page info, and `totalCount` while bounding returned rows
- retain the existing resolver as a fallback for composite relations, cursor pagination, offsets, and null relation keys

## Benchmark

Deterministic in-memory PGlite data, one child per parent, `limit: 2`, with `totalCount`; wall time measures GraphQL execution only. Each point is one isolated execution after data setup (so timings are directional rather than a statistical suite).

| Parents / child rows | Before SQL | After SQL | Before wall | After wall | Heap delta before / after |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 10 | 21 | 2 | 10.43 ms | 3.22 ms | +6126 / +980 KiB |
| 50 | 101 | 2 | 28.89 ms | 2.56 ms | -3957 / +2701 KiB |
| 250 | 501 | 2 | 134.10 ms | 7.15 ms | +15835 / -23555 KiB |
| 1000 | 2001 | 2 | 577.50 ms | 25.29 ms | +23309 / +652 KiB |

Heap deltas are process-level snapshots and include V8 GC noise (negative values indicate a collection); the batched SQL result is bounded to `limit + 1` rows per parent. The root collection is one query; before this change each parent issues one items query plus one count query.

## Correctness

- added an exact result and SQL-count regression test covering empty relations, ordering, limit/page info, and `totalCount`
- existing GraphQL relation tests cover filters, custom ordering, relation names, camel-case columns, and fallback behavior
- focused GraphQL suite: 38/38 passed on PGlite
- production core build passed (`tsc` plus ESM/type output and path rewriting)
- patch changeset included

## DX

No schema or query changes are required. Existing GraphQL documents automatically use request-scoped batching when the relation has one key and requests the first page. Unsupported shapes transparently use the existing query path.

## Caveats

- batching is intentionally limited to single-column relations with no `before`/`after` cursor and zero offset
- composite keys, null parent keys, cursor pagination, and nonzero offsets fall back to the existing implementation
- SQL uses PostgreSQL window functions supported by both Postgres and PGlite; runtime correctness was exercised on PGlite in this environment

### DIFF
--- a/.changeset/tidy-graphql-many.md
+++ b/.changeset/tidy-graphql-many.md
@@ -1,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Improved GraphQL `many` relation performance by batching compatible resolutions within each request.
--- a/packages/core/src/graphql/index.test.ts
+++ b/packages/core/src/graphql/index.test.ts
@@ -21,7 +21,11 @@ import {
   primaryKey,
 } from "@/drizzle/onchain.js";
 import { EVENT_TYPES, encodeCheckpoint } from "@/utils/checkpoint.js";
-import { buildDataLoaderCache, buildGraphQLSchema } from "./index.js";
+import {
+  buildDataLoaderCache,
+  buildGraphQLSchema,
+  buildManyDataLoaderCache,
+} from "./index.js";
 
 beforeEach(setupCommon);
 beforeEach(setupIsolatedDatabase);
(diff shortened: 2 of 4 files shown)


hellowodl:
Opened against upstream by mistake; superseded by the fork-only review at https://github.com/hellowodl/ponder/pull/9.

PR #2344, opened 2026-07-25
perf: consolidate sync interval retrieval

hellowodl:
## Summary

- replace one `range_agg` subquery per sync fragment with a relational `(request_id, fragment_id)` request map joined once to `ponder_sync.intervals`
- map result rows by request id in linear time instead of rescanning all rows for every fragment
- retain parameter-limit batching and the existing fragment/factory output identity and ordering
- add a query-count/equivalence test and a patch changeset

## Benchmark

Postgres 17 (Docker, local), Node 26.5, two adjacent interval rows per requested fragment (`[0,10]` and overlapping `[5,15]`). SQL numbers are sums across all batches from `EXPLAIN (ANALYZE, FORMAT JSON, TIMING OFF)`. JS formatting uses equivalent returned multirange rows; memory is heap growth after forced GC. One measured run after setup (directional, not a statistically rigorous suite).

| Fragments | Statements old → new | SQL bytes old → new | Planning ms old → new | Execution ms old → new | JS format ms old → new | Heap MB old → new |
|---:|---:|---:|---:|---:|---:|---:|
| 100 | 1 → 1 | 15,359 → 2,439 | 1.323 → 0.456 | 2.283 → 0.380 | 0.205 → 0.059 | 0.094 → 0.046 |
| 1,000 | 5 → 1 | 156,615 → 25,839 | 12.939 → 0.543 | 69.842 → 1.787 | 6.175 → 0.196 | 0.800 → 0.414 |
| 5,000 | 25 → 1 | 796,395 → 145,839 | 65.023 → 2.715 | 346.702 → 25.054 | 132.742 → 0.895 | 3.984 → 2.016 |
| 20,000 | 100 → 3 | 3,225,570 → 636,397 | 239.424 → 10.632 | 2,355.402 → 59.218 | 2,158.177 → 2.734 | 15.776 → 7.963 |

The old/new JS outputs were byte-equivalent at every size. The new statement count at 20k reflects the existing 16k parameter limit (two parameters per request mapping).

## Compatibility & DX

- Adjacent fragment ids remain separate request-map rows grouped by the original request id, so overlapping and adjacent ranges still aggregate together.
- Factory fragment requests use the same map but remain separate from filter fragments.
- `LEFT JOIN` preserves fragments with no stored intervals as `[]`.
- Existing result assembly order and object-key identity are unchanged.
- The query was exercised directly on both Postgres 17 and PGlite, including an absent fragment and overlapping adjacent ids.
- Large configs now generate compact, conventional relational SQL instead of deeply repeated `UNION ALL` aggregation subqueries, which also makes traces/query logs easier to inspect.

## Validation

- `CI=1 pnpm --filter ponder test src/sync-store/index.test.ts` — 35 passed
- `pnpm --filter ponder typecheck` — passed
- `pnpm build` — passed (all publishable packages)
- `pnpm --filter ponder build` — passed after final source edits
- Biome check/write on changed TypeScript files — passed
- focused PGlite SQL compatibility smoke test — passed

### DIFF
--- a/.changeset/fast-interval-queries.md
+++ b/.changeset/fast-interval-queries.md
@@ -1,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Improved sync interval retrieval performance for configurations with many filter fragments.
--- a/packages/core/src/sync-store/index.test.ts
+++ b/packages/core/src/sync-store/index.test.ts
@@ -424,6 +424,40 @@ test("getIntervals() returns factory intervals when factory_log rows exist", asy
   expect(intervals.get(factory)![0]!.intervals).toEqual([[10, 20]]);
 });
 
+test("getIntervals() retrieves many fragments in one statement", async () => {
+  const { database, syncStore } = await setupDatabaseServices();
+  const filter = {
+    ...EMPTY_LOG_FILTER,
+    address: Array.from(
+      { length: 100 },
+      (_, index) => `0x${index.toString(16).padStart(40, "0")}` as const,
+    ),
+  } satisfies LogFilter;
+
+  await syncStore.insertIntervals({
+    intervals: [{ filter, interval: [0, 10] }],
+    factoryIntervals: [],
+    chainId: 1,
+  });
+
+  const wrap = vi.spyOn(database.syncQB, "wrap");
+  const intervals = await syncStore.getIntervals({ filters: [filter] });
+
+  expect(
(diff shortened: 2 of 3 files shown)


hellowodl:
Opened against upstream by mistake; superseded by the fork-only review at https://github.com/hellowodl/ponder/pull/10.

PR #2366, opened 2026-08-11
fix: generalize historical eth_getLogs request merging

kyscott18:
## Summary

Replaces the address-keyed `eth_getLogs` request merging algorithm in `syncBlockRangeData` with a general one, derived from the request-merging algorithm used by `mergeFiltersToQueryRequests` in `sync-historical/query.ts`.

Two log filters are now merged into a single `eth_getLogs` request when:
1. their required intervals overlap by at least 80% (of the smaller interval's length), and
2. at most one of `address`, `topic0`, `topic1`, `topic2`, or `topic3` differs between them (the differing dimension is unioned into an array).

This merges cases the old algorithm could not handle, for example:
- Filters with different addresses but identical topics.
- Filters with different indexed topic values (`topic1`/`topic2`/`topic3`).
- Filters where one has a factory-address and another doesn't (or where they reference different factories), once child addresses are resolved.

### DIFF
--- a/packages/core/src/sync-historical/index.ts
+++ b/packages/core/src/sync-historical/index.ts
@@ -4,7 +4,6 @@ import {
   type Hash,
   type Hex,
   hexToNumber,
-  type LogTopic,
   numberToHex,
   type RpcError,
   toHex,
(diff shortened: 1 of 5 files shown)


packages/core/src/sync-historical/index.ts
```diff
@@ -67,6 +70,164 @@ import { promiseAllSettledWithThrow } from "@/utils/promiseAllSettledWithThrow.j
...
+const logRequestKeys = [
+  "address",
+  "topic0",
+  "topic1",
+  "topic2",
+  "topic3",
+] as const;
+
+/** Returns `true` if `left` and `right` overlap by at least 80% of the smaller interval's length. */
+const hasSufficientLogIntervalOverlap = (
```
typedarray: Not a strong opinion, but seems like these helper / unit-testable functions should be a in different file.

typedarray (commented):
LGTM, one nit

PR #2367, opened 2026-08-14
Use seconds for unfinalized window

### LINKED ISSUE #2359: Support Robinhood Chain finality defaults and config overrides
### Version

Ponder 0.17.4.

### Problem

Robinhood Chain mainnet (4663) and testnet (46630) produce blocks at roughly 100 ms. `getFinalityBlockCount` currently falls through to the 30-block default intended for roughly 2-second chains, so the retained reorg window is only about 3 seconds instead of about 60 seconds.

The chain config builder also always replaces a caller-provided finality window with the utility result, and the public `ChainConfig` type does not expose `finalityBlockCount`.

### Proposed change

- Return 600 blocks for chain IDs 4663 and 46630 in the finality utility.
- Expose optional `finalityBlockCount` in `ChainConfig`.
- Prefer that explicit override in the config builder before the chain default.

Closed PR #2079 is prior art for the override portion.

### Local verification

An in-repo pnpm patch against 0.17.4 was exercised on a chain-46630 Anvil mirror: with head 709, Ponder stored finalized checkpoint 109 (exactly 600 blocks behind), and recovered an exact eight-event set after SIGKILL during ingestion.

If this direction is still welcome, I can turn the local patch into a focused PR against the supported release line.

kyscott18:
## Motivation

Ponder currently defines the recoverable reorg window as a chain-specific number of blocks. That assumption breaks down on fast chains: Robinhood Chain produces blocks roughly every 100 ms, so the 30-block fallback retains only about 3 seconds of rollback data. Adding more per-chain block counts would continue coupling reorg safety to incomplete and potentially stale block-time assumptions.

This PR defines the window in elapsed chain time instead. A duration has consistent semantics across chains and gives apps an explicit resource-versus-safety control.

## User-facing behavior

- Adds `reorgWindow` to each chain in `ponder.config.ts`. It is an integer number of seconds from `0` through `600` and defaults to `180`.

Closes #2359.

### DIFF
--- a/.changeset/poor-friends-rest.md
+++ b/.changeset/poor-friends-rest.md
@@ -1,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Added `reorgWindow` to chain configuration in `ponder.config.ts` to control how long Ponder retains the data required to recover from reorgs. The window is measured in seconds and defaults to `180`.
--- a/.changeset/poor-friends-rest.md
+++ b/.changeset/poor-friends-rest.md
@@ -1,5 +1,5 @@
 ---
-"ponder": patch
+"ponder": minor
 ---
 
 Added `reorgWindow` to chain configuration in `ponder.config.ts` to control how long Ponder retains the data required to recover from reorgs. The window is measured in seconds and defaults to `180`.
--- a/docs/pages/docs/config/chains.mdx
+++ b/docs/pages/docs/config/chains.mdx
@@ -177,6 +177,28 @@ export default createConfig({
 });
 ```
 
+## Reorg window
+
+The `reorgWindow` option controls how long Ponder retains the data required to automatically recover from a chain reorganization. The value is measured in seconds and defaults to `180` (3 minutes).
+
+Increasing `reorgWindow` lets Ponder recover from older reorgs, but increases memory and database usage. Decreasing it reduces resource usage, but increases the chance that Ponder encounters an unrecoverable reorg and errors.
+
+The value must be an integer between `0` and `600` (10 minutes). Setting `reorgWindow` to `0` disables automatic reorg recovery by treating each block as finalized immediately. Ponder will still detect reorgs and stop rather than continue with inconsistent data.
+
+```ts [ponder.config.ts]
+import { createConfig } from "ponder";
+
+export default createConfig({
+  chains: {
+    mainnet: {
+      id: 1,
+      rpc: process.env.PONDER_RPC_URL_1,
+      reorgWindow: 300, // 5 minutes [!code focus]
+    },
+  },
+});
+```
+
 ## Disable caching
 
 Use the `disableCache` option to disable caching for RPC responses. The default is `false`.
--- a/docs/pages/docs/config/chains.mdx
+++ b/docs/pages/docs/config/chains.mdx
@@ -181,9 +181,9 @@ export default createConfig({
 
 The `reorgWindow` option controls how long Ponder retains the data required to automatically recover from a chain reorganization. The value is measured in seconds and defaults to `180` (3 minutes).
 
-Increasing `reorgWindow` lets Ponder recover from older reorgs, but increases memory and database usage. Decreasing it reduces resource usage, but increases the chance that Ponder encounters an unrecoverable reorg and errors.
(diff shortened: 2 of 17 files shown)


docs/pages/docs/config/chains.mdx
```diff
@@ -177,6 +177,28 @@ export default createConfig({
 });
 ```
 
+## Reorg window
+
+The `reorgWindow` option controls how long Ponder retains the data required to automatically recover from a chain reorganization. The value is measured in seconds and defaults to `180` (3 minutes).
+
+Increasing `reorgWindow` lets Ponder recover from older reorgs, but increases memory and database usage. Decreasing it reduces resource usage, but increases the chance that Ponder encounters an unrecoverable reorg and errors.
```
typedarray: "and crashes with a fatal error"

docs/pages/docs/config/chains.mdx
```diff
@@ -177,6 +177,28 @@ export default createConfig({
...
+
+The value must be an integer between `0` and `600` (10 minutes). Setting `reorgWindow` to `0` disables automatic reorg recovery by treating each block as finalized immediately. Ponder will still detect reorgs and stop rather than continue with inconsistent data.
```
typedarray: What's the valid / reasonable use case for providing "0"? Like as a user, when would I want to use that option?

.changeset/poor-friends-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+"ponder": patch
```
typedarray: This arguably warrants a minor. The new default is dramatically different from the old default for some chains.

packages/core/src/build/config.ts
```diff
@@ -188,30 +187,26 @@ export async function buildIndexingFunctions({
...
-          throw new Error(
-            `Unable to fetch "latest" block for chain '${chain.name}':\n${e.message}`,
-          );
-        });
+      }).catch((e) => {
+        throw new Error(
+          `Unable to fetch "latest" block for chain '${chain.name}':\n${e.message}`,
+        );
+      });
+      const blockPromise = latestBlockPromise.then((block) =>
```
typedarray: blockNumberPromise ?

packages/core/src/runtime/historical.ts
```diff
@@ -353,29 +360,20 @@ export async function* getHistoricalEventsOmnichain(params: {
       }),
     );
 
-    let shouldCatchup = false;
```
typedarray, line 356: Is it possible to test/verify these changes?

packages/core/src/sync-historical/index.ts
```diff
@@ -996,11 +996,7 @@ export const createHistoricalSync = (
       let receiptCount = 0;
       let traceCount = 0;
 
-      // Same memory usage as `sync-realtime`.
-      const MAX_BLOCKS_IN_MEM = Math.max(
-        args.chain.finalityBlockCount * 2,
-        100,
-      );
+      const MAX_BLOCKS_IN_MEM = 100;
```
typedarray, line 999: Just noting that it seems like this value could previously be much larger

packages/core/src/utils/finality.ts
```diff
@@ -1,43 +1,76 @@
...
- * Returns the number of blocks that must pass before a block is considered final.
- * Note that a value of `0` indicates that blocks are considered final immediately.
- *
- * @param chain The chain to get the finality block count for.
- * @returns The finality block count.
+ * Finds a block whose timestamp is outside the reorg window.
+ * Block timestamps are monotonic, so a binary search avoids retaining or
+ * fetching one block per second on fast chains.
  */
-export function getFinalityBlockCount({ chain }: { chain: Chain | undefined }) {
```
typedarray, line 13: Finally 🙏

docs/pages/docs/config/chains.mdx
kyscott18: I'm planning on a separate PR for an optimized `reorgWindow: 0` mode. It would have meaningful performance advantages over a non-zero value, because data structures for rollback support (like `reorg_` tables, database triggers) can be categorically skipped.

typedarray: Got it, new writing makes sense.

typedarray (approved):

Issue #2181, opened 2025-11-14
RpcProviderError Invalid RPC response


qiuchenghui:
### Version

0.15.2

### Current behavior

Hi team,

I’m encountering the following RPC error when using Alchemy’s RPC service on BNB Chain, and would like your help to understand and resolve it:

ponder startBlock: 46770851

```
03:36:39.731 WARN  Failed to fetch backfill JSON-RPC data chain=bnb block_range=[46785075,46785100] (1m 05s)
RpcProviderError: Invalid RPC response: 'log.logIndex' (4294967283) is larger than the maximum allowed value (2147483647).
Please report this error to the RPC operator.
Logs request: {
  "method": "eth_getLogs",
  "params": [
    {
      "fromBlock": "0x0",
      "toBlock": "0x0"
    }
  ]
}
03:36:39.732 ERROR unhandledRejection
RpcProviderError: Invalid RPC response: 'log.logIndex' (4294967283) is larger than the maximum allowed value (2147483647).
Please report this error to the RPC operator.
Logs request: {
  "method": "eth_getLogs",
  "params": [
    {
      "fromBlock": "0x0",
      "toBlock": "0x0"
    }
  ]
}
03:36:39.732 WARN  Started shutdown sequence
error Command failed with exit code 75
```

Could you please help check what might be causing this issue on Alchemy’s BNB Chain RPC, or advise how we should handle it from our side?

Thanks in advance!

kyscott18:
We've updated the error messages in `0.15.3`. It should give you a more actionable message that you can reproduce the issue with.

qiuchenghui:
> We've updated the error messages in `0.15.3`. It should give you a more actionable message that you can reproduce the issue with.

Hi, thanks for the update!
I’ve already upgraded to version `0.15.4`, but I’m still seeing exactly the same error as before.
Please let me know if there’s any additional information or logs I can provide to help reproduce the issue.
Thanks!

```
04:07:04.518 WARN  Failed database query query_id=6fbb0f4b retry_count=8 retry_delay=32000 (264ms)
RpcProviderError: Invalid RPC response: 'log.logIndex' (4294967277) is larger than the maximum allowed value (2147483647).
Please report this error to the RPC operator.
Logs request: {
  "method": "eth_getLogs",
  "params": [
    {
      "fromBlock": "0xc",
      "toBlock": "0xc"
    }
  ]
}
```

zillarelli:
[...]

Issue #2103, opened 2025-10-08
out of range for type integer for custom event example


mai1015:
### Version

0.13.13

### Current behavior

5:31:26 PM WARN  database   Failed 'insert_logs' database query after '10' attempts (id=813245e7)
error: value "4294967295" is out of range for type integer
    at he.Fe (file:///Volumes/Data/codes/obscure/leagues-eth-indexer/node_modules/.pnpm/@electric-sql+pglite@0.2.13/node_modules/@electric-sql/pg-protocol/src/parser.ts:432:11)
    at he.rt (file:///Volumes/Data/codes/obscure/leagues-eth-indexer/node_modules/.pnpm/@electric-sql+pglite@0.2.13/node_modules/@electric-sql/pg-protocol/src/parser.ts:210:16)
    at he.__privateGet [as parse] (file:///Volumes/Data/codes/obscure/leagues-eth-indexer/node_modules/.pnpm/@electric-sql+pglite@0.2.13/node_modules/@electric-sql/pg-protocol/src/parser.ts:102:25)
    at te.execProtocol (file:///Volumes/Data/codes/obscure/leagues-eth-indexer/node_modules/.pnpm/@electric-sql+pglite@0.2.13/node_modules/@electric-sql/pglite/src/pglite.ts:609:5)
    at te.o (file:///Volumes/Data/codes/obscure/leagues-eth-indexer/node_modules/.pnpm/@electric-sql+pglite@0.2.13/node_modules/@electric-sql/pglite/src/base.ts:130:12)
    at file:///Volumes/Data/codes/obscure/leagues-eth-indexer/node_modules/.pnpm/@electric-sql+pglite@0.2.13/node_modules/@electric-sql/pglite/src/base.ts:245:13
where: unnamed portal parameter $27 = '...'
5:31:26 PM ERROR sync       Fatal error: Unable to sync 'mainnet' from 14517920 to 14597008.
error: value "4294967295" is out of range for type integer
    at he.Fe (file:///Volumes/Data/codes/obscure/leagues-eth-indexer/node_modules/.pnpm/@electric-sql+pglite@0.2.13/node_modules/@electric-sql/pg-protocol/src/parser.ts:432:11)
    at he.rt (file:///Volumes/Data/codes/obscure/leagues-eth-indexer/node_modules/.pnpm/@electric-sql+pglite@0.2.13/node_modules/@electric-sql/pg-protocol/src/parser.ts:210:16)
[...]

PR #2156, opened 2025-10-29
validate JSON-RPC response INTEGER and BIGINT size

kyscott18:
Closes #2103

PR #2373, opened 2026-08-18
fix(core): support large log indices

kernelwhisperer:
## Summary

- Store `ponder_sync.logs.log_index` as a Postgres `bigint`.
- Accept `logIndex` values through `Number.MAX_SAFE_INTEGER`.
- Keep the validation for values that JavaScript cannot represent safely.

## Why

Some RPC providers return synthetic logs with `logIndex` values near the unsigned 32-bit limit. Examples include `0xffffffe2` in #2103 and similar values in #2181.

The Ethereum execution API defines `logIndex` as a hex-encoded unsigned integer. It does not define a signed 32-bit limit. The current limit stops the backfill when these logs occur.

We also received `0xfffffffc` from a Gnosis RPC. This change preserves that log instead of dropping it or stopping the indexer.

## Migration

The sync migration changes the existing `logs.log_index` column from `integer` to `bigint`. Existing values do not change.

## Tests

- Added a regression test for `0xfffffffc`.
- Added a test for values above `Number.MAX_SAFE_INTEGER`.
- Ran the Ponder build successfully.

### DIFF
--- a/.changeset/tidy-cows-index.md
+++ b/.changeset/tidy-cows-index.md
@@ -1,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Fixed a crash caused by RPC logs with a `logIndex` larger than a Postgres integer.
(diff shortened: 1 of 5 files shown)


kyscott18:
Ponder considers `logIndex` values larger than the Postgres integer maximum represent a bug in the RPC/chain implementation, not a case that should be accommodated. We’ve handled similar reports before (#2103 , #2181) and landed validation in #2156 to reject those responses rather than widen the schema.

If you believe this behavior is valid for a specific chain, please re-open with a link to official documentation from the implementing chain that specifies logIndex can exceed 2147483647, and we’ll revisit.

Also, you can fork Ponder and maintain this feature there.

PR #2374, opened 2026-08-19
feat: opt-in HyperSync fast path for historical sync

moose-code:
## What

An opt-in fast path for historical sync backed by [Envio HyperSync](https://docs.envio.dev/docs/HyperSync/overview). With `PONDER_HYPERSYNC=true` and an API token in `ENVIO_API_TOKEN`, bounded-range `eth_getLogs` requests are served by a single HyperSync query whose join also returns the blocks and transactions the matched logs touch. The reconstructed blocks are cached, and the per-eventful-block `eth_getBlockByNumber` requests that follow — the dominant cost of a Ponder backfill — are answered from memory instead of the network. Requests near the chain tip, cache misses, and all of realtime sync use the RPC exactly as before.

Deliberately small and inert by default:

- **One new file** (`rpc/hypersync.ts`, ~340 lines) plus two short guards in `rpc/actions.ts`. No changes to the sync engine, sync-store, config schema, or types.
- **The client is an optional peer dependency, imported lazily.** A Ponder install that never enables the path carries no native module; if the module is absent, every request permanently falls back to the RPC.
- **Reconstruction is held to Ponder's own bar**: the blocks pass `validateLogsAndBlock` (faithful `logsBloom`, log↔block hash and number equality, transaction-index lookups), and results flow through `standardizeLogs`/`standardizeBlock` like any RPC response.

## Motivation — benchmarks

Measured with the [open-indexer-benchmark](https://github.com/enviodev/open-indexer-benchmark) harness (its methodology: a verified range checked row-by-row against ground truth, then best-of-two timed throughput windows). Same machine, same harness, same Ponder build — only the env var differs. Baseline runs reproduced the benchmark's published Ponder numbers within ±25% on all four cases.

| case | Ponder (RPC) | Ponder + HyperSync | data verification |
| --- | --- | --- | --- |
| Decoded Event Stream (dense USDC transfers) | 182 events/s | **4,399 events/s (24×)** | ✅ both |
| State Aggregation (rETH balances) | 66 events/s | **1,008 events/s (15×)** | ✅ both |
| External Contract Calls | DNF at 300s (57–61% missing) | **completes, verified** | ❓ → ✅ |
| Factory Registration (82k Safe proxies) | 309 events/s | **2,234 events/s (7×)** | ✅ both |

The mechanism behind the numbers: a backfill's cost is dominated by one full-transaction block fetch per eventful block. On the dense case that's ~1,000 block requests per 1,000 blocks; through this path the same range produced **zero `eth_getLogs` and five `eth_getBlockByNumber`** over RPC (measured through a counting proxy), with everything else served by the joined HyperSync responses. Factory child discovery goes through the same `eth_getLogs` seam, so it accelerates too — with no address-list chunking, since HyperSync takes arbitrary address sets.

## Scope notes / open questions

- Env-var activation keeps the change minimal for discussion. If there's appetite, the natural evolution is a per-chain config field alongside `rpc` — happy to rework in that direction.
- The fast path only ever *adds* a source for data Ponder already fetches; correctness relies on Ponder's existing validation plus the benchmark's row-level ground-truth checks above. Pointers to any additional invariants you'd want covered by tests are welcome.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/hypersync-historical-source.md
+++ b/.changeset/hypersync-historical-source.md
@@ -1,0 +1,5 @@
+---
+"ponder": patch
+---
+
+Added an opt-in HyperSync fast path for historical sync. With `PONDER_HYPERSYNC=true` and an API token in `ENVIO_API_TOKEN`, bounded-range `eth_getLogs` requests are served by a single HyperSync query whose join also returns the blocks and transactions those logs touch, and the per-eventful-block `eth_getBlockByNumber` requests are answered from that data. Requests near the chain tip, cache misses, and realtime sync use the RPC unchanged. The client library is an optional peer dependency, imported lazily — installs that never enable the path carry no native module, and if it is absent every request falls back to the RPC.
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -47,10 +47,14 @@
   "peerDependencies": {
     "hono": ">=4.5",
     "typescript": ">=5.4.0",
-    "viem": ">=2.35"
+    "viem": ">=2.35",
+    "@envio-dev/hypersync-client": ">=0.6.5"
   },
   "peerDependenciesMeta": {
     "typescript": {
+      "optional": true
+    },
+    "@envio-dev/hypersync-client": {
       "optional": true
     }
   },
@@ -109,7 +113,8 @@
     "mitata": "^1.0.34",
     "rimraf": "^5.0.5",
     "tsx": "^4.19.2",
-    "vitest": "1.6.1"
+    "vitest": "1.6.1",
+    "@envio-dev/hypersync-client": "^0.6.5"
   },
   "engines": {
     "node": ">=22"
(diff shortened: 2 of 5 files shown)


dokugo:
There was a similar attempt by the SQD.dev (Portal) team: https://github.com/ponder-sh/ponder/pull/2310
Relevant discussion: https://github.com/ponder-sh/ponder/discussions/2302
They ultimately forked Ponder. I'm currently using their fork, and it's highly effective. It would be great to see a plugin or adapter system built into Ponder so we can easily connect to various historical sync sources, removing the need for forks altogether.

kyscott18:
Thanks for the proposal and benchmarks. We're working on a lower-level approach to address the same historical-sync problem, rather than adding a HyperSync-specific integration here. Feel free to add this feature in a fork.

PR #2382, opened 2026-09-08
Decouple the sync engine from the database layer, and expose it as `ponder/sync`

moodysalem:
Opened as a draft — this touches layering across `packages/core`, and the second commit adds public API, so it's a proposal rather than a finished call. Happy to cut it down, split it, or drop the second commit entirely.

## The problem

`sync-historical` and `sync-realtime` are already written against the `SyncStore` type and never touch Drizzle themselves. But following imports from those two modules and `rpc/index.ts` reaches **114 files and 41,975 lines** — the CLI, the server, GraphQL, the indexing store, the Drizzle schema builder. The engine is decoupled in spirit and coupled in the import graph.

Four edges caused all of it, and each pointed the wrong way:

| edge | why it matters |
|---|---|
| `internal/options.ts` → `bin/ponder.ts` for `type CliOptions` | every holder of an `Options` reached the CLI — and `bin/ponder.ts` runs a command as a side effect of being imported |
| `sync-historical` / `runtime` → `sync-store/index.ts` for `type SyncStore` | depending on the contract meant depending on its Drizzle implementation |
| `internal/types.ts` → `drizzle/kit` for `SqlStatements` | the blockchain data model and the build artefacts shared a module, so the model named `SqlStatements`, `PGlite`, `PoolConfig` and `Hono` |
| `internal/metrics.ts` → `drizzle-orm` for `isTable` | filtering a user schema to label a metric put Drizzle in the graph of everything holding a `Common` |

## Commit 1 — stop the sync engine importing the database, server, and CLI layers

Each edge is turned around rather than removed:

- `CliOptions` moves to `internal/options.ts`, beside the `buildOptions` that reads it. `bin/ponder.ts` keeps its commander definitions; each `.action` handler already passes its parsed options into a command declaring `CliOptions`, so a flag that stops matching is still a compile error, now at the call site.
- The `SyncStore` type moves to `sync-store/store.ts`. `sync-store/index.ts` re-exports it, so nothing else changes.
- `DatabaseConfig`, `PreBuild`, `SchemaBuild` and `ApiBuild` move to `internal/build.ts`.
- `initializeIndexingMetrics` takes table names instead of a schema; resolving a Drizzle table to its name becomes `getSchemaTableNames` in the schema layer. `rpc/actions.ts` reached the `pg` driver for two integer constants, which move to `utils/pgLimits.ts`.

The closure is now **36 files and 10,707 lines**, with no file from those layers and no import of `drizzle-orm`, `pg`, `@electric-sql/pglite` or `hono`.

`internal/layering.test.ts` walks the import graph and fails if that changes. Reverting any one of the four edges fails it, naming the offending file and the package it drags in — I checked each.

This commit stands on its own. Even setting packaging aside, a leaf config module importing a script that calls `process.argv` parsing at import time is worth turning around.

## Commit 2 — a `ponder/sync` entrypoint

Being decoupled is not the same as being reachable: `ponder` exports only `.` and `./virtual`. `src/sync.ts` exports the engine factories, `createRpc`, the `SyncStore` type, the interval/progress helpers, the filter and fragment predicates, the blockchain types, and the four services a `Common` is made of.

`createSyncStore` is deliberately **not** exported — that would put Drizzle back in the entrypoint's graph, and the point of the contract is that an application brings its own.

The parameters turn out to be constructible from outside, which is what makes this worth doing at all: `createHistoricalSync` takes `{ common, chain, rpc, childAddresses }`, `createRealtimeSync` adds `eventCallbacks` and a finalized block, and neither needs a build artefact. `syncStore` is an argument to `syncBlockRangeData` rather than a constructor dependency. Realtime reports `block` / `finalize` / `reorg` for the caller to handle.

So an application with its own database can implement `SyncStore` over its own tables and get the parts that are tedious to get right — `eth_getLogs` range estimation and error recovery, interval bookkeeping so a restart doesn't refetch, factory child-address discovery, block-hash reorg detection with finality tracking, and the rpc cache — without adopting the indexing model or the schema.

That is our interest here, for the record: we maintain a multi-chain indexer with its own Postgres schema and would rather use this engine than keep our own.

## Verification

- `pnpm --filter ponder build`, `pnpm --filter ponder typecheck`, `pnpm lint` — all clean.
- **The built output, not just the type graph**: walking `dist/esm/sync.js` reaches 31 emitted modules and, outside Node builtins, only `@ponder/utils`, `picocolors`, `pino`, `prom-client`, `semver`, `viem`, `ws`.
- `internal`, `runtime`, `utils`, `rpc` — 271 tests pass.
- `sync-historical`, `sync-realtime`, `sync-store` — 76 pass, 1 fails: `handleReorg() throws error for deep reorg` times out at 15s. **It fails identically on `4a2c771` without these changes**, which I verified by checking out the upstream sources in the same tree; I have not looked into it further.

No behaviour changes are intended anywhere. The only runtime change is `initializeIndexingMetrics`, which is internal and whose three callers are updated.

## What this is not

It is not the separate package that would let the engine be installed without the framework. I started there and backed off: the closure contains `internal/common|logger|metrics|options|errors` and 17 `utils/` files, which are framework-wide — `internal/types.ts` alone has 53 importers outside the closure. A `@ponder/sync` owning the logger and metrics for the whole framework would be the wrong package. Doing it properly is a three-way split (`@ponder/common` → `@ponder/sync` → `ponder`), moving 36 files and repointing 63 source and 39 test files, and the naming, versioning and release surface are yours to decide, not mine to presume. This PR is the prerequisite either way, and I'd be glad to do that follow-up if you want it.

Related: #4, from 2022, proposed separating packages so they could be plugged into the core.

The changeset is a minor. Per `AGENTS.md` this is the pre-1.0 case where changesets produces `1.0.0` by mistake — it should be `0.18.0` in the release PR.

https://claude.ai/code/session_018fCeEYi1BdAhcaKZmDwAKb

### DIFF
--- a/.changeset/tidy-sync-entrypoint.md
+++ b/.changeset/tidy-sync-entrypoint.md
@@ -1,0 +1,5 @@
+---
+"ponder": minor
+---
+
+Added a `ponder/sync` entrypoint that exposes the sync engine on its own. It exports `createHistoricalSync`, `createRealtimeSync`, `createRpc`, and the `SyncStore` type the engine reads and writes through, so an application can run Ponder's backfill, reorg handling, factory child-address discovery, and rpc cache against its own tables by implementing `SyncStore` and handling `RealtimeSyncEvent` directly. Importing `ponder/sync` does not load Drizzle, `pg`, PGlite, Hono, or the CLI.
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -33,6 +33,10 @@
       "types": "./dist/types/index.d.ts",
       "import": "./dist/esm/index.js"
     },
+    "./sync": {
+      "types": "./dist/types/sync.d.ts",
+      "import": "./dist/esm/sync.js"
+    },
     "./virtual": {
       "types": "./src/types.d.ts"
     }
--- a/packages/core/src/_test/setup.ts
+++ b/packages/core/src/_test/setup.ts
@@ -3,17 +3,16 @@ import pg from "pg";
 import { afterAll } from "vitest";
 import { buildSchema } from "@/build/schema.js";
 import { createDatabase, type Database } from "@/database/index.js";
+import type { DatabaseConfig, SchemaBuild } from "@/internal/build.js";
 import type { Common } from "@/internal/common.js";
 import { createLogger } from "@/internal/logger.js";
 import { MetricsService } from "@/internal/metrics.js";
 import { buildOptions } from "@/internal/options.js";
 import { createShutdown } from "@/internal/shutdown.js";
 import type {
-  DatabaseConfig,
   EventCallback,
   IndexingBuild,
   NamespaceBuild,
-  SchemaBuild,
 } from "@/internal/types.js";
 import { getFilterFactories, isAddressFactory } from "@/runtime/filter.js";
 import { getFactoryFragments, getFragments } from "@/runtime/fragments.js";
--- a/packages/core/src/bin/commands/codegen.ts
+++ b/packages/core/src/bin/commands/codegen.ts
@@ -1,9 +1,9 @@
 import { runCodegen } from "@/bin/utils/codegen.js";
 import { createLogger } from "@/internal/logger.js";
 import { MetricsService } from "@/internal/metrics.js";
+import type { CliOptions } from "@/internal/options.js";
 import { buildOptions } from "@/internal/options.js";
 import { createShutdown } from "@/internal/shutdown.js";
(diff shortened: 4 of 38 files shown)


kyscott18:
Thanks for opening, but we're not interested in maintaining this feature long term. Feel free to fork Ponder. We've found package manager-level patches to be useful for running unreleased forks or branches.

- https://pnpm.io/cli/patch
- https://bun.com/docs/pm/cli/patch

PR #2390, opened 2026-09-16
Fix Avalanche realtime log syncing after Helicon

JamieLivi:
## Summary

- Realtime sync now always requests logs for blocks with a settlement-scoped `logsBloom` ([Avalanche ACP-194](https://build.avax.network/docs/acps/194-continuous-execution)), instead of skipping them on a bloom miss
- `validateLogsAndBlock` allows a non-empty settlement-scoped bloom w/ no logs, and the per-log "Log not found in block.logsBloom" warning is skipped for those blocks
- Detection is a non-null `settledHeight` on the header (as [here](https://github.com/ava-labs/avalanchego/blob/v1.15.0/vms/saevm/cchain/README.md#block-header-changes)), not a chain ID, so pre-Helicon blocks and every other chain keep bloom-based skipping

## Why

[Avalanche's Helicon upgrade](https://build.avax.network/blog/helicon-upgrade) went live on Fuji on 28 Jul and hits C-Chain mainnet on 22 Sep. It turns on ACP-194 (Continuous Execution), and with that a block's `logsBloom` covers the receipts of the blocks it *settles*, i.e. heights `(parent.settledHeight, block.settledHeight]`, while `eth_getLogs` still only returns the block's own logs. Ava Labs acknowledge the mismatch in [avalanchego#5761](https://github.com/ava-labs/avalanchego/issues/5761).

That breaks realtime sync:

1. A block whose own logs match a filter but whose settled range doesn't gets skipped, so events are dropped silently
2. A block w/ no logs of its own whose settled range *does* match gets fetched, then fails `validateLogsAndBlock` with `Inconsistent RPC response data. The logs array has length 0, but the associated block has a non-empty 'block.logsBloom'.` Realtime retries it forever and `/ready` keeps returning 200

Adding 43113/43114 to `isAsyncExecutionChain` like the Monad fix (#2317) would sort out 1 but make 2 happen on almost every block, and would kick in on mainnet before activation too. Post-Helicon headers have a `settledHeight` field, which is the signal Ava Labs suggest in [avalanchego#5761](https://github.com/ava-labs/avalanchego/issues/5761#issuecomment-5412119443) ("`nil` for pre-SAE blocks"). Historical sync isnt affected

## Testing on live Fuji

- Bloom semantics: for 120/120 consecutive blocks, `logsBloom` matched the bloom of all logs in `(parent.settledHeight, settledHeight]`, and blocks that settle nothing had a zero bloom
- A/B: same app (one contract that emits in ~50% of blocks), same 75-block window, checked against `eth_getLogs`

| | events indexed | what happened |
|---|---|---|
| `ponder@0.17.10` | 15 / 22 | stalled on the next block w/ the error above (84 retries) |
| this branch | 22 / 22 | kept up with head, no warnings |

There's a bloom-preserving option too (hold each block back by the settlement lag and use the settling block's bloom), but it changes realtime latency so have left it out

## Compatibility

If `settledHeight` is missing or null, every changed expression is the same as before, so nothing changes

- **Avalanche before activation:** coreth rejects any pre-Helicon block that carries the field ([`VerifySettled`](https://github.com/ava-labs/avalanchego/blob/70bd6d063b7343fd2cd8217200aaf77b57f19f68/graft/coreth/plugin/evm/customheader/settled.go#L15-L32)) and `eth_getBlockBy*` leaves it out when it's unset. `newHeads` on a v1.15.0 node can send `"settledHeight": null` though (no `omitempty` in [`gen_header_serializable_json.go`](https://github.com/ava-labs/avalanchego/blob/70bd6d063b7343fd2cd8217200aaf77b57f19f68/graft/coreth/plugin/evm/customtypes/gen_header_serializable_json.go#L46)), hence the non-null check. Mainnet blocks and `newHeads` don't have the field today. On Fuji, blocks before 28 Jul don't have it and every block since does
- **Other chains:** no `settledHeight` in `eth_getBlockByNumber("latest")` on 43 other networks, incl. Ethereum, Sepolia, Base, Arbitrum, Optimism, Polygon, BSC, zkSync Era, Linea, Scroll, Monad, Sei, HyperEVM, Berachain, Sonic, Celo, Mantle, Flare, Songbird and a few Avalanche L1s. GitHub code search didn't turn up any other client emitting it in block JSON. Worst case on a false positive is an extra `eth_getLogs` per block and the two bloom checks being skipped, logs never get skipped
- The extra field isn't stored or exposed anywhere, since storage and event construction copy block fields by name

## Tests

- `CI=1 pnpm --filter ponder test src/rpc/actions.test.ts src/sync-realtime/bloom.test.ts src/sync-realtime/index.test.ts`
- `pnpm --filter ponder typecheck`
- `pnpm lint`

New coverage:

- settlement-scoped blocks request logs, for both full blocks and header-only (`newHeads`) blocks
- an empty settlement-scoped block doesn't fail validation
- the bloom warning is skipped for settlement-scoped blocks but still fires on standard ones
- a null `settledHeight` keeps bloom-based skipping
- `validateLogsAndBlock` unit tests for zero, standard and settlement-scoped blooms

Each one fails if its bit of the fix is reverted

## References

- ACP-194 Continuous Execution: https://build.avax.network/docs/acps/194-continuous-execution
- Helicon upgrade announcement: https://build.avax.network/blog/helicon-upgrade
- avalanchego v1.15.0 release (schedules mainnet activation): https://github.com/ava-labs/avalanchego/releases/tag/v1.15.0
- C-Chain SAE README (`settledHeight` = height of the block this header settles): https://github.com/ava-labs/avalanchego/blob/v1.15.0/vms/saevm/cchain/README.md
- avalanchego#5573, adds the settled block marker to the header: https://github.com/ava-labs/avalanchego/pull/5573


kyscott18:
I simplified by 1) fixing the `logsBloom` validations for asynchronous execution chains and 2) considering Avalanche and Avalanche testnet an async execution chain.

packages/core/src/rpc/actions.ts
```diff
@@ -378,8 +379,13 @@ export const validateLogsAndBlock = (
     RequestParameters,
     { method: "eth_getBlockByNumber" | "eth_getBlockByHash" }
   >,
+  isisAsyncExecutionChain: boolean,
```
typedarray: Typo

JamieLivi:
Cool cheers @kyscott18 yeah thats simpler. The concern about flagging Avalanche before activation no longer applies as Helicon is live on mainnet as of 22 Sept. Will update description to match

